PR #143, opened 2022-02-17
EXP DA with refund

ryley-o:
>Note - this organizes everything by-project. Once refund base price is locked in for a project, that is forever the case for a project (at least on this minter). Alternatively, this could be organized by-auction, and multiple auctions could occur for a given project. Organizing by-auction becomes overly complex and confusing in my opinion, but it is possible.

A starting point for DA with refund. Only wrote for EXP Decay DA (we could make a linear DA as well or instead).

In general:
- General auction price mechanics same as current EXP DA minter
- Users are able to collect refunds AFTER either is true:
  - A project sells out according to MINTER (i.e. `projectMaxHasBeenInvoked[_projectId] == true`)
  - Admin declares valid refund base price after a valid, configured auction has reached its base price
    - Note: this creates admin maintenance costs and overhead for projects that do not sell-out, but prevents artist from rugging users' refunds (see `Rug A.2` below)
    - Note: No security concerns with changing this to not have requirements of configured auction at base price, but opted to include to avoid accidental admin to calls (e.g. wrong projectId could cause chaos)
- Artist or admin able to collect and split revenues once users are able to collect refunds
- Enumerable Set with getter functions used to track which projects a user has uncollected refunds on
  - This enables users to call `collectAllAvailableRefunds` for stupid simple refund call (loses some gas efficiency)
  - Getter functions also allow user to use more gas performant `collectRefundForProjects(uint256[] _projectIds)` by using etherscan alone, if desired
  - Events emitted such that our subgraph also knows which projects have refunds available, and front-end can call gas-performant `collectRefundForProjects(uint256[] _projectIds)` function
  - Users can also collect refunds for a single project via `collectRefundForProject(_projectId)` function
- new `purchaseRefundable()` and `purchaseToRefundable()` functions allow users to safely purchase and assume their net mint price will be at the eventual refundable base price (See `Rug D` below)

I have not written tests. We 10000% **need** to test this contract more than any other since it will hold funds, but I think writing tests *after* we go through some amount of iteration on this contract will speed up development.

## Security Notes
Below are a few security-related scenarios that I considered while writing this. 
>(Maybe these belong in a Design Doc or something?)

**Rug A**: 
During ongoing auction, (malicious) artist reduces maxInvocations on core contract, and attempts to artificially end auction early before settling price is reached, reducing users refunds and increasing the artist’s net revenue.
**Solution A**: 
Core contract already deployed - but admin can remedy the situation. AB admin can manually reduce refund base price (call function adminOverrideReduceRefundableBasePrice). Additionally, only admin is allowed to call setProjectMaxInvocations on a minter. Thus, if an artist reduces maxInvocations on the core contract, purchases will fail on the minter, but the minter’s projectMaxHasBeenInvoked[_projectId] will remain false and projectRefundsAvailable[_projectId] will remain false, preventing a purchase from accidentally locking in an artificially high refundable base price.
**TLDR:** after an auction has started, admin has to agree with the artist to any auction changes that could artificially inflate refundable base price. That includes changes to prices, times, or maxInvocations. Admin has ability to manually reduce refundableBasePrice if a project’s refund base price has not been locked.

**Rug A.2**: 
Artist says we have to pause project (some seemingly valid reason), requires auction to be reset by admin, then sets new auction terms with ~instant end time at a high resting price. One item is purchased, and if base price were automatically locked and refunds enabled when a piece is minted at base price, an artificially high refund base price could be maliciously achieved by an artist
**Solution A.2**: 
Refund base price can only be locked in (+ refunds enabled) when (1.) by admin when base price has been reached that admin thinks is valid and fair, or (2) when projectMaxHasBeenInvoked is true on the minter. This allows admin to enable refunds once base price has been reached, but before a project sells out. This is important functionality for users who mint on a project that does not sell out immediately. It does create overhead for admin, but the alternative is that a malicious or careless artist could accidentally lock in an artificially high refund base price, hurting users.

**Rug B**: 
(malicious) Artist buys last ~20% of tokens (even after pausing core contract, but not required), causing an artificially high resting price, and only sacrifices 10% of the cost of the last 20% of tokens to AB admin. Gets to reduce user’s refunds and collect net higher profit, user’s didn’t have a fair auction. 
**Solution B**: 
There is no solution. Even if artist were prevented from minting with their own address, malicious artists could purchase from an anonymized address and get around any rules. To maximally protect users, propose that we require artists sign a contract stating that they will not purchase their own tokens during an active, ongoing DA w/Refund.

**Rug C**: 
Artist changes auction details mid-auction, artificially changing the terms of the “refund”, possibly increasing refundable base price and reducing refunds.
**Solution C**: 
Artist is not allowed to change auction after start time < block.timestamp. If something unexpected occurs and the auction needs to be paused & shifted in time, admin can reset the auction of a paused project, allowing the artist to create a new auction (new times and prices). If malicious artist refuses to create a new valid auction, admin can change the artist address of the project and eliminate a malicious artist.

**Rug D:** 
User submits purchase tx with low gas and value of 5 ETH, but expects to pay eventual refund base price. Auction settles to base price of 0.1 ETH, does not sell-out, so admin locks refund base price at 0.1 ETH to enable refunds at end of auction. Malicious or ignorant artist increases price above base price after this to e.g. 1 ETH. User’s tx is mined and user pays 1 ETH, even though they expected to only pay auction’s settling price.
**Solution D**: 
Add functions purchaseRefundable(_projectId) and purchaseToRefundable(_to, _projectId) that will only execute if refunds are not yet available. This ensures user will either pay refund base price, or the tx will revert. 
(recommend that our frontend call the “Refundable” purchase and purchaseTo functions prior to refunds becoming available, call the normal "purchase" and "purchaseTo" functions *after* refunds become available)


ryley-o:
closing in favor of #382

PR #166, opened 2022-04-19
Token holder minter

ryley-o:
Add new token holder minter. Requires purchaser to own an NFT from an allowlisited AB or PBAB project to claim the right to mint on a project. Unlimited mints per token (up to MaxInvocations) if at least one token is owned.

Note that a custom subgraph handlers will be required. Additionally, the contract does not easily enumerate which projects are allowlisted to mint a given project, meaning that we will rely on our subgraph indexing layer to provide that enumeration capability (based on historical sequence of emitted events). The contract does readily enable querying of a specific project's holder is allowed to mint a given project.

This minter is susceptible to flash loan attacks: Anyone could use a "flash-loan" to become owner of a token that allows the user to mint, mint as many tokens as they want, then return the flash loan token before the end of their transaction. The only way to prevent this would be to, e.g. only allow one mint per token used as a mint-pass. We didn't like that constraint during early iterations, and I think being susceptible to flash loan attacks is ~okay, as long as we are aware of this. For example, we should not advise artists to set a price of zero for their tokens and let holders of their previous projects mint for free - instead they should take a snapshot of acceptable token holders and use a Merkle allowlist to gate addresses and allow one mint per address.


jakerockland:
Planning to re-review this one today.

contracts/minter-suite/MinterHolderV0.sol
```diff
@@ -0,0 +1,494 @@
...
+     * @notice Allowlisted holders of NFTs at address `_nftAddress` to be
+     * considered for minting.
+     */
+    event AllowlistedNftAddress(address indexed _nftAddress);
+
+    /**
+     * @notice Removed holders of NFTs at address `_nftAddress` to be
+     * considered for minting.
+     */
+    event RemovedNftAddress(address indexed _nftAddress);
```
jakerockland, line 32: Nit: why not capitalize all of `NFT` in `{Allowlisted,Removed}NftAddress`?

contracts/minter-suite/MinterHolderV0.sol
```diff
@@ -0,0 +1,494 @@
...
+
+    /**
+     * @notice Purchases a token from project `_projectId` and sets
+     * the token's owner to `_to`.
+     * @param _to Address to be the new token's owner.
+     * @param _projectId Project ID to mint a token on.
+     * @param _ownedNftAddress ERC-721 NFT address owned by msg.sender being used to
+     * claim right to purchase.
+     * @param _ownedNftTokenId ERC-721 NFT token ID owned by msg.sender being used
+     * to claim right to purchase. Must have not yet been used to claim a mint
```
jakerockland, line 328: `Must have not yet been used to claim a mint` I think this part is vestigial based on updates we made for V0 scope ya?

contracts/minter-suite/MinterHolderV0.sol
```diff
@@ -0,0 +1,494 @@
...
+        ) {
+            projectMaxHasBeenInvoked[_projectId] = true;
+        }
+
+        // INTERACTIONS
+        // require sender to own NFT used to redeem
+        require(
+            IERC721(_ownedNftAddress).ownerOf(_ownedNftTokenId) == msg.sender,
+            "Only owner of NFT"
+        );
```
jakerockland, line 379: I think I may be overlooking something simple here, but could you clarify why is in the interactions phase and not a check?

jakerockland (approved):
Some minor nits and things to discuss re: naming conventions, but otherwise LGTM 🔥 🙏

contracts/minter-suite/MinterHolderV0.sol
ryley-o: @jakerockland I put this in interactions because we are calling `ownerOf` on a (allowlisted) token contract. Since we plan to limit this to Art Blocks+PBAB tokens, I can see why it could be classified as "checks" (since we should control the token contracts in this case).

Realistically, I don't think this check will fail very often, so I'm fine putting this anywhere. We also have nonReentrant, so there isn't a security risk placing this anywhere.

contracts/minter-suite/MinterHolderV0.sol
ryley-o: @jakerockland I can do that - should I do it everywhere? (e.g. `function removeNftAddress` -> `function removeNFTAddress`) or are you thinking only on these UpperCamelCase events?

contracts/minter-suite/MinterHolderV0.sol
jakerockland: Roger that–makes sense to me; I don't actually have a strong opinion here in either direction (within the known ecosystem of our contracts it is a check, but if expanded to things outside the ecosystem I def understand why we'd want to flag as an interaction). 

Happy to defer to your call on where to have this assertion live structurally (as CHECK vs. INTERACTIOn) but can we just elaborate on the inline documentation here to clarify with a comment the nuance we're capturing in our convo?

contracts/minter-suite/MinterHolderV0.sol
jakerockland: I think we should do it everywhere personally (using convention of always fully capitalizing acronyms). 🤷‍♂️

contracts/minter-suite/MinterHolderV0.sol
ryley-o: yes ser - fixed!

contracts/minter-suite/MinterHolderV0.sol
```diff
@@ -0,0 +1,494 @@
...
+     * considered for adding to future allowlists.
+     * @param _nftAddress NFT core address to be removed from allowlist.
+     */
+    function removeNftAddress(address _nftAddress)
+        external
+        onlyCoreWhitelisted
+    {
+        _allowedNftAddresses.remove(_nftAddress);
+        emit RemovedNftAddress(_nftAddress);
+    }
```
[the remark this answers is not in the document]
ryley-o: 🙏 much better nomenclature! Updated!

contracts/minter-suite/MinterHolderV0.sol
ryley-o: I went ahead and left it as a interaction since I expect this check to very rarely cause a failure (meaning we won't be wasting gas), and added an @dev comment in the file 😄

contracts/minter-suite/MinterHolderV0.sol
ryley-o: perfect, updated!

PR #195, opened 2022-06-10
Tune `_mintToken` method

jakerockland:
Follow up from #194, this tunes `_mintToken` further, including removing the `tokenIdToProjectId` mapping entirely.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -638,8 +638,8 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
      * @notice Gets the project ID for a given `_tokenId`.
      */
     function tokenIdToProjectId(uint256 _tokenId)
-        public
-        view
+        external
+        pure
         returns (uint256 _projectId)
     {
         return _tokenId / ONE_MILLION;
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -635,6 +635,17 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
         ];
         royaltyFeeByID = projectIdToSecondaryMarketRoyaltyPercentage[projectId];
     }
+    
+    /**
+     * @notice Gets the project ID for a given `_tokenId`.
+     */
+    function tokenIdToProjectId(uint256 _tokenId) 
+        public
+        view
+        returns (uint256 _projectId)
+    {
+        return _tokenId / ONE_MILLION
+    }
 
     /**
      * @notice Gets token URI for token ID `_tokenId`.
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -630,25 +630,13 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
             uint256 royaltyFeeByID
         )
     {
-        uint256 projectId = tokenIdToProjectId(_tokenId);
+        uint256 projectId = _tokenId / ONE_MILLION;
         artistAddress = projectIdToArtistAddress[projectId];
         additionalPayee = projectIdToAdditionalPayee[projectId];
         additionalPayeePercentage = projectIdToAdditionalPayeePercentage[
             projectId
         ];
         royaltyFeeByID = projectIdToSecondaryMarketRoyaltyPercentage[projectId];
-    }
-
-    /**
-     * @notice Derives the `projectId` for a given `_tokenId`, for convenience
-     * purposes.
-     */
-    function tokenIdToProjectId(uint256 _tokenId)
-        public
-        view
-        returns (uint256 _projectId)
-    {
-        return _tokenId - (_tokenId % ONE_MILLION);
     }
 
     /**
@@ -664,7 +652,7 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
         return
             string(
                 abi.encodePacked(
-                    projects[tokenIdToProjectId(_tokenId)].projectBaseURI,
+                    projects[_tokenId / ONE_MILLION].projectBaseURI,
                     Strings.toString(_tokenId)
                 )
             );
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -163,7 +163,7 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
         uint256 nextTokenId = (_projectId * ONE_MILLION) +
             projects[_projectId].invocations;
 
-        projects[_projectId].invocations = projects[_projectId].invocations + 1;
+        projects[_projectId].invocations++;
 
         bytes32 tokenHash = keccak256(
             abi.encodePacked(
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -176,7 +176,7 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
 
         // Do not need to also log `projectId` in event, as the `projectId` for
         // a given token can be derived from the `tokenId` with:
-        //   projectId = tokenId - (tokenId % 1_000_000)
+        //   projectId = tokenId / 1_000_000
         emit Mint(_to, nextTokenId);
 
         return nextTokenId;
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -50,8 +50,6 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
     /// Percentage of mint revenue allocated to Art Blocks
     uint256 public artblocksPercentage = 10;
 
-    mapping(uint256 => uint256) public tokenIdToProjectId;
-    //mapping(uint256 => uint256[]) internal projectIdToTokenIds;
     mapping(uint256 => bytes32) public tokenIdToHash;
     mapping(bytes32 => uint256) public hashToTokenId;
 
@@ -162,12 +160,12 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
         internal
         returns (uint256 _tokenId)
     {
-        uint256 tokenIdToBe = (_projectId * ONE_MILLION) +
+        uint256 nextTokenId = (_projectId * ONE_MILLION) +
             projects[_projectId].invocations;
 
         projects[_projectId].invocations = projects[_projectId].invocations + 1;
 
-        bytes32 hash = keccak256(
+        bytes32 tokenHash = keccak256(
             abi.encodePacked(
                 projects[_projectId].invocations,
                 block.number,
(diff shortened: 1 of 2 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -162,30 +160,31 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
...
                 blockhash(block.number - 1),
                 randomizerContract.returnValue()
             )
         );
-        tokenIdToHash[tokenIdToBe] = hash;
-        hashToTokenId[hash] = tokenIdToBe;
 
-        _mint(_to, tokenIdToBe);
+        tokenIdToHash[nextTokenId] = tokenHash;
+        hashToTokenId[tokenHash] = nextTokenId;
```
grantoesterling, line 178: I know it's been here since the beginning, but I never understood the use case - why do we need a hash to token mapping?

contracts/GenArt721CoreV3.sol
```diff
@@ -631,16 +630,25 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
...
+    /**
+     * @notice Derives the `projectId` for a given `_tokenId`, for convenience
+     * purposes.
+     */
+    function tokenIdToProjectId(uint256 _tokenId)
+        public
+        view
+        returns (uint256 _projectId)
+    {
+        return _tokenId - (_tokenId % ONE_MILLION);
```
grantoesterling, line 651: just checking bc i don't know - is this cheaper than doing `_tokenId / ONE_MILLION`?

jakerockland: IIUC, they should cost the same unless the compiler is doing some goofy stuff, as the `MOD` and `DIV` opcodes both have the same gas costs associated with them: https://ethereum.org/en/developers/docs/evm/opcodes/

That said, I don't think `_tokenId / ONE_MILLION` would give us what we want actually and am now realizing that this logic as implemented doesn't either. I believe this should actually be:

```
return (_tokenId - (_tokenId % ONE_MILLION)) / ONE_MILLION; 
```

This is making me realizing we don't have proper test coverage here. Moving this PR back into draft mode and will update with fixed logic + tests.

contracts/GenArt721CoreV3.sol
ryley-o: I also support removing this due to the extra storage gas we are adding on mint. I am not aware of any important use-cases, and our subgraph can easily perform this type of lookup.

contracts/GenArt721CoreV3.sol
```diff
@@ -162,30 +160,31 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
...
 
-        tokenIdToProjectId[tokenIdToBe] = _projectId;
-        //projectIdToTokenIds[_projectId].push(tokenIdToBe);
+        _mint(_to, nextTokenId);
 
-        emit Mint(_to, tokenIdToBe, _projectId);
+        // Do not need to also log `projectId` in event, as the `projectId` for
+        // a given token can be derived from the `tokenId` with:
+        //   projectId = tokenId - (tokenId % 1_000_000)
+        emit MintToken(_to, nextTokenId);
```
ryley-o, line 185: I'm good with this. We should just be aware that with this pattern, w can no longer filter mint events to a specific project when using web3 calls. This is because events use bloom filter logs, so only searches for specific values are possible. In reality, we would rely on the subgraph for this functionality, and I think this makes total sense!

contracts/GenArt721CoreV3.sol
```diff
@@ -162,30 +160,31 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
...
+        uint256 nextTokenId = (_projectId * ONE_MILLION) +
             projects[_projectId].invocations;
 
         projects[_projectId].invocations = projects[_projectId].invocations + 1;
 
-        bytes32 hash = keccak256(
+        bytes32 tokenHash = keccak256(
             abi.encodePacked(
                 projects[_projectId].invocations,
                 block.number,
```
ryley-o, line 171: `block.number` doesn't add any additional randomness (we are already including blockhash of prior block), so we should strip it out.

contracts/GenArt721CoreV3.sol
```diff
@@ -162,30 +160,31 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
...
                 blockhash(block.number - 1),
                 randomizerContract.returnValue()
```
ryley-o, line 173: we should better understand how much gas the randomizer contract is using here.

contracts/GenArt721CoreV3.sol
jakerockland: I don't know of any use case for this, but initially left just to not break backwards compatibility given that this is exposed publicly. That said, if we don't imagine it having any purpose we should definitely just strip it for V3. @ryley-o WDYT?

contracts/GenArt721CoreV3.sol
```diff
@@ -162,30 +160,31 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
         internal
         returns (uint256 _tokenId)
     {
-        uint256 tokenIdToBe = (_projectId * ONE_MILLION) +
+        uint256 nextTokenId = (_projectId * ONE_MILLION) +
             projects[_projectId].invocations;
 
         projects[_projectId].invocations = projects[_projectId].invocations + 1;
```
ryley-o, line 166:
```suggestion
        projects[_projectId].invocations++;
```

contracts/GenArt721CoreV3.sol
ryley-o: Any reason we shouldn't just call the new event `Mint`?

contracts/GenArt721CoreV3.sol
jakerockland: I was not thinking straight at all this morning `_tokenId / ONE_MILLION` is definitely what we want, and is equivalent to `(_tokenId - (_tokenId % ONE_MILLION)) / ONE_MILLION`. 

smh 🤦‍♂️ 

Thank you for flagging @grantoesterling

ryley-o: why wouldn't this just be `_tokenId / 1_000_000`? (noting this because this logic is used in the in-work tokenHolderMinter)

ryley-o (commented):
added some thoughts!

jakerockland:
@grantoesterling @ryley-o Do we think it necessary to block this PR on proper test coverage? Given V3 is still in early draft form and we broadly need full integration tests for V3 as a separate workstream, I wonder if we want to not block this as I'm not sure how much bandwidth I'll have to pick this back up next week and the changes here are ~smol?

contracts/GenArt721CoreV3.sol
ryley-o: updating at same time 😆 see comment above - let's remove.
Cc @lyaunzbe in case he is aware of anything using this in the pipeline

contracts/GenArt721CoreV3.sol
jakerockland: Roger that–that makes sense to me and seems like an OK trade off given that we use the subgraph anyway for reading indexed data. Only being able to filter with web3 calls for tokenID or minter address seems OK to me for the mint events themselves.

> Any reason we shouldn't just call the new event Mint?
No strong opinion here at all. Figured we may just want to name differently than old `Mint` event to make it clearer that a change was made.

contracts/GenArt721CoreV3.sol
jakerockland: True. Removing.

contracts/GenArt721CoreV3.sol
jakerockland: Yeah +++; though I do think that is probably a workstream worth investigation on its own for another day.

contracts/GenArt721CoreV3.sol
ryley-o: >Figured we may just want to name differently than old Mint event to make it clearer that a change was made.

I think rolling the contract version and changing the qty of args is enough to distinguish the old Mint from the new Mint 🤷 no super strong opinions other than MintToken seems a bit verbose.

jakerockland: Yeah that's true + fair enough. I have ~0 opinion here, so happy to switch back to `Mint` to avoid verbosity.

contracts/GenArt721CoreV3.sol
jakerockland: SGTM–I'll strip it :)

lyaunzbe: 👍 shouldn't be in use anywhere on the pipeline side

ryley-o:
>Do we think it necessary to block this PR on proper test coverage?

@jakerockland I don't think we need test coverage at this point on V3. I think V3 will need an entire audit + I want to properly use hardhat coverage plugin, so in favor of quicker iteration at this point.

contracts/GenArt721CoreV3.sol
```diff
@@ -631,16 +627,24 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
...
+    /**
+     * @notice Gets the project ID for a given `_tokenId`.
+     */
+    function tokenIdToProjectId(uint256 _tokenId) 
+        public
+        view
+        returns (uint256 _projectId)
+    {
+        return _tokenId / ONE_MILLION
     }
```
jakerockland, line 648: Given how simple this logic is, rather than actually call this `tokenIdToProjectId` method through this contract where token->project ID mapping is needed, I'm just doing `_tokenId / ONE_MILLION` inline to avoid extra gas associated w/ internal call vs. just the DIV operation if that rationale tracks from your perspective @ryley-o

jakerockland: Note that `tokenIdToProjectId` is still implemented as it is part of the public interface, which I think probably makes sense as I think providing this as a public view method is helpful for integrators/folks reading the contract who may not already know the `_tokenId / ONE_MILLION` assumption without seeing this–though we could strip this method as well.

jakerockland:
> > Do we think it necessary to block this PR on proper test coverage?
> 
> @jakerockland I don't think we need test coverage at this point on V3. I think V3 will need an entire audit + I want to properly use hardhat coverage plugin, so in favor of quicker iteration at this point.

Roger that and share the same sentiment. Moving out of draft mode.

contracts/GenArt721CoreV3.sol
lyaunzbe: I personally like keeping `tokenIdToProjectId` in the public interface, but agree with not using it internally as you describe, but it's a pretty core assumption to our contract and I think it's good to 'formalize' that on chain

ryley-o: yup, good to not add storage costs to mint sequence storing deterministic information, but instead keep the interface avail through a new public function.

jakerockland: Roger that :) glad this approach SG. Resolving.

ryley-o: oh I see what you are saying about inline elsewhere. yes that tracks. The cheapest way to do this will always be inline `_tokenId / ONE_MILLION`.
nit: can you make this external and pure (I think we can do pure since ONE_MILLION is a constant)

contracts/GenArt721CoreV3.sol
```diff
@@ -629,16 +625,24 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
...
+    /**
+     * @notice Gets the project ID for a given `_tokenId`.
+     */
+    function tokenIdToProjectId(uint256 _tokenId)
+        public
+        view
+        returns (uint256 _projectId)
+    {
+        return _tokenId / ONE_MILLION;
     }
```
ryley-o, line 646: nit: let's codify using inline elsewhere in this contract by making this external. should also be pure since ONE_MILLION is a constant.

contracts/GenArt721CoreV3.sol
```diff
@@ -160,30 +157,29 @@ contract GenArt721CoreV3 is ERC721Enumerable, IGenArt721CoreContractV3 {
...
+        tokenIdToHash[nextTokenId] = tokenHash;
 
-        tokenIdToProjectId[tokenIdToBe] = _projectId;
-        //projectIdToTokenIds[_projectId].push(tokenIdToBe);
+        _mint(_to, nextTokenId);
 
-        emit Mint(_to, tokenIdToBe, _projectId);
+        // Do not need to also log `projectId` in event, as the `projectId` for
+        // a given token can be derived from the `tokenId` with:
+        //   projectId = tokenId - (tokenId % 1_000_000)
```
ryley-o, line 179: update logic here too 😄

contracts/GenArt721CoreV3.sol
jakerockland: Yessser. Pushing a commit with those `external`/`pure` changes.

ryley-o (approved):
baased
couple of minor nits

contracts/GenArt721CoreV3.sol
jakerockland: Whoops–lol thank you.

contracts/GenArt721CoreV3.sol
jakerockland: Yessser.

PR #197, opened 2022-06-16
New PBAB minter burner 

ryley-o:
## Change Summary
Adds a new PBAB minter that has an admin-configurable mapping of `ERC20Address => bool` that defines if an ERC20 token used to purchase a project should be burned during purchase (instead of divided up between artist, additional payee, platform, and render provider).

Default behavior is to NOT burn ERC20 tokens used for payment (no change from today).

New minter is a fork of the current `GenArt721Minter_PBAB` contract.

## Motivation
This will allow projects that want to use ERC20 tokens as a "mint pass" to ensure that tokens used to purchase a project can never be "double spent" (whether accidental or maliciously).

Without this functionality, there can be incentive for PBAB platforms to escrow primary sales via additional payee functionality, resulting in an inability to properly define on-chain royalties.

## Implementation Warning
It is NOT recommended to allowlist this *in addition* to any other PBAB minters on a PBAB contract. Any existing minters on a PBAB contract should be **REMOVED** prior to allowlisting this contract.

## TODO
Prior to merging, the following must be completed:

- [ ]  test coverage for the new minter

### DIFF
--- a/contracts/PBAB+Collabs/GenArt721MinterBurner_PBAB.sol
+++ b/contracts/PBAB+Collabs/GenArt721MinterBurner_PBAB.sol
@@ -1,0 +1,472 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+import "../interfaces/0.8.x/IGenArt721CoreV2_PBAB.sol";
+import "../interfaces/0.8.x/IBonusContract.sol";
+
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
+import "@openzeppelin/contracts/interfaces/IERC20.sol";
+
+pragma solidity 0.8.9;
+
+/**
+ * @title Powered by Art Blocks minter contract that allows tokens to be
+ * minted with ETH or any ERC-20 token. Has ability to burn configured ERC-20
+ * tokens during purchase events to avoid re-use of tokens.
+ * @author Art Blocks Inc.
+ */
+contract GenArt721MinterBurner_PBAB is ReentrancyGuard {
+    /**
+     * @notice ERC-20 tokens at address `_ERC20Address` will be burned during
+     * purchases if `_doBurnDuringPurchase` is true, otherwise they will be
+     * distributed to the artist, additional payee, platform, and render
+     * provider.
+     */
+    event BurnERC20DuringPurchaseSet(
+        address indexed _ERC20Address,
+        bool _doBurnDuringPurchase
+    );
+
+    /// PBAB core contract this minter may interact with.
+    IGenArt721CoreV2_PBAB public genArtCoreContract;
+
+    uint256 constant ONE_MILLION = 1_000_000;
+
+    address payable public ownerAddress;
+    uint256 public ownerPercentage;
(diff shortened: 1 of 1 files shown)


ryley-o:
@jakerockland would appreciate a review, then I can follow up with tests.
diff between PBAB minter and new PBAB minter here: https://www.diffchecker.com/dKNpBABk

jakerockland (approved):
Excellent @ryley-o – this LGTM ser. 🙏 🚀

ryley-o:
@jakerockland just a heads-up that tests are now implemented.

I had one update to burn to `0x00...dEaD` instead of `0x00..00` because OpenZeppelin specifically prevents transfers to the zero address. 
I prefer to send to a dead address instead of assuming the tokens are burnable (especially since the Bright Moments tokens do not implement `burn()`). Try/catching the burn gives a bad ux because etherscan reports the tx as "some errors occurred".

contracts/PBAB+Collabs/GenArt721MinterBurner_PBAB.sol
```diff
@@ -46,6 +46,9 @@ contract GenArt721MinterBurner_PBAB is ReentrancyGuard {
 
     /// ERC20 address => burn during purchase
     mapping(address => bool) public burnERC20DuringPurchase;
+    // don't rely on optional ERC20 burn() to burn tokens, send here instead
+    address internal constant ERC20_BURN_ADDRESS =
+        0x000000000000000000000000000000000000dEaD;
```
jakerockland, line 51: Just curious use of `0x000000000000000000000000000000000000dEaD` vs `address(0)` here. I have no strong opinion whatsoever aside from slight preference towards standardizing our approach here in our contracts repo (that is, we should maybe make an opinion and add it to the README or a new STYLE.md file).

jakerockland: Ah whoops, I just saw the rationale for this above in https://github.com/ArtBlocks/artblocks-contracts/pull/197#issuecomment-1165990018

This approach SGTM and is probably just something worth documenting more explicitly outside the file, perhaps in starting a new STYLE.md. 

In any case, that can be handled in a new separate PR ofc.

jakerockland:
> I had one update to burn to `0x00...dEaD` instead of `0x00..00` because OpenZeppelin specifically prevents transfers to the zero address. I prefer to burn via transferring to a dead address instead of assuming the tokens are burnable (especially since the Bright Moments tokens do not implement `burn()`). Try/catching the burn gives a bad ux because etherscan reports the tx as "some errors occurred".
> 
> We can merge if you agree with this update.

Fantastic, yep this SGTM to merge!

> Can also tell Bright Moments they can deploy this upon merge. They need to b diligent when migrating to a new minter, ensuring that once this is new minter is allowlisted, their old minter is revoked. Otherwise tokens could potentially not be burned if someone were to use the old minter.

This SGTM as well, though I would say we should go further and suggest that they actually just remove the old minter before allowlisting the new one to prevent any accidental foot-gunning in the transition process.

ryley-o:
@jakerockland merging this because it is complete - anyone feel free to continue discussion on any changes requested by Bright Moments here, or submit a new issue 👍

PR #199, opened 2022-06-22
Improve Goerli docs

ryley-o:
Adding the Goerli MinterFilter addresses to our readme.
That makes it possible to understand which minters are allowlisted on our minter without looking at subgraphs or searching through many, many transactions 🥳 

(sorry for all of the markdown reformatting that was applied)


jakerockland (approved):
🙏 🙏 🙏 🙏

PR #202, opened 2022-06-29
add new bulk add-remove holders of projects functionality

ryley-o:
## Change Summary
Update MinterHolderV0 functions/events to enable bulk add/remove of projects whose owners may purchase a given project.

updated/added tests for new functionality.

## Motivation
Enabling bulk add/remove of projects will enable a better artist UX. This update will allow an artist to configure which project token holders may mint their project in a single transaction instead of requiring a transaction for every project added/removed.

## other required changes
The Approved/Removed project events have been changed to use arrays for improved gas efficiency. A minor update to our subgraph will be a follow-on to this. No changes will be required to our subgraph schema - only the event handlers will require updating.

## alternatives considered
Opted to always use approve/remove arrays of projects in events and function inputs to eliminate redundant code in the smart contract. Functions that add/remove a single project are a subset of the new functions allowing one or more projects, and defining arrays when calling functions on etherscan is easy to do.

### DIFF
--- a/contracts/minter-suite/MinterHolderV0.sol
+++ b/contracts/minter-suite/MinterHolderV0.sol
@@ -232,6 +232,11 @@ contract MinterHolderV0 is ReentrancyGuard, IFilteredMinterHolderV0 {
      * allowlisted.
      * @param _ownedNFTProjectIdsAdd Project IDs on `_ownedNFTAddressesAdd`
      * whose holders shall be allowlisted to mint project `_projectId`.
+     * @param _ownedNFTAddressesRemove NFT core addresses of projects to be
+     * removed from allowlist.
+     * @param _ownedNFTProjectIdsRemove Project IDs on
+     * `_ownedNFTAddressesRemove` whose holders will be removed from allowlist
+     * to mint project `_projectId`.
      * @dev if a project is included in both add and remove arrays, it will be
      * removed.
      */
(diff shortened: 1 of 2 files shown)


jakerockland (approved):
I think one test needs to be tweaked, but generally this LGTM @ryley-o 

One minor thing–I think that we should make the documentation more explicit here in calling out that the array of addresses and array of projects need to be matching in-order.

One other broader design question here though: is having two arrays of projects and contracts what makes the most sense vs. an input contract (singular) plus list of projects on said contract?

ryley-o:
>One other broader design question here though: is having two arrays of projects and contracts what makes the most sense vs. an input contract (singular) plus list of projects on said contract?

@jakerockland My mental model has a "project" being defined as the combination of `{core contract}-{projectId}` (consistent with our database), which is why I opted to treat adding a set of projects as a set of `{coreContractAddresses[]}-{projectIds[]}`. I think if we were to try and bundle many projects on a given contract address in calldata, we would probably want to limit to a single contract per call, which could cause the artist to submit multiple transactions when configuring their project (e.g. allow Chromie Squiggles and Stippled Sunsets), resulting in a potentially degraded UX. I think if we were to try and bundle many projects on a given contract address when emitting events, we would add complexity to our smart contract and subgraph handlers, and the added complexity may not save the artist much gas.

Overall, this tokenHolder minter is kind of assuming the artist isn't approving holders of a *significant* number of projects to mint their project. If an artist were to want to allow, e.g. *all curated AB projects*, they would have to pay a decent amount of gas, and we would want to do something along the lines of using a Merkle tree for the set of allowed projects, or even integrate with some kind of on-chain registry we might build with more on-chain project metadata.

Jbern16 (approved):

PR #204, opened 2022-07-11
CoreV2_PBAB_FLEX

lyaunzbe:
Fork of the existing V2 Core contract with added external asset dependency tracking (per project).


jakerockland:
Not ready for review IIUC, but added myself and @ryley-o just for vis.

lyaunzbe:
ready for review @jakerockland @yoshiwarab

lyaunzbe:
note: add version data

jakerockland:
Roger that–thank you @lyaunzbe. I'll try to make sure I give this an initial review pass by EOD today ideally, and if that ends up being overly optimistic I will certainly make sure to do so by EOW 🙏 

@lyaunzbe if you could throw up a diff into this PR between this and the existing PBAB/Engine template via Diffchecker, that would be super appreciated to facilitate a faster review for @yoshiwarab, @ryley-o and myself.

ryley-o:
> if you could throw up a diff into this PR between this and the existing PBAB/Engine template via Diffchecker

@jakerockland 
https://www.diffchecker.com/C4fN9UiE

lyaunzbe:
@ryley-o ty for the diff ❤️

lyaunzbe:
adding another note to remove ERC721Enumerable for gas savings

jakerockland (approved):
Couple things to discuss, but generally LGTM. 

Would be great to get @yoshiwarab or @ryley-o (or both) as a second approver before 🚢 , but I think this is right 🔥 🙌 

Great work! 💜

ryley-o (commented):
mostly nits 🔥 🔥 🔥 🙏

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,862 @@
...
+    enum ExternalAssetDependencyType {
+        IPFS,
+        ARWEAVE
+    }
+    struct ExternalAssetDependency {
+        string cid;
+        ExternalAssetDependencyType dependencyType;
+    }
+    mapping(uint256 => ExternalAssetDependency[])
+        public projectIdToExternalAssetDependencies;
```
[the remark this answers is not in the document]
ryley-o: I think that makes sense since scripts does this.

ryley-o:
@lyaunzbe - for externalAssetDependencies, if you wanted to mirror the scripts property in the Project struct and use a mapping instead of an array, it's possible that you might find gas savings. Arrays are sometimes very inefficient. If you choose to use a mapping, you will need to track your own `externalAssetDependencyCount`, otherwise, I think that can be a derived value based on the array length (can make an external getter function for it if desired).

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,862 @@
...
+
+        for (
+            uint256 i = _index;
+            i < projectIdToExternalAssetDependencies[_projectId].length - 1;
+            i++
+        ) {
+            projectIdToExternalAssetDependencies[_projectId][
+                i
+            ] = projectIdToExternalAssetDependencies[_projectId][i + 1];
+        }
```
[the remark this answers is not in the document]
ryley-o: (if we decide to use a mapping instead of array, can perform the same general operation of moving last index to index being deleted, then delete last index.)

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,862 @@
...
+        projectIdToExternalAssetDependencies[_projectId].pop();
+        projects[_projectId].externalAssetDependencyCount =
```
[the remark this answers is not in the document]
ryley-o: if we use a mapping, we will need to keep this

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,862 @@
...
+            _index < projectIdToExternalAssetDependencies[_projectId].length,
+            "Asset index out of range"
+        );
+
+        string memory _cid = projectIdToExternalAssetDependencies[_projectId][
+            _index
+        ].cid;
+        ExternalAssetDependencyType _dependencyType = projectIdToExternalAssetDependencies[
+                _projectId
+            ][_index].dependencyType;
```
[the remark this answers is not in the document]
lyaunzbe: this is actually done intentionally for easier subgraph handling. we can sync on this async to discuss more

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,862 @@
...
+        }
+        projectIdToExternalAssetDependencies[_projectId].pop();
+        projects[_projectId].externalAssetDependencyCount =
+            projects[_projectId].externalAssetDependencyCount -
+            1;
+        emit ExternalAssetDependencyRemoved(
+            _projectId,
+            _index,
+            _cid,
+            _dependencyType
```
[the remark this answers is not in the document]
lyaunzbe: this is actually done intentionally for easier subgraph handling. we can sync on this async to discuss more

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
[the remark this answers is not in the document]
lyaunzbe: yes thinking more on this it makes sense to me and aligns with potential gas savings @ryley-o mentioned below... will just have to make sure to continue tracking `externalAssetDependencyCount`

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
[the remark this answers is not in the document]
lyaunzbe: discussed async in DMs but came to the conclusion that order really doesn't matter for us, will move forward with these changes

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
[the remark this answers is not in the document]
lyaunzbe: discussed async and it does seem like we can get rid of this specifically for the Remove event

lyaunzbe:
for easier visibility, added getter functions:
`projectExternalAssetDependencyByIndex`
`projectExternalAssetDependencyCount`

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,893 @@
...
+            _index < projects[_projectId].externalAssetDependencyCount,
+            "Asset index out of range"
+        );
+
+        uint256 lastElementIndex = projects[_projectId]
+            .externalAssetDependencyCount - 1;
+
+        projects[_projectId].externalAssetDependencies[_index] = projects[
+            _projectId
+        ].externalAssetDependencies[lastElementIndex];
```
ryley-o, line 611: I'm pretty sure this is okay and will deep-copy the entire struct at index `lastElementIndex` to the struct at index `_index`.
@lyaunzbe - would appreciate if you double checked this ref and make sure you agree: https://docs.soliditylang.org/en/develop/types.html#data-location-and-assignment-behaviour

Also would be good to double check that tests ensure this deepcopys as expected.

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,893 @@
...
+        delete projects[_projectId].externalAssetDependencies[lastElementIndex];
```
ryley-o, line 612: This is good and will reset the externalAssetDependencies.
A super insane quirk when deleting structs (that doesn't apply here, but thought you might find interesting):
https://docs.soliditylang.org/en/develop/types.html#data-location-and-assignment-behaviour

Specifically:
"delete has no effect on mappings (as the keys of mappings may be arbitrary and are generally unknown). So if you delete a struct, it will reset all members that are not mappings and also recurse into the members unless they are mappings. However, individual keys and what they map to can be deleted: If a is a mapping, then delete a[x] will delete the value stored at x."

😅 😅

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,893 @@
...
+        external
+        onlyUnlockedProjectExternalAssetDependencies(_projectId)
+        onlyArtistOrWhitelisted(_projectId)
+    {
+        ExternalAssetDependency memory asset = ExternalAssetDependency({
+            cid: _cid,
+            dependencyType: _dependencyType
+        });
+        projects[_projectId].externalAssetDependencies[
+            projects[_projectId].externalAssetDependencyCount
```
ryley-o, line 641: nit: loading `projects[_projectId].externalAssetDependencyCount++` into memory will save you gas by eliminating redundant SLOAD

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,893 @@
...
+     */
+    function projectExternalAssetDependencyByIndex(
+        uint256 _projectId,
+        uint256 _index
+    ) public view returns (ExternalAssetDependency memory) {
+        return projects[_projectId].externalAssetDependencies[_index];
+    }
+
+    /**
+     * @notice Returns external asset dependency for project `_projectId` at index `_index`.
```
ryley-o, line 822: update docstring here

ryley-o (approved):
LGTM! - just nits/optimizations/double-checks 🔥 🙏

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
lyaunzbe: ahhh yes, saw that when I was transitioning. definitely funky heh

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,893 @@
...
+        string ipfsHash;
+        bool active;
+        bool locked;
+        bool paused;
+        bool externalAssetDependenciesLocked;
+        uint256 externalAssetDependencyCount;
+        mapping(uint256 => ExternalAssetDependency) externalAssetDependencies;
+    }
+
+    event ExternalAssetDependencyUpdated(
```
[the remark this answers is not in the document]
lyaunzbe: talked about this with @yoshiwarab  and we DO want to use an interface, for the events and the functions we call in the new subgraph handlers. will add this next week, don't want to block our demo work for now.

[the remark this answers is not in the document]
yoshiwarab: Yeah, if the event signatures are exactly the same between contracts we could technically reuse the same handler but I definitely feel more comfortable if the shared event types are explicit through a shared interface rather than implicit.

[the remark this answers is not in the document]
ryley-o: thanks, I'm going to tag you in a PR for V3 to apply this same logic there 🙏

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
jakerockland: @lyaunzbe could you try to take a look at this one today and then I'll do a final review pass over this PR as a sanity check and think we can 🚢 merging this.

jakerockland:
Doing a final review pass now, updated Diffchecker of this PR (at commit https://github.com/ArtBlocks/artblocks-contracts/pull/204/commits/35345f547866ffc779d91446d44f9f3119e2dac6) compared against the current PBAB template: https://www.diffchecker.com/fXTNvVyt

scripts/PBAB+Collabs/flex-engine-demo/1_flex_engine_demo_contract_suite_deployer_staging_goerli.ts
```diff
@@ -0,0 +1,257 @@
...
+  // tx = await minterFilter
+  //   .connect(deployer)
+  //   .addApprovedMinter(minterSetPrice.address);
+  // await tx.wait();
+  // console.log(`Allowlisted minter ${minterSetPrice.address} on minter filter.`);
+  // delay(EXTRA_DELAY_BETWEEN_TX);
+  // tx = await minterFilter
+  //   .connect(deployer)
+  //   .addApprovedMinter(minterSetPriceERC20.address);
+  // await tx.wait();
```
jakerockland, line 168: I have a strong preference that we don't include commented out code in PRs, can we just delete all of this commented out code?

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,899 @@
...
+        uint256 _pricePerTokenInWei
+    ) public onlyWhitelisted {
+        uint256 projectId = nextProjectId;
+        projectIdToArtistAddress[projectId] = _artistAddress;
+        projects[projectId].name = _projectName;
+        projectIdToCurrencySymbol[projectId] = "ETH";
+        projectIdToPricePerTokenInWei[projectId] = _pricePerTokenInWei;
+        projects[projectId].paused = true;
+        projects[projectId].maxInvocations = ONE_MILLION;
+        projects[projectId].externalAssetDependencyCount = 0;
```
jakerockland, line 391: Isn't the default value here `0` anyway so we don't need this line?

jakerockland (approved):
Made 2 more minor comments, which would be good to address I think before ship but otherwise would say that this LGTM to merge! 🙏 TY for all the work on this @lyaunzbe

scripts/PBAB+Collabs/flex-engine-demo/1_flex_engine_demo_contract_suite_deployer_staging_goerli.ts
ryley-o: I think this is not meant to be commented out, and was just for some deployment debugging re: needed to generate typechain

ryley-o: I think this is not meant to be commented out, and was just for some deployment debugging re: needed to generate typechain

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
```diff
@@ -0,0 +1,899 @@
...
+        uint256 maxInvocations;
+        string scriptJSON;
+        mapping(uint256 => string) scripts;
+        uint256 scriptCount;
+        string ipfsHash;
+        bool active;
+        bool locked;
+        bool paused;
+        bool externalAssetDependenciesLocked;
+        uint256 externalAssetDependencyCount;
```
ryley-o, line 38: follow-on nit: re: lessons learned here: https://github.com/ArtBlocks/artblocks-contracts/pull/248, recommend making this `uint24 externalAssetDependencyCount`, and casting to/from uint256 at points of external interfaces.

I don't think we need to fully optimize/pack this struct, but when we do that in a V4, I think this will certainly be a uint24, so starting with that will make things more consistent in the future.

ryley-o (approved):
one comment thanks to a lesson learned today.

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
jakerockland: 🙏 ➕

scripts/PBAB+Collabs/flex-engine-demo/1_flex_engine_demo_contract_suite_deployer_staging_goerli.ts
lyaunzbe: uncommented 👍

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
lyaunzbe: https://github.com/ArtBlocks/artblocks-contracts/blob/9d09585301e214dc8d54d521ad4a610b987bb0b7/test/core/GenArt721CoreV2_PBAB_FLEX_Integration.tests.ts#L134 tested here that deepcopy as working as expected 👍

contracts/PBAB+Collabs/GenArt721CoreV2_PBAB_FLEX.sol
lyaunzbe: updated to use uint24 internally, but the public convenience getter function `projectExternalAssetDependencyCount` casts and returns uint256 still

contracts/PBAB+Collabs/GenArt721CoreV2_ENGINE_FLEX.sol
```diff
@@ -0,0 +1,894 @@
...
+
+import "../interfaces/0.8.x/IRandomizer.sol";
+import "../interfaces/0.8.x/IGenArt721CoreV2_PBAB.sol";
+import "@openzeppelin-4.5/contracts/utils/Strings.sol";
+import "@openzeppelin-4.5/contracts/token/ERC721/ERC721.sol";
+
+pragma solidity 0.8.9;
+
+/**
+ * @title Powered by Art Blocks ERC-721 core contract.
```
ryley-o, line 12: nit: might want to update this to "Engine" nomenclature, and maybe something about flex?

ryley-o (approved):
LGTM!

contracts/PBAB+Collabs/GenArt721CoreV2_ENGINE_FLEX.sol
lyaunzbe: good call 👯

PR #205, opened 2022-07-20
Update Merkle library to reduce gas

ryley-o:
## Motivation
Use recent OpenZeppelin library update to pass proofs as calldata instead of memory to be more gas performant on Merkle Minter. Reduces mint costs by ~1% for 1k allowlist, based on the added test in this PR.

## Change Summary
 - alias all OpenZeppelin versions to leave existing contracts unchanged (to ensure our test suite properly tests against already-deployed contracts)
 - Add OpenZeppelin 4.7.0 because it includes new Merkle-calldata functions (instead of memory)
 - Update MerkleMinter to use calldata functions to reduce gas
   - _no change to interface of minter_
 - Add gas test for MerkleMinter for a 1k-length allowlist
 
 ## Notes
 - I'm fine with skipping deployment of this to dev and moving straight to artist-staging when we are ready (changes are very minimal here, don't affect any interfaces/any frontend or subgraph actions)
 - 1k-length allowlist adds very minimal gas. Opted to not include a 1M-length allowlist due to excessive time to generate address array, and excessive time to generate the Merkle tree. This makes me think we should recommend artists do not use allowlists longer than ~10k unless we better optimize the frontend to store the Merkle tree in our db.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -4,9 +4,9 @@
 import "./interfaces/0.8.x/IRandomizer.sol";
 import "./interfaces/0.8.x/IGenArt721CoreContractV3.sol";
 
-import "@openzeppelin/contracts/utils/Strings.sol";
+import "@openzeppelin-4.5/contracts/utils/Strings.sol";
 
-import "@openzeppelin/contracts/token/ERC721/extensions/ERC721Enumerable.sol";
+import "@openzeppelin-4.5/contracts/token/ERC721/extensions/ERC721Enumerable.sol";
 
 pragma solidity 0.8.9;
 
--- a/contracts/PBAB+Collabs/GenArt721CoreV2_PBAB.sol
+++ b/contracts/PBAB+Collabs/GenArt721CoreV2_PBAB.sol
@@ -4,9 +4,9 @@
 import "../interfaces/0.8.x/IRandomizer.sol";
 import "../interfaces/0.8.x/IGenArt721CoreV2_PBAB.sol";
 
-import "@openzeppelin/contracts/utils/Strings.sol";
+import "@openzeppelin-4.5/contracts/utils/Strings.sol";
 
-import "@openzeppelin/contracts/token/ERC721/extensions/ERC721Enumerable.sol";
+import "@openzeppelin-4.5/contracts/token/ERC721/extensions/ERC721Enumerable.sol";
 
 pragma solidity 0.8.9;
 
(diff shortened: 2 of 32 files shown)


jakerockland (approved):
Fantastic !! 🙌 

PR LGTM @ryley-o 

> This makes me think we should recommend artists do not use allowlists longer than ~10k unless we better optimize the frontend to store the Merkle tree in our db.

This part makes sense to me–let's remember to add this to the creator docs? Could you please file a ticket to follow up on this when we ship @ryley-o ?

cc @sarahrossien @madpinney for visibility.

ryley-o:
Added ticket for creator docs update, thanks for flagging!

sarahrossien:
thanks for the tag! tbh, I just heard of this allowlist mintersuite for the first time yesterday. I'd like a bit more information on how you all envision artists using this option. Once this is shipped, could I ask that someone from engineering sets up a quick overview meeting with Mad and me so that we can best advise artists? Thanks!

jakerockland:
Gah, I'm sorry that that was not on your radar sooner @sarahrossien 😓 

The primary/simplest use case for this minter is for contexts where artists were previously using a custom ERC20 token to allowlist a set of users, they can now just use an actual direct allowlist instead.

From previous conversations with @snowfro and @jeffgdavis, I believe that this minter option will be one that we only allow artists to use _outside of Curated releases_. 

Whether or not this is permissible for Collab releases, or only for Select (trying to start using our future-looking nomenclature) releases I think is something that has not yet fully been decided but inclination was to start with only Select IIRC from last convo including @snowfro and @jeffgdavis.

Is this helpful additional context?

> Once this is shipped, could I ask that someone from engineering sets up a quick overview meeting with Mad and me so that we can best advise artists?

Most certainly yes, we can and will do this of course!! 💯 

@ryley-o, @mchrupcala and/or myself should all be able to help with that.

jeffgdavis:
I would think the same, that we would likely not allow Curated projects to have allowlisting.  But definitely can be an option for Select projects (we still need to come to consensus on what that name is lol) as well as Collab projects.

— 
Jeff Davis

Art Blocks
http://www.artblocks.io <http://www.artblocks.io/>
Join us on Discord! <https://discord.gg/hHppcaETSG> 
@jeffgdavis <https://twitter.com/jeffgdavis>



On Jul 20, 2022, at 8:53 AM, Jake Rockland ***@***.***> wrote:


Gah, I'm sorry that that was not on your radar sooner @sarahrossien <https://github.com/sarahrossien> 😓

The primary/simplest use case for this minter is for contexts where artists were previously using a custom ERC20 token to allowlist a set of users, they can now just use an actual direct allowlist instead.

From previous conversations with @snowfro <https://github.com/snowfro> and @jeffgdavis <https://github.com/jeffgdavis>, I believe that this minter option will be one that we only allow artists to use outside of Curated releases.

Whether or not this is permissible for Collab releases, or only for Select (trying to start using our future-looking nomenclature) releases I think is something that has not yet fully been decided but inclination was to start with only Select IIRC from last convo including @snowfro <https://github.com/snowfro> and @jeffgdavis <https://github.com/jeffgdavis>.

Is this helpful additional context?

Once this is shipped, could I ask that someone from engineering sets up a quick overview meeting with Mad and me so that we can best advise artists?

Most certainly yes, we can and will do this of course!! 💯

@ryley-o <https://github.com/ryley-o>, @mchrupcala <https://github.com/mchrupcala> and/or myself should all be able to help with that.

PR #208, opened 2022-07-20
Make v3 ownable

ryley-o:
## Change Summary
- Implements OpenZeppelin's `Ownable` of V3 core.
- Adds a public getter `admin()` for backwards-compatibility purposes only, which returns the contract owner.
- Adds changelog to track core changes from V1 -> V3 core

## Motivation
The main benefit here is better automatic integration with third-party OpenSea/aggregators.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -5,7 +5,7 @@ import "./interfaces/0.8.x/IRandomizer.sol";
 import "./interfaces/0.8.x/IGenArt721CoreContractV3.sol";
 
 import "@openzeppelin-4.7/contracts/utils/Strings.sol";
-
+import "@openzeppelin-4.7/contracts/access/Ownable.sol";
 import "@openzeppelin-4.7/contracts/token/ERC721/ERC721.sol";
 
 pragma solidity 0.8.9;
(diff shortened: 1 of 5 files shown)


jakerockland:
#207 has been merged, making this diff'd off of main.

jakerockland:
If it is ok @ryley-o, will let you resolve the merge conflicts here before I do my review?

ryley-o:
@jakerockland ready for you 😄

jakerockland (approved):
LGTM! 🚢

contracts/GenArt721CoreV3.sol
```diff
@@ -667,4 +652,21 @@ contract GenArt721CoreV3 is ERC721, IGenArt721CoreContractV3 {
...
     }
+
+    function owner()
+        public
+        view
+        override(Ownable, IGenArt721CoreContractV3)
+        returns (address)
+    {
+        return Ownable.owner();
+    }
```
jakerockland, line 663: Nit: can we add a documentation comment here that clarifies that by inheriting Ownable, the owner is set to the deployer by default on deployment and link to this: https://docs.openzeppelin.com/contracts/2.x/api/ownership#Ownable ? 

Alternatively may make more sense to add this note in the initializer; WDYT?

ryley-o: good catch - added docstrings to this function, noted deployment behavior, and linked to OpenZeppelin docs for v4 ownable!

jakerockland: TYVM!! :)

PR #209, opened 2022-07-21
break out scriptJSON into explicit fields

ryley-o:
## Change Summary

Break out project script JSON into the fields:
- `scriptType` (e.g. "p5js")
- `scriptTypeVersion` (e.g. "1.0.0")
- `aspectRatio` (e.g. "1.7777778")

## Design Considerations
Per discussion between @lyaunzbe & @ryley-o, opted to keep these as strings instead of use something like 18 decimals and a uint. This was to keep on-chain readability maximized, and also has the side benefit of being backwards-compatible. Downside is artist could mess up formatting while setting, but existing frontend should help with that.

We also opted to not keep some kind of "miscProjectJSON" field to encourage consistency, and to avoid keeping raondom, non-standard info on-chain.

## Follow-on
Will add a follow-up PR that validates `scriptType` and `scriptTypeVersion` when being updated by artist.

### DIFF
--- a/test/core/GenArt721CoreV3_Integration.test.ts
+++ b/test/core/GenArt721CoreV3_Integration.test.ts
@@ -84,14 +84,14 @@ describe("GenArt721CoreV3", async function () {
     });
 
     describe("reverts on project locked", async function () {
-      it("reverts if try to modify script", async function () {
+      it("reverts if try to add script", async function () {
         await this.genArt721Core
           .connect(this.accounts.deployer)
           .toggleProjectIsLocked(this.projectZero);
         await expectRevert(
           this.genArt721Core
             .connect(this.accounts.artist)
-            .updateProjectScript(this.projectZero, 0, "lorem ipsum"),
+            .addProjectScript(this.projectZero, "lorem ipsum"),
           "Only if unlocked"
         );
       });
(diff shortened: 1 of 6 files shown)


jakerockland:
@ryley-o The design considerations here all LGTM :) 

I dig the approach here, will let this sit in my brain a little bit more to see if I have any additional thoughts, but generally support this direction.

cc @snowfro for visibility in case he has any thoughts/concerns.

jakerockland (approved):

ryley-o:
Cc @jakerockland - the bug fix above includes aspectRatio in view function - I accidentally left it out.

ryley-o:
gentle ping to @lyaunzbe on this one, since I think it affects you the most

test/core/GenArt721CoreV3_Integration.test.ts
```diff
@@ -91,7 +91,7 @@ describe("GenArt721CoreV3", async function () {
         await expectRevert(
           this.genArt721Core
             .connect(this.accounts.artist)
-            .updateProjectScriptJSON(this.projectZero, "lorem ipsum"),
+            .updateProjectScript(this.projectZero, 0, "lorem ipsum"),
```
lyaunzbe, line 94: nit: i know this should revert anyway, but should this be `updateProjectScriptType` here?

lyaunzbe (approved):
lgtm 🔥 🔥

test/core/GenArt721CoreV3_Integration.test.ts
ryley-o: good call, actually intended this to be addProjectScript - pushed the update!

PR #210, opened 2022-07-22
Re-organize project view functions

ryley-o:
Reorganize view functions projectInfo projectScriptInfo and projectDetails with a more intuitive set of view functions.

V3's project view functions are re-organized into the following:
- `projectStateData`
  - _"Information relevant to minters/purchasers"_
  - returns project state information - `invocations`, `maxInvocations`, `active`, `paused`, `locked`
- `projectArtistPaymentInfo`
  - _"Information relevant to managing artist/additional payment state"_
  - returns artist and additionalPayee addresses, and their split
  - (will include additionalPrimaryPayee and additionalRoyaltyPayee addrs upon future PR updates)
- `projectScriptDetails`
  - _"Information relevant to rendering"_
  - Returns similar info to the old `projectScriptInfo` function on V1, individual fields instead of a JSON (see previous PR), and no locked/paused state data.
  - Opted to rename function to reduce confusion of any integrating with our V1 and V3 contracts (i.e. be very clear that return values have changed)
- `projectDetails`
  - _"Information relevant to understanding the project as a work of art"_
  - same as V1 function  
New tests included for the view functions.

## Alternatives considered
We could cram all of these return values into one or two functions, but the return values get a bit unruly. The chosen divisions seemed more natural and intuitive to me.

We could also change nothing to minimize downstream impact, but V3 seems like the right time to organize these functions because of how much overall change there is between V1 and V3 cores.


jakerockland:
`projectDetails`-->`projectMetadata` is the only change here that at a high level I'm not so sure about–given that this is truly just a rename and we're not changing anything else about the semantics here, I'm inclined to just leave this one as-is for backwards compatibility reasons if we don't see reason to change any of the actual return values of the method.

jakerockland (approved):
This change makes sense to me and LGTM, with exception of the `projectMetadata` change which I'd like to discuss more before merge.

Appreciate the design thoughtfulness here in reworking these in a semantically coherent way 🙏 💜

ryley-o:
>Also, semantically speaking if we did this, having a projectDetails and projectScriptDetails makes sense to me.

@jakerockland Thinking about this more, I agree that just leaving this as projectDetails is good semantically, and has the benefit of being backwards-compatible. I went ahead and changed it back to use projectDetails 😄 🙏

jakerockland:
> > Also, semantically speaking if we did this, having a projectDetails and projectScriptDetails makes sense to me.
> 
> @jakerockland Thinking about this more, I agree that just leaving this as projectDetails is good semantically, and has the benefit of being backwards-compatible. I went ahead and changed it back to use projectDetails 😄 🙏

!vibe. SGTM ser. LGTM to merge but in this case given the semantic changes would love a LGTM from @yoshiwarab or @lyaunzbe for additional eyes/thoughts.

contracts/GenArt721CoreV3.sol
```diff
@@ -504,6 +504,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
      * @return description Project description
      * @return website Project website
      * @return license Project license
+     * @dev this function was named projectDetails prior to V3 core contract.
```
lyaunzbe, line 507: can we remove this comment now that we aren't changing the name?

lyaunzbe (approved):
lgtm 👍 (note: update comments/readme to reflect removal of projectMetadata, etc)

ryley-o:
> update comments/readme to reflect removal of projectMetadata, etc)

nice catch, updated!

PR #211, opened 2022-07-22
Never allow increases in edition size

ryley-o:
Never allow artists to increase edition size.

V3 (and V1) core already defaults every project to have 1M invocations when created. Artists may now only reduce maxInvocations to any value greater than or equal to current invocations.

## Design Comments
We could have stuck with the old requirement to only allow maxInvocations to be `>` current invocations. However, that is pretty quirky (e.g. "the last Rhythm"), so preferred to let artists set maxInvocations to be `>=` current invocations.

There is a side-effect of this update - artists can no longer set maxInvocations to some value for a given minter, then increase it later for a different minter. This has been done in the past, at least for PBAB partners. This side-effect is considered acceptable, and as our minter suite is expanded, more intentional ways of limiting portions of mints to certain methods are becoming available (e.g. use of allowlists).

### DIFF
--- a/contracts/V3_CHANGELOG.md
+++ b/contracts/V3_CHANGELOG.md
@@ -21,3 +21,4 @@ _This document is intended to document and explain the Art Blocks Core V3 change
     - `projectScriptDetails` - Information relevant to rendering tokens
     - `projectMetadata` - Information relevant to understanding the project as a work of art
     - `projectArtistPaymentInfo` - Information relevant to artists as they manage their primary and additional payment accounts
+- Never allow an increase in Project edition size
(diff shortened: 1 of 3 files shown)


contracts/V3_CHANGELOG.md
```diff
@@ -21,3 +21,4 @@ _This document is intended to document and explain the Art Blocks Core V3 change
     - `projectScriptDetails` - Information relevant to rendering tokens
     - `projectMetadata` - Information relevant to understanding the project as a work of art
     - `projectArtistPaymentInfo` - Information relevant to artists as they manage their primary and additional payment accounts
+- Never allow an increase in Project edition size
```
jakerockland, line 24: May be worth adding a note here in the changelog, or filing a ticket internally, but we should follow up pre-launch with an update to our creator docs making it really aggressively explicit that this is the new behavior to avoid any artist confusion on the matter.

jakerockland (approved):

contracts/V3_CHANGELOG.md
ryley-o: Thanks for the callout - I've added an internal ticket about communicating this (and any other important V3 details).

ryley-o:
I'm comfortable merging this relatively simple change with one approval. Please do to take a look retroactively @Jbern16 @lyaunzbe @yoshiwarab

PR #212, opened 2022-07-22
Automatically lock projects after four weeks

ryley-o:
Automatically lock projects four weeks after the project's completion.

Project's locked state now becomes a derived value based on time after project's complete timestamp. Project is considered "complete" when its invocations == maxInvocations. That can occur during a mint, or during an update to a project's maxInvocations (performed by artist).

A new event is emitted when a project is completed - this enables us to determine when a project will be considered "locked" on our subgraph, frontend, etc. (note that this will likely change to a generalized "ProjectUpdated" event when contract converted to use events for everything)

## follow-on work
Identified follow-on work will be to potentially adjust functions that are currently `onlyUnlocked` gated and determine if we want admin-overridable adjustments to be possible post-lock. This is because auto-locking a project could accidentally prematurely lock data, whereas in V1, locking is very intentional.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -15,6 +15,14 @@ pragma solidity 0.8.9;
  * @author Art Blocks Inc.
  */
 contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
+    event ProjectCompleted(
+        uint256 indexed _projectId,
+        uint256 _completedTimestamp
+    );
+
+    uint256 constant ONE_MILLION = 1_000_000;
+    uint256 constant FOUR_WEEKS_IN_SECONDS = 2_419_200;
+
     /// randomizer contract
     IRandomizer public randomizerContract;
 
@@ -34,11 +42,10 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         uint256 scriptCount;
         string ipfsHash;
         bool active;
-        bool locked;
         bool paused;
+        uint256 completedTimestamp;
     }
 
-    uint256 constant ONE_MILLION = 1_000_000;
     mapping(uint256 => Project) projects;
 
     // All financial functions are stripped from struct for visibility
@@ -73,7 +80,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     }
 
     modifier onlyUnlocked(uint256 _projectId) {
-        require(!projects[_projectId].locked, "Only if unlocked");
+        require(!projectLocked(_projectId), "Only if unlocked");
         _;
     }
 
@@ -133,8 +140,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
             "Must mint from the allowed minter contract."
         );
         require(
-            projects[_projectId].invocations <
-                projects[_projectId].maxInvocations,
+            projects[_projectId].completedTimestamp != 0,
             "Must not exceed max invocations"
         );
         require(
@@ -155,14 +161,21 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         internal
         returns (uint256 _tokenId)
     {
+        // checks & effects
         uint256 nextTokenId = (_projectId * ONE_MILLION) +
             projects[_projectId].invocations;
 
-        projects[_projectId].invocations++;
+        // increment invocations, register complete timestamp if applicable
+        if (
+            ++projects[_projectId].invocations ==
+            projects[_projectId].maxInvocations
+        ) {
+            _completeProject(_projectId);
+        }
 
         bytes32 tokenHash = keccak256(
             abi.encodePacked(
-                projects[_projectId].invocations,
+                nextTokenId,
                 blockhash(block.number - 1),
                 randomizerContract.returnValue()
             )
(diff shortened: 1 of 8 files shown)


ryley-o:
@jakerockland - do you mind giving this a look before I go write a bunch of tests?

contracts/GenArt721CoreV3.sol
```diff
@@ -155,21 +158,29 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+        // checks & effects
         uint256 nextTokenId = (_projectId * ONE_MILLION) +
             projects[_projectId].invocations;
 
-        projects[_projectId].invocations++;
+        // increment invocations, register complete timestamp if applicable
+        if (
+            ++projects[_projectId].invocations ==
+            projects[_projectId].maxInvocations
+        ) {
```
jakerockland, line 169: I'm tempted to split this up for readability purposes.

jakerockland:
> @jakerockland - do you mind giving this a look before I go write a bunch of tests?

Yesser–just did a pass and this approach looks right to me 🔥 💜

contracts/GenArt721CoreV3.sol
ryley-o: The thought here is to eliminate one SLOAD per mint to better optimize gas.
Want me to quantify with a test?

jakerockland: Yeah it would be nice to quantify just to make sure we're not being overly optimizooooring here–that said, given that this is in the mint flow, my inclination is to go as optimized as possible, so the solution to my concern here may just be that we make it more explicit in the comment why we do this in one line to avoid the extra SLOAD.

ryley-o: The following adds 162 gas, or $0.03/mint @100gwei, @ETH=$2k
```projects[_projectId].invocations++;
        // increment invocations, register complete timestamp if applicable
        if (
            ++projects[_projectId].invocations ==
            projects[_projectId].maxInvocations
        ) {
```
I can get even fancier and get the cost down to 34 less gas with the following, which might be overall actually easier to read?
```
function _mintToken(address _to, uint256 _projectId)
        internal
        returns (uint256 _tokenId)
    {
        // checks & effects
        // increment project's invocations, them move it to memory
        uint256 _invocationsAfter = ++projects[_projectId].invocations;
        uint256 _invocationsBefore = _invocationsAfter - 1;
        uint256 thisTokenId = (_projectId * ONE_MILLION) + _invocationsBefore;
        if (_invocationsAfter == projects[_projectId].maxInvocations) {
            _completeProject(_projectId);
        }

        bytes32 tokenHash = keccak256(
            abi.encodePacked(
                thisTokenId,
                blockhash(block.number - 1),
                randomizerContract.returnValue()
            )
        );

        tokenIdToHash[thisTokenId] = tokenHash;

        // interactions
        _mint(_to, thisTokenId);

        // Do not need to also log `projectId` in event, as the `projectId` for
        // a given token can be derived from the `tokenId` with:
        //   projectId = tokenId / 1_000_000
        emit Mint(_to, thisTokenId);

        return thisTokenId;
    }
```

jakerockland: Hrm yeah my inclination is to go with the later 34-gas-less approach, as it I think is more readable indeed and also saves users money. WDYT @ryley-o ?

ryley-o: agreed! pushed! 😄

contracts/GenArt721CoreV3.sol
```diff
@@ -155,29 +158,66 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+    }
+
+    /**
+     * @notice Returns whether a project is completed.
+     * A project is considered completed when its invocations equals its max
+     * invocations.
+     * @param _projectId Project ID to check.
+     */
+    function projectCompleted(uint256 _projectId) public view returns (bool) {
+        return projects[_projectId].completedTimestamp != 0;
```
jakerockland, line 212: Is `!=` two ops, an `EQ` and a `NOT`? I can't tell from https://ethereum.org/en/developers/docs/evm/opcodes/

If so, I wonder if we want to just make this `projectOpen` rather than `projectCompleted` to avoid the extra `!` here, and then make `projectLocked` to be `projectUnlocked` as well so there is less negating within the function and with its usage (because we actually check for unlocked status, not locked status).

jakerockland (approved):

contracts/GenArt721CoreV3.sol
ryley-o: I quite like this suggestion because projectUnlocked is actually much easier to prove (can short-circuit when completedTimestamp == 0). It also seems more intuitive (less negating).

I implemented it, and reduced the public function clutter by making the functions internal, since project state is available on `projectStateData()`. Note that I am not exposing complete timestamp because I assume having that available in our subgraph is sufficient. We could expose it in projectStateData as another return value, but we would have to educate the user what it means when the value returned is zero. I also return `locked` in the projectStateData function because in V1, we return `bool locked` in the view function `projectScriptInfo`.

jakerockland: > I implemented it, and reduced the public function clutter by narrowing everything down to a single internal function (since a view of project state is available on projectStateData()). Note that I am not exposing complete timestamp because I assume having that available in our subgraph is sufficient. We could expose it in projectStateData as another return value, but we would have to educate the user what it means when the value returned is zero. I also return locked in the projectStateData function because in V1, we return locked (not unlocked) in the V1 view function projectScriptInfo.

This rationale sounds good to me–agreed.

jakerockland (approved):

ryley-o:
gentle ping to @Jbern16 @lyaunzbe @yoshiwarab on this one

Jbern16 (approved):
LGTM! Another potential follow on work is utilizing this project complete event as a field in hasura to ultimately be used in the frontend. I believe right now we just make a calculation based on invocations

PR #213, opened 2022-07-22
Adding new contract GenArt721CoreV2WithEvents: V2 but with added events

lyaunzbe:
refer to newly added events from this list: https://docs.google.com/document/d/1lm7panDssjmfnX94WQ_1ChTsf1rIL8eh5bPQPMwG_hY/edit#heading=h.v3x6rmvh5ft3

TODO: add tests for emitted events

### DIFF
--- a/contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
+++ b/contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
@@ -196,6 +196,8 @@ contract GenArt721CoreV2WithEvents is
      */
     function updateAdmin(address _adminAddress) public onlyAdmin {
         admin = _adminAddress;
+
+        emit PlatformUpdated("admin");
     }
 
     /**
@@ -206,6 +208,8 @@ contract GenArt721CoreV2WithEvents is
         onlyAdmin
     {
         renderProviderAddress = _renderProviderAddress;
+
+        emit PlatformUpdated("renderProviderAddress");
     }
 
     /**
@@ -218,6 +222,8 @@ contract GenArt721CoreV2WithEvents is
     {
         require(_renderProviderPercentage <= 25, "Max of 25%");
         renderProviderPercentage = _renderProviderPercentage;
+
+        emit PlatformUpdated("renderProviderPercentage");
     }
 
     /**
@@ -225,6 +231,8 @@ contract GenArt721CoreV2WithEvents is
      */
     function addWhitelisted(address _address) public onlyAdmin {
         isWhitelisted[_address] = true;
+
+        emit AllowlistUpdated(true, _address);
     }
 
     /**
@@ -232,6 +240,8 @@ contract GenArt721CoreV2WithEvents is
      */
     function removeWhitelisted(address _address) public onlyAdmin {
         isWhitelisted[_address] = false;
+
+        emit AllowlistUpdated(false, _address);
     }
 
     /**
@@ -239,6 +249,8 @@ contract GenArt721CoreV2WithEvents is
      */
     function addMintWhitelisted(address _address) public onlyAdmin {
         isMintWhitelisted[_address] = true;
+
+        emit MintAllowlistUpdated(true, _address);
     }
 
     /**
@@ -246,6 +258,8 @@ contract GenArt721CoreV2WithEvents is
      */
     function removeMintWhitelisted(address _address) public onlyAdmin {
         isMintWhitelisted[_address] = false;
+
+        emit MintAllowlistUpdated(false, _address);
     }
 
     /**
@@ -256,6 +270,8 @@ contract GenArt721CoreV2WithEvents is
         onlyWhitelisted
     {
         randomizerContract = IRandomizer(_randomizerAddress);
+
+        emit PlatformUpdated("randomizerAddress");
     }
 
     /**
@@ -267,6 +283,8 @@ contract GenArt721CoreV2WithEvents is
         onlyUnlocked(_projectId)
     {
         projects[_projectId].locked = true;
+
+        emit ProjectUpdated(_projectId, "locked");
     }
 
     /**
@@ -274,6 +292,8 @@ contract GenArt721CoreV2WithEvents is
      */
     function toggleProjectIsActive(uint256 _projectId) public onlyWhitelisted {
         projects[_projectId].active = !projects[_projectId].active;
+
+        emit ProjectUpdated(_projectId, "active");
     }
 
     /**
@@ -284,6 +304,8 @@ contract GenArt721CoreV2WithEvents is
         address payable _artistAddress
     ) public onlyArtistOrWhitelisted(_projectId) {
         projectIdToArtistAddress[_projectId] = _artistAddress;
+
+        emit ProjectUpdated(_projectId, "artistAddress");
     }
 
     /**
@@ -294,6 +316,8 @@ contract GenArt721CoreV2WithEvents is
         onlyArtist(_projectId)
     {
         projects[_projectId].paused = !projects[_projectId].paused;
+
+        emit ProjectUpdated(_projectId, "paused");
     }
 
     /**
@@ -315,6 +339,8 @@ contract GenArt721CoreV2WithEvents is
         projects[projectId].paused = true;
         projects[projectId].maxInvocations = ONE_MILLION;
         nextProjectId = nextProjectId + 1;
+
+        emit ProjectUpdated(projectId, "created");
     }
 
     /**
@@ -331,6 +357,7 @@ contract GenArt721CoreV2WithEvents is
     ) public onlyArtist(_projectId) {
         projectIdToCurrencySymbol[_projectId] = _currencySymbol;
         projectIdToCurrencyAddress[_projectId] = _currencyAddress;
+        emit ProjectUpdated(_projectId, "currencyInfo");
     }
 
     /**
@@ -342,6 +369,8 @@ contract GenArt721CoreV2WithEvents is
         uint256 _pricePerTokenInWei
     ) public onlyArtist(_projectId) {
         projectIdToPricePerTokenInWei[_projectId] = _pricePerTokenInWei;
+
+        emit ProjectUpdated(_projectId, "projectIdToPricePerTokenInWei");
     }
 
     /**
@@ -353,6 +382,8 @@ contract GenArt721CoreV2WithEvents is
         onlyArtistOrWhitelisted(_projectId)
     {
         projects[_projectId].name = _projectName;
+
+        emit ProjectUpdated(_projectId, "projectName");
     }
 
     /**
@@ -364,6 +395,7 @@ contract GenArt721CoreV2WithEvents is
         string memory _projectArtistName
     ) public onlyUnlocked(_projectId) onlyArtistOrWhitelisted(_projectId) {
         projects[_projectId].artist = _projectArtistName;
+        emit ProjectUpdated(_projectId, "projectName");
     }
 
     /**
@@ -381,6 +413,8 @@ contract GenArt721CoreV2WithEvents is
         projectIdToAdditionalPayeePercentage[
             _projectId
         ] = _additionalPayeePercentage;
+
+        emit ProjectUpdated(_projectId, "additionalPayeeInfo");
     }
 
     /**
@@ -395,6 +429,8 @@ contract GenArt721CoreV2WithEvents is
         projectIdToSecondaryMarketRoyaltyPercentage[
             _projectId
         ] = _secondMarketRoyalty;
+
+        emit ProjectUpdated(_projectId, "secondaryMarketRoyaltyPercentage");
     }
 
     /**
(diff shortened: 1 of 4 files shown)


contracts/interfaces/0.8.x/IGenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * @notice Allowlist updated with bool `_isApproved` on address `_address`.
+     */
+    event AllowlistUpdated(bool _isApproved, address _address);
+
+    /**
+     * @notice Mint aAllowlist updated with bool `_isApproved` on address `_address`.
+     */
+    event MintAllowlistUpdated(bool _isApproved, address _address);
```
ryley-o, line 42:
```suggestion
    event MintAllowlistUpdated(address indexed _address, bool _isApproved);
```

(same reasoning as `AllowlistUpdated`)

ryley-o (changes requested):
This is great!

Strings in events are pretty inefficient, so I would prefer we use bytes32-encoded strings. That means we will need strings that fit in <32 bytes for everything, but I think that is totally doable.

The general pattern here would be:
(in core contract):
```
// near top of contract
bytes32 constant ARTIST_UPDATED = "artistUpdated`;
...
// in update function
emit ProjectUpdated(projectId, ARTIST_UPDATED);
```

Additionally, it would make sense for the event to have the update be indexed:
```
event ProjectUpdated(uint256 indexed _projectId, bytes32 indexed _update);
```

same goes for platformUpdated, etc.

To help keep strings <32 bytes, I recommend we drop "project" from all `_update` strings, since the event already indicates that a project is being updated. e.g. `"projectIpfsHash"` -> `"ipfsHash"`

contracts/interfaces/0.8.x/IGenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,122 @@
...
+     */
+    event ProjectUpdated(uint256 indexed _projectId, string _update);
+
+    /**
+     * @notice Project ID `_projectId` updated script at index `_scriptId`.
+     */
+    event ProjectScriptUpdatedAtIndex(
+        uint256 indexed _projectId,
+        uint256 indexed _scriptId
+    );
```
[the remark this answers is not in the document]
lyaunzbe: got it, great find 🙌

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,758 @@
...
+        return tokenIdToBe;
+    }
+
+    /**
+     * @notice Updates contract admin to `_adminAddress`.
+     */
+    function updateAdmin(address _adminAddress) public onlyAdmin {
+        admin = _adminAddress;
+
+        emit PlatformUpdated("admin");
```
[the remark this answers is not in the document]
lyaunzbe: yes agree `owner` sounds better in this case then

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,758 @@
...
+    /**
+     * @notice Updates artist name for project `_projectId` to be
+     * `_projectArtistName`.
+     */
+    function updateProjectArtistName(
+        uint256 _projectId,
+        string memory _projectArtistName
+    ) public onlyUnlocked(_projectId) onlyArtistOrWhitelisted(_projectId) {
+        projects[_projectId].artist = _projectArtistName;
+        emit ProjectUpdated(_projectId, "projectName");
```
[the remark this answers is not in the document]
lyaunzbe: oops :O

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,758 @@
...
+     * @param _randomizerContract Randomizer contract.
+     */
+    constructor(
+        string memory _tokenName,
+        string memory _tokenSymbol,
+        address _randomizerContract
+    ) ERC721(_tokenName, _tokenSymbol) {
+        admin = msg.sender;
+        isWhitelisted[msg.sender] = true;
+        renderProviderAddress = payable(msg.sender);
```
ryley-o, line 123: I think we need to emit events for these three variables being updated.

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,758 @@
...
+     * @notice Updates price per token of project `_projectId` to be
+     * '_pricePerTokenInWei`, in Wei.
+     */
+    function updateProjectPricePerTokenInWei(
+        uint256 _projectId,
+        uint256 _pricePerTokenInWei
+    ) public onlyArtist(_projectId) {
+        projectIdToPricePerTokenInWei[_projectId] = _pricePerTokenInWei;
+
+        emit ProjectUpdated(_projectId, "projectIdToPricePerTokenInWei");
```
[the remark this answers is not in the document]
yoshiwarab: +1

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,758 @@
...
+        public
+        onlyUnlocked(_projectId)
+        onlyArtistOrWhitelisted(_projectId)
+    {
+        projects[_projectId].scripts[
+            projects[_projectId].scriptCount
+        ] = _script;
+        projects[_projectId].scriptCount = projects[_projectId].scriptCount + 1;
+
+        emit ProjectUpdated(_projectId, "addProjectScript");
```
yoshiwarab, line 510: Since we always refresh the whole script in the subgraph let's just make the field here "script"

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,758 @@
...
+        require(
+            projects[_projectId].scriptCount > 0,
+            "there are no scripts to remove"
+        );
+        delete projects[_projectId].scripts[
+            projects[_projectId].scriptCount - 1
+        ];
+        projects[_projectId].scriptCount = projects[_projectId].scriptCount - 1;
+
+        emit ProjectUpdated(_projectId, "removeProjectLastScript");
```
yoshiwarab, line 550: Since we always refresh the whole script in the subgraph let's just make the field here "script"

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,758 @@
...
+    /**
+     * @notice Updates description of project `_projectId`.
+     */
+    function updateProjectDescription(
+        uint256 _projectId,
+        string memory _projectDescription
+    ) public onlyArtist(_projectId) {
+        projects[_projectId].description = _projectDescription;
+
+        emit ProjectUpdated(_projectId, "projectDescription");
```
yoshiwarab, line 445: In general I'd remove the project "project" prefix from the update field name since it's redundant considering the event name.

yoshiwarab (changes requested):

lyaunzbe:
resolved the above but still need to add tests

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,793 @@
...
+ */
+contract GenArt721CoreV2WithEvents is
+    ERC721Enumerable,
+    IGenArt721CoreV2WithEvents
+{
+    /// randomizer contract
+    IRandomizer public randomizerContract;
+
+    // generic platform event fields
+    bytes32 constant FIELD_ARTBLOCKS_OWNER = "owner";
```
ryley-o, line 25: nit: I believe this can be a PBAB contract, so WDYT about naming this `FIELD_OWNER`?

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,793 @@
...
+    bytes32 constant FIELD_PROJECT_ARTIST_NAME = "artistName";
+    bytes32 constant FIELD_SECONDARY_MARKET_ROYALTY_PERCENTAGE =
+        "secondaryMarketRoyaltyPercentage";
+    bytes32 constant FIELD_PROJECT_DESCRIPTION = "description";
+    bytes32 constant FIELD_PROJECT_WEBSITE = "website";
+    bytes32 constant FIELD_PROJECT_LICENSE = "license";
+    bytes32 constant FIELD_MAX_INVOCATIONS = "maxInvocations";
+    bytes32 constant FIELD_PROJECT_SCRIPT = "script";
+    bytes32 constant FIELD_PROJECT_IPFS_HASH = "ipfsHash";
+    bytes32 constant FIELD_PROJECT_BASE_URI = "baseURI";
```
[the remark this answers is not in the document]
ryley-o: actually, since we already emit these on a project event, I'm good just keeping this as baseURI - resolving comment myself!

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,793 @@
...
+        "renderProviderPercentage";
+    // generic project event fields
+    bytes32 constant FIELD_PROJECT_ACTIVE = "active";
+    bytes32 constant FIELD_ARTIST_ADDRESS = "artistAddress";
+    bytes32 constant FIELD_PROJECT_PAUSED = "paused";
+    bytes32 constant FIELD_PROJECT_CREATED = "created";
+    bytes32 constant FIELD_PROJECT_NAME = "name";
+    bytes32 constant FIELD_PROJECT_ARTIST_NAME = "artistName";
+    bytes32 constant FIELD_SECONDARY_MARKET_ROYALTY_PERCENTAGE =
+        "secondaryMarketRoyaltyPercentage";
```
ryley-o, line 38: This string takes up all 32 bytes.
on-chain, that should be fine, but ethers doesn't like formatting strings as bytes32 when they are >31 bytes (ref: https://docs.ethers.io/v5/api/utils/strings/#utils-formatBytes32).
Thus, what do you think about staying from edge cases here and just make this `"royaltyPercentage"`?

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
ryley-o: also, I have to change my earlier feedback on naming convention here...

I was thinking we would use this same event and the V3 convention of `owner` when implementing events on V3, but forgot that "Ownable" already implements an [OwnershipTransferred](https://docs.openzeppelin.com/contracts/2.x/api/ownership#Ownable-OwnershipTransferred-address-address-) event, so keeping this as `"admin"` probably makes the most sense since the event won't be used in future versions.

contracts/PBAB+Collabs/GenArt721CoreV2WithEvents.sol
```diff
@@ -0,0 +1,798 @@
...
+        emit PlatformUpdated(FIELD_OWNER);
+
+        isWhitelisted[msg.sender] = true;
+        emit AllowlistUpdated(msg.sender, true);
+
+        renderProviderAddress = payable(msg.sender);
+        emit PlatformUpdated(FIELD_RENDER_PROVIDER_ADDRESS);
+
+        randomizerContract = IRandomizer(_randomizerContract);
+        emit PlatformUpdated(FIELD_RANDOMIZER_ADDRESS);
```
ryley-o, line 161: to follow DRY principle, we could have update of admin, isWhitelisted, renderProviderAddress, and randomizerContract all be modified + events emitted in internal functions, something like calling the following from the constructor and updateAdmin function:
```
    /**
     * @notice Updates contract admin to `_adminAddress`.
     */
    function _updateAdmin(address _adminAddress) internal {
        admin = _adminAddress;
        emit PlatformUpdated(FIELD_OWNER);
    }
    ```
    WDYT? I can update the V3 events PR to match if you want to do this 👍

lyaunzbe: yep good call, i say lets do it 👍

lyaunzbe: will update here with relevant changes shortly

ryley-o (approved):
LGTM!
I've updated #220 to match the latest commit with internal functions being called in constructor 👍 

As a follow-on PR, can you go ahead and re-organize the functions in this to follow this style guide? https://docs.soliditylang.org/en/develop/style-guide.html#order-of-functions
(If there is a desire to keep ordering similar to existing V2, which I think is a valid position to take, I think we should at least move all internal functions below the constructor)

ryley-o:
as part of this review, if @lyaunzbe & @yoshiwarab could also review/approve #220, that would be 🔥 🔥 🙏

jakerockland:
This LGTM but per our convo should probably just be closed in favor of coalescing around using V3 directly for L2 work.

lyaunzbe:
closing in lieu of #220

PR #214, opened 2022-07-25
Add Art Blocks Reference Contracts

ryley-o:
Add enumerable mapping of reference contracts to be used for documenting items such as prior Art Blocks core contracts, dependency registry, curation registry, etc.

This adds a very general enumerable mapping of `address -> string` of reference contracts. Helps Art Blocks concretely identify canon information.

OpenZeppelin does not yet support an enumerable mapping For this implementation, we use an enumerable set of addresses and a private mapping of `address -> string` to document description. From the user's perspective, the enumerable map's length and address+description at indexes are available via new view functions.

## Design Decisions
We could break out previous token contracts into a separate mapping if desired. This mapping is quite general, so I can see a slight benefit in doing that. We could even make a static function `previousContracts` that returns the two previous AB contracts because they are known a priori.

Description strings could be limited to a bytes32 encoded string to remain more gas efficient, but I expect descriptions to generally be much longer than 32 bytes, so opted to not limit to bytes32 encoded strings.

Add/remove of reference contracts is gated to be `onlyWhitelisted`. It could be `onlyAdmin`, but would add overhead for admin, and this is just metadata.

## Follow-on PRs
Add generic platform events for add/remove references when converting V3 contract to emit events for everything (to remove subgraph call handlers).

## TODO

- [ ] Add tests

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -560,6 +560,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         _referenceContractsDescriptions[
             _referenceContract
         ] = _referenceContractDescription;
+
         // TODO - emit event when generic platform events are implemented
     }
 
@@ -816,7 +817,9 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
 
     /**
      * @notice Get reference Art Blocks contract address and description at
-     * index `_index` of enumerable mapping.
+     * index `_index` of enumerable mapping. Note that the index of a specific
+     * reference may change if references are added or removed by contract
+     * administrators.
      * @param _index enumerable map index to query.
      * @return referenceContractAddress Reference Art Blocks contract address
      * @return referenceContractDescription Reference Art Blocks contract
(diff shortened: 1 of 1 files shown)


ryley-o:
@jakerockland would like your thoughts on this architecture prior to writing tests. Specifically, regarding:
>We could break out previous token contracts into a separate mapping if desired. This mapping is quite general, so I can see a slight benefit in doing that. We could even make a static function previousContracts that returns the two previous AB contracts because they are known a priori.

jakerockland:
> Add/remove of reference contracts is gated to be onlyWhitelisted. It could be onlyAdmin, but would add overhead for admin, and this is just metadata.

IMHO, if we're abstracting out admin/ACLs as a separate abstraction as part of V3, we should remove `onlyWhitelisted` entirely in favor of admin-only approach in the core contract itself.

> We could break out previous token contracts into a separate mapping if desired. This mapping is quite general, so I can see a slight benefit in doing that. We could even make a static function previousContracts that returns the two previous AB contracts because they are known a priori.

Yeah personally I think that we should split these things out and hard-code them as constants because it offers a stronger guarantee of immutability/canon than if they are part of a more generalized mapping.

I'm actually a bit cautious towards a generalized mapping in general vs. a more domain-specific mapping, happy to chat through this async or to hop on a quick call to brainstorm together.

ryley-o:
Closing this PR in favor of something less general 🎉

PR #215, opened 2022-07-26
Gate update project description

ryley-o:
## Merge only after #212 

Gate V3 `updateProjectDescription` to be artist only when project is unlocked, owner (admin) only when project is locked.

Uses simple logic to gate calls to `updateProjectDescription`. If this pattern becomes common, will break this pattern into its own modifier.

Tests added for the new logic.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -370,11 +370,20 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
 
     /**
      * @notice Updates description of project `_projectId`.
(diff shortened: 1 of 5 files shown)


ryley-o:
@jakerockland Wanted to flag this as something never updatable if owner were to renounce ownership, which I believe is the desired behavior.

Jbern16 (approved):

yoshiwarab (approved):

jakerockland:
> @jakerockland Wanted to flag this as something never updatable if owner were to renounce ownership, which I believe is the desired behavior.

Yep, that sounds right to me. This PR LGTM :)

jakerockland (approved):

lyaunzbe (approved):

PR #216, opened 2022-07-26
Artist address updates

ryley-o:
Replace artist additionalPayee address and percentage with additionalPayeePrimarySales and additionalPayeeSecondaryRoyalties, each with a configurable percentage split.

Straightforward change that gives the artist the ability to have different additional payees for primary and secondary sales revenues. Opted to only include the primary sales payee + address in V3 core interface since those are the only values used by minters, and we already expose secondary royalties address/percentages in the getRoyaltyData function.

While digging into royalty address/percentage stuff, I reduced the cap of artist royalty percentage to 30% instead of the V1 cap of 100% (which didn't make sense, considering Art Blocks always gets 2.5%, making the cap at 102.5%).

This was the result of a discussion between accounting, creative, and engineering.

## Follow-on
This is the first step of re-organizing how artist additional payee wallets are updated. It DOES NOT implement the proposal/execute flow of updating the addresses. For this reason, full coverage of tests for this functionality are not yet included, and will be included in a follow-on PR (due to the amount of additional changes that are coming).

## ~unrelated
bumped OpenZeppelin version to from 4.7.0 to 4.7.1 to catch some bug fixes in at least one of their library (don't think it would affect us).

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -390,7 +390,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         uint256 _projectId,
         uint256 _secondMarketRoyalty
     ) public onlyArtist(_projectId) {
-        require(_secondMarketRoyalty <= 100, "Max of 100%");
+        require(_secondMarketRoyalty <= 30, "Max of 30%");
         projectIdToSecondaryMarketRoyaltyPercentage[
             _projectId
         ] = _secondMarketRoyalty;
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -349,34 +349,41 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
      * @param _projectId Project ID.
      * @param _additionalPayeePrimarySales Address that may receive a
      * percentage split of the artit's primary sales revenue.
-     * @param _additionalPayeePercentagePrimarySales Percent of artist's
+     * @param _additionalPayeePrimarySalesPercentage Percent of artist's
      * portion of primary sale revenue that will be split to address
      * `_additionalPayeePrimarySales`.
-     * @param _additionalPayeeRoyalties Address that may receive a percentage
+     * @param _additionalPayeeSecondarySales Address that may receive a percentage
      * split of the secondary sales royalties.
-     * @param _additionalPayeePercentageRoyalties Percent of artist's portion
+     * @param _additionalPayeeSecondarySalesPercentage Percent of artist's portion
      * of secondary sale royalties that will be split to address
      * `_additionalPayeeRoyalties`.
      */
     function updateProjectAdditionalPayees(
         uint256 _projectId,
         address payable _additionalPayeePrimarySales,
-        uint256 _additionalPayeePercentagePrimarySales,
-        address payable _additionalPayeeRoyalties,
-        uint256 _additionalPayeePercentageRoyalties
+        uint256 _additionalPayeePrimarySalesPercentage,
+        address payable _additionalPayeeSecondarySales,
+        uint256 _additionalPayeeSecondarySalesPercentage
     ) external onlyArtist(_projectId) {
+        // checks
+        require(
+            _additionalPayeePrimarySalesPercentage <= 100 &&
+                _additionalPayeeSecondarySalesPercentage <= 100,
+            "Max of 100%"
+        );
+        // effects
         projectIdToAdditionalPayeePrimarySales[
             _projectId
         ] = _additionalPayeePrimarySales;
         projectIdToAdditionalPayeePrimarySalesPercentage[
             _projectId
-        ] = _additionalPayeePercentagePrimarySales;
+        ] = _additionalPayeePrimarySalesPercentage;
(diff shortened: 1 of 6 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -47,8 +47,14 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
-    mapping(uint256 => address payable) public projectIdToAdditionalPayee;
-    mapping(uint256 => uint256) public projectIdToAdditionalPayeePercentage;
+    mapping(uint256 => address payable)
+        public projectIdToAdditionalPayeePrimarySales;
+    mapping(uint256 => uint256)
+        public projectIdToAdditionalPayeePrimarySalesPercentage;
+    mapping(uint256 => address payable)
+        public projectIdToAdditionalPayeeRoyalties;
+    mapping(uint256 => uint256)
+        public projectIdToAdditionalPayeeRoyaltiesPercentage;
```
jakerockland, line 57: Readability nit:

Would prefer a naming convention here of `projectIdToAdditionalPayeePrimarySales` / `projectIdToAdditionalPayeeSecondarySales` + `projectIdToAdditionalPayeePrimarySalesPercentage` / `projectIdToAdditionalPayeeSecondarySalesPercentage` for readability, I think making the two as parallel as possible makes them more idiomatic.

contracts/GenArt721CoreV3.sol
```diff
@@ -338,31 +344,53 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+     * This DOES NOT include the secondary market royalty percentages collected
+     * by Art Blocks; this is only the total percentage of royalties that will
+     * be split to artist and additionalSecondaryPayee.
      */
     function updateProjectSecondaryMarketRoyaltyPercentage(
         uint256 _projectId,
         uint256 _secondMarketRoyalty
     ) public onlyArtist(_projectId) {
-        require(_secondMarketRoyalty <= 100, "Max of 100%");
+        require(_secondMarketRoyalty <= 30, "Max of 30%");
```
jakerockland, line 393: I feel like 30% is an overly opinionated take to make at the core-contract level. I understand the concern around 100% not making sense as it adds up to 102.5%, but would propose we may just want to make this 95% then to avoid that issue? I can see an artist in the future down the line doing something interesting with decaying royalties when on-chain royalties are supported by marketplaces (though ultimately w/ wrappers, royalties do need to be enforced socially above all else) and am not sure we want to permanently peg this at 30% max which feels somewhat arbitrary to me?

jakerockland (approved):

contracts/GenArt721CoreV3.sol
ryley-o: 30% is ~arbitrary, but my thoughts here were somewhere along the lines of, "as a token holder, I would like some level of guarantee that an artist won't decide to extract more than some percentage of my token's sale value while I hold the token".

That being said, I don't have any strong opinions on the best number, and I would be fine with making this 95% (especially since royalties are socially enforced at the end of the day).

I'll update to 95%, but am open to further discussion/changes at any time prior to contract deployment!

jakerockland: Yeah that SGTM to be open to changing in the future, but 95% for now makes sense to me.

> 30% is ~arbitrary, but my thoughts here were somewhere along the lines of, "as a token holder, I would like some level of guarantee that an artist won't decide to extract more than some percentage of my token's sale value while I hold the token".

Definitely agree with your sentiment here but also feel more strongly that this stuff just needs to be socially enforced at the end of the day given it must be incorporated by the marketplace.

mchrupcala: ^for what it's worth I'd feel the same way about this: I think the contract should set the absolute upper-limit, and then we can socially guide the artist & users with our policy decisions. I totally agree that the contract gives minters peace of mind but they if they're worried about secondary rates for a particular contract they could check the projectDetails() and see the hard-coded split %

contracts/GenArt721CoreV3.sol
```diff
@@ -338,31 +344,57 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+     * of secondary sale royalties that will be split to address
+     * `_additionalPayeeRoyalties`.
      */
-    function updateProjectAdditionalPayeeInfo(
+    function updateProjectAdditionalPayees(
         uint256 _projectId,
-        address payable _additionalPayee,
-        uint256 _additionalPayeePercentage
-    ) public onlyArtist(_projectId) {
-        require(_additionalPayeePercentage <= 100, "Max of 100%");
```
mchrupcala, line 350: Did we still want to double-check that primary sales & 2ndary sales %'s don't exceed 100 in this method? Maybe there's a separate check somewhere else and I missed it 😅

mchrupcala (approved):
Approved with 2 brief comments 🚀

contracts/GenArt721CoreV3.sol
ryley-o: realized I missed another place where we had "Royalties" instead of "SecondarySales" in a function, updated and fixed.

contracts/GenArt721CoreV3.sol
ryley-o: nice catch! updated to check that the percentages are `<= 100` 🙏

PR #217, opened 2022-07-27
Delegate V3 Admin to Admin ACL contract

ryley-o:
Delegate all V3 admin authentication to a generalized Admin ACL contract, and assume Admin ACL is core contract owner.

This implementation is based on discussions between @ryley-o, @jakerockland, and @laureli(https://github.com/laureli).

In general, whenever admin may have authority to call some external function on the core contract, the core contract asks its owning ACL "is `msg.sender` allowed to call the external function being called on `address(this)`?". The ACL responds with true or false, and the core contract either rejects or accepts the call based on the response.

It is expected that eventually there will be some AdminACL contract designed that utilizes groups and sets of addresses/multisigs to determine which groups may call which functions on the core contract.

The details of this implementation are somewhat elegant in that function interfaces are relatively easy to generate in solidity, resulting in fairly readable code. This is also true in ethers, making tests fairly straightforward.

## Design Alternatives
Delegating all calls to an AdminACL (i.e. requiring all admin calls have `msg.sender == owner()`) was considered, specifically because it would be backwards compatible with V0 and V1 Art Blocks core contracts. However, if generalized, that pattern would require admin EOAs and/or multisigs to generate raw calldata in some way, and interact with the AdminACL contract directly.

Perhaps the most competitive and still relatively user-friendly alternative AdminACL contract that could be designed would have external functions that correspond 1:1 with all core contract admin-callable functions. For example, `updateProjectDescription` could be a function on the AdminACL that forwards the call to the core contract's `updateProjectDescription` function, if `msg.sender` was allowed to call the function. This AdminACL would have the drawback of having a bunch of boilerplate forwarding code for every function, and it would not be able to add support for new functions (e.g. on new minters) without making the ACL upgradeable (or reverting back to the generalized "call with this raw calldata" function, making it not user-friendly). After some discussion with the team, this was not considered a better option than the implementation in the PR.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -24,6 +24,9 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     /// randomizer contract
     IRandomizer public randomizerContract;
 
+    /// admin ACL contract
+    IAdminACLV0 public adminACLContract;
+
     struct Project {
         string name;
         string artist;
@@ -128,6 +131,17 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     }
 
     /**
+     * @dev Transfers ownership of the contract to a new account (`newOwner`).
+     * Internal function without access restriction.
+     * @dev Overrides and wraps OpenZeppelin's _transferOwnership function to
+     * also update adminACLContract for improved introspection.
+     */
+    function _transferOwnership(address newOwner) internal override {
+        Ownable._transferOwnership(newOwner);
+        adminACLContract = IAdminACLV0(newOwner);
+    }
+
+    /**
      * @notice Mints a token from project `_projectId` and sets the
      * token's owner to `_to`.
      * @param _to Address to be the minted token's owner.
@@ -203,11 +217,13 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
      * call function with selector `_selector`, as determined by the Admin ACL
      * contract.
      * @param _selector Function selector to check.
-     * @dev assumes the Admin ACL contract is the owner of this contract
-     * @dev assumes the Admin ACL contract implements the IAdminACLV0 interface
+     * @dev assumes the Admin ACL contract is the owner of this contract, which
+     * is expected to always be true.
      */
     function _adminAllowed(bytes4 _selector) internal returns (bool) {
-        return IAdminACLV0(owner()).allowed(msg.sender, _selector);
+        return
+            owner() != address(0) &&
+            adminACLContract.allowed(msg.sender, address(this), _selector);
     }
 
     /**
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -206,7 +206,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
      * @dev assumes the Admin ACL contract is the owner of this contract
      * @dev assumes the Admin ACL contract implements the IAdminACLV0 interface
      */
-    function _adminAllowed(bytes4 _selector) internal view returns (bool) {
+    function _adminAllowed(bytes4 _selector) internal returns (bool) {
         return IAdminACLV0(owner()).allowed(msg.sender, _selector);
     }
 
@@ -435,7 +435,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         require(
             _projectUnlocked(_projectId)
                 ? msg.sender == projectIdToArtistAddress[_projectId]
-                : msg.sender == owner(),
+                : _adminAllowed(this.updateProjectDescription.selector),
             "Only artist when unlocked, owner when locked"
         );
         // effects
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -2,6 +2,7 @@
 // Created By: Art Blocks Inc.
 
 import "./interfaces/0.8.x/IRandomizer.sol";
+import "./interfaces/0.8.x/IAdminACLV0.sol";
 import "./interfaces/0.8.x/IGenArt721CoreContractV3.sol";
 
 import "@openzeppelin-4.7/contracts/utils/Strings.sol";
@@ -64,9 +65,6 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
 
     mapping(uint256 => bytes32) public tokenIdToHash;
 
-    /// true if address is whitelisted
-    mapping(address => bool) public isWhitelisted;
-
     /// single minter allowed for this core contract
     address public minterContract;
 
@@ -87,6 +85,11 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         _;
     }
 
+    modifier onlyAdminACL(bytes4 _selector) {
+        require(_adminAllowed(_selector), "Only Admin ACL allowed");
+        _;
+    }
+
     modifier onlyArtist(uint256 _projectId) {
         require(
             msg.sender == projectIdToArtistAddress[_projectId],
@@ -95,16 +98,11 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         _;
     }
 
-    modifier onlyWhitelisted() {
-        require(isWhitelisted[msg.sender], "Only whitelisted");
-        _;
-    }
-
-    modifier onlyArtistOrWhitelisted(uint256 _projectId) {
+    modifier onlyArtistOrAdminACL(uint256 _projectId, bytes4 _selector) {
         require(
-            isWhitelisted[msg.sender] ||
-                msg.sender == projectIdToArtistAddress[_projectId],
-            "Only artist or whitelisted"
+            msg.sender == projectIdToArtistAddress[_projectId] ||
+                _adminAllowed(_selector),
+            "Only artist or Admin ACL allowed"
         );
         _;
     }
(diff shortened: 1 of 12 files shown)


jakerockland:
> @jakerockland - would especially appreciate hearing your thoughts on the following. I am totally open to changing to this implementation if we like it more.
> Perhaps the most competitive and still relatively user-friendly alternative AdminACL contract that could be designed would have external functions that correspond 1:1 with all core contract admin-callable functions. For example, `updateProjectDescription` could be a function on the AdminACL that forwards the call to the core contract's `updateProjectDescription` function, if `msg.sender` was allowed to call the function. This AdminACL would have the drawback of having a bunch of boilerplate forwarding code for every function, however it would be backwards-compatible with V0 and V1 contracts.

@ryley-o could we jam on this a bit synchronously and then loop back? I don't think I'm fully groking the implications here.

contracts/interfaces/0.8.x/IAdminACLV0.sol
```diff
@@ -0,0 +1,12 @@
...
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+interface IAdminACLV0 {
+    // Type of the Admin ACL contract of the form "AdminACLV0"
+    function AdminACLType() external view returns (string memory);
+
+    // Checks if sender is allowed to call function with selector `_selector`
+    function allowed(address _sender, bytes4 _selector) external returns (bool);
```
jakerockland, line 11: I think we want this method to also take a smart contract address, for the address of contract that the `_selector` is associated with–unless intention here is to use `msg.sender` in which case we should make this explicitly documented in the interface (though personally I prefer just sending address along explicitly vs. relying on `msg.sender`).

contracts/interfaces/0.8.x/IGenArt721CoreContractV3.sol
```diff
@@ -23,6 +23,7 @@ interface IGenArt721CoreContractV3 {
     function coreType() external view returns (string memory);
 
     // owner (pre-V3 was named admin) of contract
+    // this is expected to be an Admin ACL contract for V3
```
jakerockland, line 26: We may want to expose a readable field for the ACL checker of type `IAdminACLV0` in the interface itself for introspect-ability reasons–WDYT?

contracts/GenArt721CoreV3.sol
```diff
@@ -114,15 +112,19 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
         string memory _tokenSymbol,
-        address _randomizerContract
+        address _randomizerContract,
+        address _adminACLContract
     ) ERC721(_tokenName, _tokenSymbol) {
-        isWhitelisted[msg.sender] = true;
         artblocksAddress = payable(msg.sender);
         randomizerContract = IRandomizer(_randomizerContract);
+        // set AdminACL management contract as owner
+        _transferOwnership(_adminACLContract);
```
jakerockland, line 127: I feel like we may want to write this explicitly onto a variable as well to expose it directly for introspection purposes–though maybe not worth doing, unsure.

contracts/GenArt721CoreV3.sol
```diff
@@ -196,6 +198,18 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+    /**
+     * @notice Internal function that returns whether msg.sender is allowed to
+     * call function with selector `_selector`, as determined by the Admin ACL
+     * contract.
+     * @param _selector Function selector to check.
+     * @dev assumes the Admin ACL contract is the owner of this contract
+     * @dev assumes the Admin ACL contract implements the IAdminACLV0 interface
+     */
+    function _adminAllowed(bytes4 _selector) internal returns (bool) {
+        return IAdminACLV0(owner()).allowed(msg.sender, _selector);
```
jakerockland, line 210: Storing an explicit `IAdminACLV0` instance on the contract would also allow us to avoid casting here.

jakerockland (approved):
This generally LGTM, though added a couple notes on things that I think require a bit more hashing out :)

jakerockland:
cc @laureli

contracts/interfaces/0.8.x/IAdminACLV0.sol
ryley-o: I have a preference of using `msg.sender` here since it is the true source. e.g. you would never be able to accidentally have a typo in a core contract.

That being said, I can see a use case of a contract wanting to check if a sender has approval on some contract other than itself. How do you feel about overloading `allowed` and adding another function to the interface:
```
function allowed(address _sender, bytes4 _selector, address _contract) external returns (bool);
```
... and updating documentation like you suggest?

jakerockland: I don't think the typo reason is a good reason, as the same could impact the method signature and that part is passed explicitly. 

The overloading approach sounds good to me if we think we do really want to use the `msg.sender` approach, but I am quite reluctant as I just really don't love how non-explicit it is. 

Even with proper documentation it feels a bit obfuscated (taking the philosophy here that methods should ideally be self-documenting, not requiring reading of the implicit assumptions in dev documentation). 

How compelling do we think the benefit of using `msg.sender` is (may be worth some synchronous convo here, I'm a bit torn).

ryley-o: hmm, yeah chewing on this more, from the perspective of "ideally self documenting", I'm good with adding an additional arg here instead of using `msg.sender`. If I dig deeper into my psyche here, my initial preference for `msg.sender` was probably driven by gas efficiency more than anything else, but realistically, these functions are not in our mint sequence, and readability/conciseness is more important than <1% gas savings.

Thanks for chewing on this one, will update shortly!

jakerockland: Yeah I think if this intersected with the minting sequence at all I would definitely feel differently here, but given that these are just for administration method calls, and will have a very low % impact on that (we should definitely still be mindful of _high gas impact_ changes to administration methods) would definitely prefer to optimize for readability in favor of a slight gas trade off. SGTM ser 🙏 TY for the back and forth.

contracts/interfaces/0.8.x/IGenArt721CoreContractV3.sol
ryley-o: yes good call. will update!
Side note, I definitely had a blind spot here with respect to "internalType" return types in ABIs post solidity v5.<something> (ref: https://ethereum.stackexchange.com/questions/76953/what-is-the-purpose-of-internaltype-now-generated-by-the-solidity-compiler-in)

contracts/GenArt721CoreV3.sol
ryley-o: agreed - updated to override and wrap `_transferOwnership` to also update public variable `IAdminACLV0 public adminACLContract` whenever ownership is transferred.

contracts/GenArt721CoreV3.sol
ryley-o: yes, updated 🎉

contracts/interfaces/0.8.x/IGenArt721CoreContractV3.sol
ryley-o: (updated)

contracts/interfaces/0.8.x/IAdminACLV0.sol
ryley-o: (updated)

ryley-o:
@jakerockland - I think I've addressed all conversations above - prefer to have you resolve your comments since some discussion on this one if you have the time 🙏 thanks ser!

contracts/GenArt721CoreV3.sol
jakerockland: Fantastic–TYVM! :)

contracts/GenArt721CoreV3.sol
```diff
@@ -196,6 +212,20 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+     * @notice Internal function that returns whether msg.sender is allowed to
+     * call function with selector `_selector`, as determined by the Admin ACL
+     * contract.
+     * @param _selector Function selector to check.
+     * @dev assumes the Admin ACL contract is the owner of this contract, which
+     * is expected to always be true.
+     */
+    function _adminAllowed(bytes4 _selector) internal returns (bool) {
+        return
+            owner() != address(0) &&
```
jakerockland, line 225: This check probably warrants a @dev comment above or an inline comment for improved readability.

jakerockland (approved):
One additional minor comment but updated version LGTM @ryley-o 

Would love one second set of eyes before merge given the sensitivity here, but no strong opinion on who that is (just nice to have another sanity check).

contracts/GenArt721CoreV3.sol
jakerockland: LGTM! Thank you :)

contracts/interfaces/0.8.x/IGenArt721CoreContractV3.sol
jakerockland: 🙌 🙌 🙌 

>  Side note, I definitely had a blind spot here with respect to "internalType" return types in ABIs post-solidity v5.something (ref: https://ethereum.stackexchange.com/questions/76953/what-is-the-purpose-of-internaltype-now-generated-by-the-solidity-compiler-in)

Ah interesting–that was a blind spot for me as well. 😅

contracts/interfaces/0.8.x/IAdminACLV0.sol
jakerockland: LGTM 🙌  🙌  🙌

yoshiwarab (approved):

contracts/GenArt721CoreV3.sol
ryley-o: Good callout! updated!

PR #218, opened 2022-07-29
Propose/Accept pattern for V3 updated Artist payment addresses

ryley-o:
Use a propose/accept pattern when artist addresses are to be updated.

This allows Art Blocks to verify all artist + additional addresses meet regulatory and tax requirements prior to them being changed by an artist. The pattern still allows artists to propose updates publicly, on-chain, instead of through private communication channels.

## Design Decisions
While admin only cares about updated addresses, it is expected that artists would most often prefer to atomically update their additional payee address(es) at the same time as updating split percentages. Thus, all payment addresses and split percentages are proposed and accepted as a complete set. The exception is the secondary market royalty percentage, which remains completely in the control of a project's artist.

When an artist proposes an updated set of addresses and payment splits, and event is emitted with the requested values, and only the hash of the set of proposed values is stored on chain. This is to significantly reduce gas of the artist's transaction by avoiding writing every proposed value to storage. The emitted event allows our subgraph to retain the proposed values. admin may accept the values, and a hash of the values is verified on-chain to ensure admin did not modify any of the proposed values.

If admin [renounces ownership](https://docs.openzeppelin.com/contracts/4.x/api/access#Ownable-renounceOwnership--), artists may propose **and** accept updates on their own, to ensure the contract continues to be operational for artists. 

## Alternate Solutions
The artist's proposed values could be written to storage and we could avoid using our subgraph to retain proposed values. This would non-negligibly increase the cost of the artist's proposal transaction, and would require a more complex frontend to query each on-chain value when an admin wants to inspect a set of proposed values (as opposed to a graphql query in the current implementation).

## Tests
Tests are updated to provide coverage of the new functionality.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -125,6 +125,25 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     }
 
     /**
+     * This modifier allows the artist of a project to call a function if the
+     * owner of the contract has renounced ownership. This is to allow the
+     * contract to continue to function if the owner decides to renounce
+     * ownership.
+     */
+    modifier onlyAdminACLOrRenouncedArtist(
+        uint256 _projectId,
+        bytes4 _selector
+    ) {
+        require(
+            _adminAllowed(_selector) ||
+                (owner() == address(0) &&
+                    msg.sender == projectIdToArtistAddress[_projectId]),
+            "Only Admin ACL allowed, or artist if owner has renounced"
+        );
+        _;
+    }
+
+    /**
      * @notice Initializes contract.
      * @param _tokenName Name of token.
      * @param _tokenSymbol Token symbol.
@@ -401,7 +420,13 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         uint256 _additionalPayeePrimarySalesPercentage,
         address payable _additionalPayeeSecondarySales,
         uint256 _additionalPayeeSecondarySalesPercentage
-    ) external onlyAdminACL(this.adminAcceptArtistAddressesAndSplits.selector) {
+    )
+        external
+        onlyAdminACLOrRenouncedArtist(
+            _projectId,
+            this.adminAcceptArtistAddressesAndSplits.selector
+        )
+    {
         // checks
         require(
             proposedArtistAddressesAndSplitsHash[_projectId] ==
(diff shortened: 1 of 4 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -18,6 +18,17 @@ pragma solidity 0.8.9;
...
+    event ProposedArtistAddressesAndSplits(
+        uint256 indexed _projectId,
+        address _artistAddress,
+        address _additionalPayeePrimarySales,
+        uint256 _additionalPayeePrimarySalesPercentage,
+        address _additionalPayeeSecondarySales,
+        uint256 _additionalPayeeSecondarySalesPercentage
+    );
+
+    event AcceptedArtistAddressesAndSplits(uint256 indexed _projectId);
```
jakerockland, line 30: ~Should we also emit the hash of the accepted proposal here, to allow for reconciliation in the case of multiple proposals where only the most recent is accepted?~

I think my initial comment isn't necessary because we have the mapping based on projectId so there can only be one proposal at a time, but wanted to flag anyway for discussion just in case I'm overlooking antyhing.

contracts/GenArt721CoreV3.sol
```diff
@@ -305,18 +338,136 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+            _projectId
+        ] = _additionalPayeeSecondarySalesPercentage;
+        // emit event for off-chain indexing
+        emit AcceptedArtistAddressesAndSplits(_projectId);
+    }
+
+    /**
+     * @notice Updates artist of project `_projectId` to `_artistAddress`.
+     * This is to only be used in the event that the artist address is
+     * compromised or illegal.
```
jakerockland, line 465: Nit: may suggest language of `sanctioned` rather than `illegal` as it reads more naturally to me, no strong preference.

ryley-o: ++ updated!

contracts/GenArt721CoreV3.sol
ryley-o: Yes you are correct - artist can only have a single set of proposed values at a time. Additionally, since we store and check the hash of the set of values when admin approves, artists cannot do anything weird like front-run our approval transaction and cause us to accidentally approve a different set of values than we intended to approve.

ryley-o:
@yoshiwarab - narrowed down the reviewer list to just you on this. This will cause us to change up the artist edit project pages, and will probably add an admin "approve project" page. Mostly, want to make sure you think this is a good way to go about implementing a propose/reject pattern considering the impacts it will have on the subgraph + frontend.

contracts/GenArt721CoreV3.sol
jakerockland: Yeah that makes sense–I dig this design. cc @yoshiwarab on this comment thread as the second reviewer.

yoshiwarab (approved):
LGTM!

PR #219, opened 2022-07-29
Expose relevant contracts on core

ryley-o:
Expose public addresses for prior Art Blocks flagship token addresses, as well as the new dependency and curation registries.

As opposed to a generalized key-value enumerable mapping, opted to concretely define the four references added in this PR. The dependency and curation registry references are updatable by Admin-ACL-allowed addresses only.

References for dependency and curation registries are intentionally left out of constructor because these addresses may not be known at the time of V3 core deployment.

Note that events will need to be emitted when curation and dependency addresses are updated, but those events will be added in a subsequent PR (when removing all required call handlers on V3 core).

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -36,7 +36,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     address public constant ART_BLOCKS_TOKEN_ADDRESS_0 =
         0x059EDD72Cd353dF5106D2B9cC5ab83a52287aC3a;
     address public constant ART_BLOCKS_TOKEN_ADDRESS_1 =
-        0x059EDD72Cd353dF5106D2B9cC5ab83a52287aC3a;
+        0xa7d8d9ef8D8Ce8992Df33D8b8CF4Aebabd5bD270;
 
     /// Curation registry managed by Art Blocks
     address public artblocksCurationRegistryAddress;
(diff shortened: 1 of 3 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -32,6 +32,17 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     uint256 constant ONE_MILLION = 1_000_000;
     uint256 constant FOUR_WEEKS_IN_SECONDS = 2_419_200;
 
+    /// Art Blocks previous flagship token addresses (for reference)
+    address public constant ART_BLOCKS_TOKEN_ADDRESS_0 =
+        0x059EDD72Cd353dF5106D2B9cC5ab83a52287aC3a;
+    address public constant ART_BLOCKS_TOKEN_ADDRESS_1 =
+        0xa7d8d9ef8D8Ce8992Df33D8b8CF4Aebabd5bD270;
```
jakerockland, line 39: One minor thing: before deployment we should add in documentation comments the project ID ranges for each historic token address.

Also naming nit, WDYT about `ART_BLOCKS_ERC721TOKEN_ADDRESS_V0`/`_V1` nomenclature?

jakerockland (approved):

jakerockland:
Generally LGTM! One minor comment for discussion.

contracts/GenArt721CoreV3.sol
ryley-o: yes and yes 👍 end range for V1 contract currently flagged as a TODO comment

Jbern16 (approved):

PR #220, opened 2022-07-29
Emit events on V3 to eliminate subgraph call handlers

ryley-o:
_(update - #213 will not be merged) Commonality was retained as much as possible with respect to a similar update on our V2 contract in PR #213 (Cc @lyaunzbe )._

Emit events for all indexed platform and project state changes on V3 core contract.

This change enables our subgraph indexing layer to completely avoid using less performant call handlers when tracking state of the V3 core contract.

In general, bytes32-encoded strings are used to flag which field has been updated. This is more gas performant than strings (thanks to a pre-defined upper-limit on length of bytes instead of an arbitrary length of bytes that must be reported for strings). Plenty of library support for converting bytes32-encoded strings to strings (see https://docs.ethers.io/v5/api/utils/strings/#utils-parseBytes32)

We have opted to only include the field name that was updated instead of also including the field value to both reduce the number of events that must be included on the contract, and reduce the quantity of bytes being emitted. We expose all data that is being updated via public view functions, so our subgraph indexers may query our contract directly for the updated state information.

Note that if an array field is updated, such as the project script, it must be entirely re-synced in our subgraph. This is because the subgraph can only query for the updated contract at the end of a given block, and many updates may occur in separate transactions in a single block. Because a full re-sync is required in the subgraph whenever a script is added, removed, or deleted, a single "projectScriptTouched" field flag is emitted for any of the three actions to indicate that a full re-sync is required (i.e. the required subgraph action is the same for any of the three operations).

## Design Alternatives
The event `MinterUpdated` could be an operation included in the `PlatformUpdated` event. However, because Minters being updated are a very important event that has a lot of downstream impacts, keeping it as its own event that defines the new minter address event seemed prudent. Additionally, a very-unexpected edge case where the minter is updated twice in the same block with some operations in-between the transactions could lead to state errors in our subgraph if we don't include the minter address in the event (this is due to the limitation of the subgraph only being able to query the contract state at the end of the block instead of understanding the intermediate state). That is never expected to happen, and is probably not reason enough to give `MinterUpdated` and event on its own - this is more to enable users and integrators to more-easily understand the state when our minter is updated, which is an important and rare event.

### DIFF
--- a/test/core/V3/GenArt721CoreV3_Events.test.ts
+++ b/test/core/V3/GenArt721CoreV3_Events.test.ts
@@ -69,35 +69,6 @@ describe("GenArt721CoreV3 Events", async function () {
       .addProject("name", this.accounts.artist2.address);
   });
 
-  describe("ProjectCompleted", function () {
-    it("emits ProjectCompleted when being minted out", async function () {
-      for (let i = 0; i < this.maxInvocations - 1; i++) {
-        await this.genArt721Core
-          .connect(this.accounts.artist)
-          .mint(
-            this.accounts.artist.address,
-            this.projectZero,
-            this.accounts.artist.address
-          );
-      }
-      // emits event when being minted out
-      expect(
-        await this.genArt721Core
-          .connect(this.accounts.artist)
-          .mint(
-            this.accounts.artist.address,
-            this.projectZero,
-            this.accounts.artist.address
-          )
-      )
-        .to.emit(this.genArt721Core, "ProjectUpdated")
-        .withArgs(
-          this.projectZero,
-          ethers.utils.formatBytes32String("completed")
-        );
-    });
-  });
-
   describe("MinterUpdated", function () {
     it("emits MinterUpdated when being updated", async function () {
       // emits expected event arg(s)
--- a/test/core/V3/GenArt721CoreV3_Events.test.ts
+++ b/test/core/V3/GenArt721CoreV3_Events.test.ts
@@ -69,31 +69,34 @@ describe("GenArt721CoreV3 Events", async function () {
       .addProject("name", this.accounts.artist2.address);
   });
 
-  // describe("ProjectCompleted", function () {
-  //   it("emits ProjectCompleted when being minted out", async function () {
-  //     for (let i = 0; i < this.maxInvocations - 1; i++) {
-  //       await this.genArt721Core
-  //         .connect(this.accounts.artist)
-  //         .mint(
-  //           this.accounts.artist.address,
-  //           this.projectZero,
-  //           this.accounts.artist.address
-  //         );
-  //     }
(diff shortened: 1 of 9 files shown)


ryley-o:
@lyaunzbe tagging only you for review on this one since you have the corresponding V2 PR #213 👍 will add @jakerockland as a reviewer once we feel good about this one and #213

ryley-o:
Added @jakerockland & @yoshiwarab as additional reviewers - anyone definitely feel free to make appropriate changes to this PR and/or merge while I'm out. 🙏 Largely connected to whatever solution we end up with on #213

jakerockland:
> Added @jakerockland & @yoshiwarab as additional reviewers - anyone definitely feel free to make appropriate changes to this PR and/or merge while I'm out. 🙏 Largely connected to whatever solution we end up with on #213

@ryley-o just to clarify, is this review-ready?

ryley-o:
@jakerockland yes, but it should be reviewed in the context of staying consistent with approaches taken in #213. They should be consistent at the moment.
Cc @lyaunzbe

ryley-o:
update - #213 will no longer be merged, but that doesn't affect the overall approach taken in this PR

test/core/V3/GenArt721CoreV3_Events.test.ts
```diff
@@ -67,29 +69,373 @@ describe("GenArt721CoreV3 Events", async function () {
...
+  //         .mint(
+  //           this.accounts.artist.address,
+  //           this.projectZero,
+  //           this.accounts.artist.address
+  //         )
+  //     )
+  //       .to.emit(this.genArt721Core, "ProjectCompleted")
+  //       .withArgs(this.projectZero);
+  //   });
+  // });
```
jakerockland, line 96: Was this intended to be commented out?

jakerockland (approved):

test/core/V3/GenArt721CoreV3_Events.test.ts
ryley-o: ah, thank you! unintentionally commented out, when it actually needed to be watching for the new event type. updated!

ryley-o: actually, this was an old test for the old event that was replaced with the new ProjectUpdated one. Pushed another update that removed this test instead of commenting it out.

ryley-o:
@lyaunzbe gentle ping on this one since you wrote #213 would love your input before merging.

test/util/common.ts
```diff
@@ -114,11 +114,12 @@ export async function safeAddProject(
   }
 }
 
-export async function fullyMintProject(
+export async function mintProjectUntilRemaining(
```
lyaunzbe, line 117: this convenience func is niiiice 😎

PR #221, opened 2022-08-01
Get revenue splits view on v3 core

ryley-o:
## Only merge after #220 

Add an external view function on V3 core that minters may query while splitting primary sales revenues.

This offloads the often-repeated primary sale payment splitting logic from V3's minter contracts onto the V3 core itself. Placing the payment splitting logic on the core makes sense because the core contract has already been called during a purchase (see costs associated with cold (initial) contract call in [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929)). _(note: this update may very slightly increase or decrease gas costs to split a token's payment, can quantify when minters are updated to work with V3)_

Tests are added for this new payment splitting logic.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -1067,19 +1067,20 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     {
         recipients_ = new address payable[](3);
         revenues_ = new uint256[](3);
-        // Art Blocks
-        recipients_[2] = artblocksAddress;
+        // calculate revenues
         uint256 _artblocksRevenue = (_price * artblocksPercentage) / 100;
-        revenues_[2] = (_price * artblocksPercentage) / 100;
-        // additional payee
-        recipients_[1] = projectIdToAdditionalPayeePrimarySales[_projectId];
         uint256 _projectFunds = _price - _artblocksRevenue;
         uint256 _additionalPayeeRevenue = (_projectFunds *
             projectIdToAdditionalPayeePrimarySalesPercentage[_projectId]) / 100;
-        revenues_[1] = _additionalPayeeRevenue;
-        // artist
+        // Artist
         recipients_[0] = projectIdToArtistAddress[_projectId];
         revenues_[0] = _projectFunds - _additionalPayeeRevenue;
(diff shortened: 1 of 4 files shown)


ryley-o:
@mchrupcala - tagging you for review on this one since we had some discussion on this topic a while ago 🚀

contracts/GenArt721CoreV3.sol
```diff
@@ -1039,6 +1039,49 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+        revenues_[2] = (_price * artblocksPercentage) / 100;
+        // additional payee
+        recipients_[1] = projectIdToAdditionalPayeePrimarySales[_projectId];
+        uint256 _projectFunds = _price - _artblocksRevenue;
+        uint256 _additionalPayeeRevenue = (_projectFunds *
+            projectIdToAdditionalPayeePrimarySalesPercentage[_projectId]) / 100;
+        revenues_[1] = _additionalPayeeRevenue;
+        // artist
+        recipients_[0] = projectIdToArtistAddress[_projectId];
+        revenues_[0] = _projectFunds - _additionalPayeeRevenue;
```
jakerockland, line 1082: Total nit: but would prefer readability-wise that these values are returned in the order they are constructed–e.g. I find it confusing and potentially error prone that we build the array from the "back forward", is there a performance reason  I may be overlooking that we do this vs. performing our calculations (which are in-order as we need to subtract out pieces) and then doing array assignment all at the end?

jakerockland (approved):
One nit comment/question but LGTM 🙏 🙌

contracts/GenArt721CoreV3.sol
ryley-o: Yeah I'll update the assignment order. Already assigning to variable in memory, so should be same cost to assign at the end 👍

ryley-o: (update pushed)

jakerockland: 🙏 🙏

ryley-o:
Want to note that this implementation was slightly adjusted in #225 to improve gas efficiency. Thus, I think this is okay to be merged with a single approval (still will wait to merge until after #220)

mchrupcala (approved):

PR #222, opened 2022-08-02
Refactor minter directory structure

ryley-o:
## Merge only after #221 

Refactor minter directory structure to improve organization as we continue to add new minters.

This is a no-op that simply moves files around because we keep adding new minters.


jakerockland (approved):
BLESSED – tyvm @ryley-o

PR #223, opened 2022-08-02
Add Minters for V3 core

ryley-o:
## Only merge after #222 

Update all V3 minters to integrate with the new V3 core contract interface.

The primary drivers of this update are:
- V3 now uses an AdminACL contract instead of a group of allowlisted admin addresses
  - thus, we need new minters that delegate to core's ACL contract instead of `onlyAdminWhitelisted()`
- V3 has new view function for getting projectMaxInvocations
  - we need to update the `setProjectMaxInvocations` logic on minters (if we decide to keep that pattern - currently investigating if that is most gas-efficient way)
- V3 has a single MinterFilter contract at a time, so can remove `event IsCanonicalMinterfilter` from the new `MinterFilterV1`

In addition, the V3 core contract was shuffled/refactored a bit to expose a public `adminACLAllowed` function. This function can be easily used by peripheral contracts, such as minters, to determine if the core contract's active AdminACL contract allows `msg.sender` to be calling the function on the minter contract.

## Other updates
Tests for all new minters and minter filters are added. Refactoring existing tests where it made sense to do so was done (e.g. more common tests due to similarities of V2 and V1 minters).

## Not included / follow-on work
This update does not change minters to use the public `getPrimaryRevenueSplits` on the V3 core when splitting payments. This will be included in a follow-on PR to ensure we capture the gas impacts of that (optional) change.

This update does not change V3 core tests to implement V3 minters. That will be in a follow-on PR to limit the amount of changes in this PR.

Additionally, no optional minter or minter filter gas improvements are included in this change - those can be standalone follow-on PRs and their cost/benefits can be determined individually.


jakerockland (approved):

lyaunzbe (approved):
🔥 🔥

PR #224, opened 2022-08-03
V3-compatible minters use V3 core revenue splitting logic

ryley-o:
## only merge after #223 

Update all V3-compatible minters to use the V3 core's revenue splitting logic, exposed on new external function `getPrimaryRevenueSplits` function.

Placing a common primary revenue split logic in a single place on the V3 core removes some redundant code on minters. It also avoids adding another contract to the purchase flow, since the core contract has already been called during a purchase, avoiding gas costs associated with loading cold addresses.

## Gas Impacts
Gas tests indicate expected gas costs of minting are impacted as indicated in the table below:

_(All gas costs assume gas price of 100 gwei)_
_(note: This is cost of purchasing token zero, which is artificially lower than most tokens, but percent changes are ~valid)_
| Minter | Previous gas cost to purchase | New gas cost to purchase | % change |
| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 0.0182068 | 0.0185271 | +1.75 % |
| MinterDAExpV2_V3Core | 0.0193763 | 0.0196995 | +1.67 % |

⚠️  These gas impacts are relatively high, and are probably due to the use of arrays as return values in the core contract's function. A different PR will likely be opened that finds a more gas-efficient solution.

### DIFF
--- a/contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV2.sol
+++ b/contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV2.sol
@@ -324,47 +324,36 @@ contract MinterDAExpV2 is ReentrancyGuard, IFilteredMinterV0 {
         uint256 _currentPriceInWei
     ) internal {
         if (msg.value > 0) {
+            bool success_;
+            // send refund to sender
             uint256 refund = msg.value - _currentPriceInWei;
             if (refund > 0) {
-                (bool success_, ) = msg.sender.call{value: refund}("");
+                (success_, ) = msg.sender.call{value: refund}("");
                 require(success_, "Refund failed");
             }
-            uint256 foundationAmount = (_currentPriceInWei *
-                genArtCoreContract.artblocksPercentage()) / 100;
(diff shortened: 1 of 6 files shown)


ryley-o:
@jakerockland 
Just wanted to flag this for visibility if we end up at a different, more efficient solution.

I suspect the relatively large gas cost increase to mint a token is due to the use of arrays. I'll probably open a different PR that also changes the function on the core contract to become more gas efficient.

jakerockland:
Thank you very much for flagging this @ryley-o / definitely worth profiling struct returning / returning the 6 values as individual params. Was that what you're planning to do next or is there a more fancy approach to try?

ryley-o:
> definitely worth profiling struct returning / returning the 6 values as individual params. Was that what you're planning to do next or is there a more fancy approach to try?

Yup that was going to be my next iteration on this 👍

jakerockland:
Amazing 🙏

ryley-o:
closing in favor of more gas-efficient #225

PR #225, opened 2022-08-09
Update V3 minters to use efficient core revenue split

ryley-o:
## Only merge after #223 

Update core primary sales splitter function to return revenues and addresses for Art Blocks, Artist, and Artist's Additional Payee for primary sales. Also update minters to use the new core splitter function.

_(This is a more gas-efficient iteration of #224)_

This is fairly straightforward and allows minters to rely on the core contract for a portion of payment splitting logic.

## Design Choices
Payment splitting occurs during purchase transactions, so gas efficiency is important. In order to keep gas costs low, the core contract returns null addresses for Artist or Additional payees if their respective revenues are zero (avoiding the unnecessary SLOAD during the core function). This somewhat unintuitive behavior is clearly documented in the core function's comments.

## Gas impacts
Mint costs are slightly reduced, as shown in the table below. It is important to note that the original gas splitting code on each minter was not optimal, so this solution is likely not a global optimum. However, it seems like a good fit between concise and readable code, while also remaining gas-efficient.

_(All gas costs assume project 1 mint 501 of 1000)_
| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 147850 | 146059 | -1.2% (cheaper) |
| MinterDAExpV2_V3Core | 159575 | 157783 | -1.1% (cheaper) |

## Other
Standardized gas optimization reference tests are added for MinterSetPriceV2_V3Core and MinterDAExpV2_V3Core. These represent "typical" purchase transactions of our users.


jakerockland (approved):
LGTM 🙏 🙏 Thank you for such thorough analysis / description in this PR. 🔥

mchrupcala (approved):
I really like this new splitter pattern! It's so clean and very intuitive, on top of the gas savings 🔥

PR #227, opened 2022-08-09
"prime" V3 core to minimize checks during mint

ryley-o:
"Prime" V3 core to minimize checks during mint, with the goal of reducing mint gas costs.

## Recommendation
**It is RECOMMENDED TO NOT IMPLEMENT THIS SOLUTION.** The complexity and risk of managing an additional security-critical state variable does not seem to be worth the very minimal gas cost reductions observed.

This PR is created for documentation purposes only.

## Overview

This adds a mapping named `projectMayBeMintedByNonArtist` that is true only iff a project:
- is active
- is not paused
- invocations < max invocations

The mapping's state is managed during admin/artist configuration updates, such that checks during mints by non-artists may be reduced to a single SLOAD on the core.

Unfortunately, since invocations and maxInvocations must be loaded for other reasons during mint (to derive tokenId and check if project has reached max invocations, respectively), this only eliminates a single SLOAD during mint and has minimal gas savings. 

## Gas impacts
Mint costs are almost negligibly reduced, as shown in the table below.

_(All gas costs assume project 1 mint 501 of 1000)_
| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 143814 | 143600 | -0.1% (cheaper) |
| MinterDAExpV2_V3Core | 155538 | 155324 | -0.1% (cheaper) |

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -99,6 +99,16 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
 
     mapping(uint256 => Project) projects;
 
+    /**
+     * @dev mapping to gas-efficiently determine if project may be minted by
+     * non-artist address. Must be managed when configuring or affecting a
+     * project. Only intended to be true when:
+     * - project is active
+     * - project is not paused
+     * - project invocations < project max invocations
+     */
+    mapping(uint256 => bool) public projectMayBeMintedByNonArtist;
+
     // All financial functions are stripped from struct for visibility
     mapping(uint256 => address payable) public projectIdToArtistAddress;
     mapping(uint256 => address payable)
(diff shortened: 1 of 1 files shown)


ryley-o:
@jakerockland - If you agree the complexity vs. gas benefit for this means we should not implement this, can you close this with a comment for documentation purposes? Thanks!

jakerockland:
Yeah definitely agree that we should not do this–too obfuscated / error prone for the very minor gas improvement allowed by this approach. 

> It is RECOMMENDED TO NOT IMPLEMENT THIS SOLUTION. The complexity and risk of managing an additional security-critical state variable does not seem to be worth the very minimal gas cost reductions observed.
> 
> This PR is created for documentation purposes only.

Appreciate very much you creating this PR as a reference for documentation purposes. Closing out 🙏

PR #226, opened 2022-08-09
Reduce V3 mint gas costs

ryley-o:
## Only merge after #225 

Reduce V3 mint gas costs by better managing SLOAD operations during mint.

This reduces V3 core contract mint gas costs by better managing SLOAD operations on the core contract during mint. This change is very straightforward and likely doesn't affect readability of the contract in a significant way.

## Gas impacts
Mint costs are slightly reduced, as shown in the table below.

_(All gas costs assume project 1 mint 501 of 1000)_
| Gas Test | Previous gas used to purchase | New gas used to purchase | % change |
| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 146059 | 143814 | -1.5% (cheaper) |
| MinterDAExpV2_V3Core | 157783 | 155538 | -1.4% (cheaper) |

PR #228, opened 2022-08-09
Compare gas costs to mint on V1 & V3 core contracts

ryley-o:
## only merge after #226 
Update tests to directly compare the gas cost to mint on V1 vs. V3 minters.

This does not alter smart contracts - only gas tests.

## Gas Comparison
Mint costs on the V3 core are significantly reduced, as shown in the table below.

_(All gas costs for both cores assume project 3 mint 501 of 1000)_
_(All gas costs for both cores assume a simplified randomizer with artificially low gas costs for that portion of the transaction)_
| Gas Test | V1 core gas to purchase | V3 core gas to purchase | % change |
| --- | --- | --- | --- |
| MinterSetPrice | 316117 | 143830 | -54.5% (cheaper) |
| MinterDAExp | 327842 | 155554 | -52.6% (cheaper) |

Note that a decent chunk of gas reduction is due to V3 inheriting from ERC721 instead of ERC721Enumerable, meaning that the (mostly unused by us internally) [functions documented here](https://docs.openzeppelin.com/contracts/3.x/api/token/erc721#IERC721Enumerable) will not be available. If V3 core were ERC721Enumerable, gas costs to mint would be `238097` for MinterSetPrice, indicating that ~45% of the gas reductions in the V3 core are due to not making the core contract enumerable.

Note that additional work is planned to potentially further reduce V3 core gas costs, but likely only a few percent more reduction will be found.


jakerockland (approved):
This is fantastic 🙏

mchrupcala (approved):
Gas at the pump 📉  Gas at the Core 📉📉📉

PR #229, opened 2022-08-10
Update V3 core to follow style guide

ryley-o:
Update V3 core to follow [Solidity Style Guide](https://docs.soliditylang.org/en/develop/style-guide.html#)

This updates the V3 core to follow the Solidity Style Guide. Specifically, the [order of functions](https://docs.soliditylang.org/en/develop/style-guide.html#order-of-functions) and [order of layout](https://docs.soliditylang.org/en/develop/style-guide.html#order-of-layout) sections were impacting.

## Non-goals
This does not update the minter suite to follow the Solidity Style Guide. Most of those contracts are similar to the style guide, but there are some differences. I opted to not update the style of the minter suite to retain commonality to prior minter suite versions, making comparisons with previous versions more straightforward. The V3 core, on the other hand, is greatly changed from prior core contract versions, so it seemed worth updating.


jakerockland (approved):
LGTM! 

A couple notes:

1) I disagree with your rationale re: minter suite above and think we should update to match this style guide for future versions of minter (our V3-compatible versions)–my rationale for disagreement being that it is better to optimize for future-looking-consistency than past-looking-consistency

2) Can we add a link to https://docs.soliditylang.org/en/develop/style-guide.html# and that it should be followed somewhere in our main README?

PR #230, opened 2022-08-10
Optimize V3 core mint function signature for gas usage

ryley-o:
## Do not merge until after #229 

Update V3 core's mint function signature to use minimal gas when being called by MinterFilterV1.

This uses [Solidity Optimize Name](https://emn178.github.io/solidity-optimize-name/) to minimize our mint function's signature to be `mint_Ecf(address,uint256,address): 0x00005de5`.

## Design Choices
This PR does not currently optimize the mint function name on the MinterFilterV1 contract because the function is actually already has the lowest signature value! (`mint(address,uint256,address): 0x0d4d1513`, which is luckily pretty low!)

## Gas impacts
Mint costs are slightly reduced, as shown in the table below.

_(All gas costs assume project 3 mint 501 of 1000)_
| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 143830 | 143687 | -0.1% (cheaper) |
| MinterDAExpV2_V3Core | 155554 | 155411 | -0.1% (cheaper) |

## Recommendation
This saves roughly 0.1% of gas during a mint for changing a function that is only called by our Minter Suite, so I think it is worth using the new signature. However, not a very strong opinion, so would appreciate @jakerockland to weigh in with their thoughts considering our contract's readability is slightly affected.


jakerockland (approved):
Definitely agree that this change is minor enough that its worth shaving off the small % in gas costs here, as I don't think this impacts readability in a negative way when paired with the dev comments here inline that you had added. 

> This PR does not currently optimize the mint function name on the MinterFilterV1 contract because the function is actually already has the lowest signature value! (mint(address,uint256,address): 0x0d4d1513, which is luckily pretty low!)

How lucky! SGTM that this isn't needed within the `MinterFilterV1` implementation, how about for the minters mint function namings themselves?

ryley-o:
>how about for the minters mint function namings themselves?
@jakerockland I can quantify that in a follow-on PR. There is definitely some potential there, but the drawback would be a more noticeable impact to our frontend and users since `purchase_TBD` would have to be called instead of `purchase`. For minters that have different input args, such as `MinterMerkle` that needs a proof arg, the purchase function name would be changed yet again. Probably more benefit, but also more drawback!

jakerockland:
> > how about for the minters mint function namings themselves?
> 
> @jakerockland I can quantify that in a follow-on PR. There is definitely some potential there, but the drawback would be a more noticeable impact to our frontend and users since `purchase_TBD` would have to be called instead of `purchase`. For minters that have different input args, such as `MinterMerkle` that needs a proof arg, the purchase function name would be changed yet again. Probably more benefit, but also more drawback!

Makes sense to me–I think should definitely be via a separate PR, but worth exploring. Would also say that this work is less urgent than other V3 things as minters can always be updated later down the road vs. core contract.

PR #233, opened 2022-08-11
Update randomizer interface for V3 core

ryley-o:
## Only merge after #230 

Update randomizer interface for V3 core contract.

This frees some constraints on our randomizer contract at the cost of some additional gas during mints.

## Gas impacts
Mint costs are increased, as shown in the table below.

_(All gas costs assume project 3 mint 501 of 1000)_
| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 143687 | 146523 | +2.0% |
| MinterDAExpV2_V3Core | 155411 | 158247 | +1.8% |

## Tests

- [x] Update existing tests 
- [x] Add new tests for new functionality

### DIFF
--- a/contracts/BasicRandomizerV2.sol
+++ b/contracts/BasicRandomizerV2.sol
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Creatd By: Art Blocks Inc.
+
+pragma solidity 0.8.9;
+
+import "./interfaces/0.8.x/IRandomizerV2.sol";
+import "./interfaces/0.8.x/IGenArt721CoreContractV3.sol";
+
+import "@openzeppelin-4.7/contracts/access/Ownable.sol";
+
+contract BasicRandomizerV2 is IRandomizerV2, Ownable {
+    // The core contract that may interact with this randomizer contract.
+    IGenArt721CoreContractV3 public genArt721Core;
+
+    function assignCoreAndRenounce(address _genArt721Core) external onlyOwner {
+        renounceOwnership();
+        genArt721Core = IGenArt721CoreContractV3(_genArt721Core);
+    }
+
+    // When `genArt721Core` calls this, it can be assured that the randomizer
+    // will set a bytes32 hash for tokenId `_tokenId` on the core contract.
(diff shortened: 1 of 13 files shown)


ryley-o:
@jakerockland - would appreciate your thoughts on this one

contracts/BasicRandomizerV2.sol
```diff
@@ -0,0 +1,35 @@
...
+
+    function assignCoreAndRenounce(address _genArt721Core) external onlyOwner {
+        renounceOwnership();
+        genArt721Core = IGenArt721CoreContractV3(_genArt721Core);
+    }
+
+    // When `genArt721Core` calls this, it can be assured that the randomizer
+    // will set a bytes32 hash for tokenId `_tokenId` on the core contract.
+    function assignTokenHash(uint256 _tokenId) external {
+        uint256 time = block.timestamp;
```
jakerockland, line 23: I think this should have an assertion that the `msg.sender` is `genArt721Core` ya?

jakerockland: Ah nevermind, I see we have the assertion in `setTokenHash_8PT`. That said, probably worth a dev documentation comment making this assumption by the randomizer and explicit one.

jakerockland (approved):

contracts/BasicRandomizerV2.sol
ryley-o: no you are correct - this does need that assertion. I implemented the fix last night but forgot to push the commit! updated!

jakerockland: Ah yeah the assertion definitely does need to go both ways 😅 

I probably shouldn't review PRs before coffee. TY for updating!

jakerockland:
Only follow up here is storing randomizer as a push-only type stack, vs. as a single field, per our sync today–though that can also be done in a separate PR and we can merge this one without that change 👍 😄

ryley-o:
>Only follow up here is storing randomizer as a push-only type stack, vs. as a single field, per our sync today–though that can also be done in a separate PR and we can merge this one without that change 

see #235 👍

PR #234, opened 2022-08-11
Require token to exist before randomizer may set hash

ryley-o:


contracts/GenArt721CoreV3.sol
```diff
@@ -288,6 +287,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
             tokenIdToHash[_tokenId] == bytes32(0),
             "Token hash already set."
         );
+        require(_exists(_tokenId), "Token does not exist");
```
jakerockland, line 290: May be worth noting in a doc comment that `_exists` is an internal function provided by ERC721 OpenZepplin implementation.

jakerockland (approved):

contracts/GenArt721CoreV3.sol
ryley-o: ah, thanks for calling this out. just realized we already have a modifier for this check called `onlyValidTokenId`, so updated to use that! 🙏

PR #235, opened 2022-08-11
Add history of randomizers to V3 core

ryley-o:
## Only merge after #234

Add queryable history of randomizer to V3 core.

This adds an append-only historical array of all randomizer addresses ever used by the V3 core. This helps with documenting on-chain history of the contract for our users.

Under the hood, opted to use a private array of historical randomizer addresses. From a user's perspective, there are two new functions that enable enumeration of all historical randomizers, in the order they were used by the contract. Chose to give views for qty and query at a given index because it is our typical pattern, and guarantees return values fit within block limits. It also minimizes the cost associated with return value size if a contract wants to query our contract at a specific index.

Note that this could have been implemented with a private mapping and index counter, but a dynamic storage array is a bit more more intuitive for this use case.

## Alternatives
If we care about surfacing which randomizer was being used at a given block, this could be implemented via a Checkpoints data structure instead: https://docs.openzeppelin.com/contracts/4.x/api/utils#Checkpoints
(I don't think we care enough about supporting that query to import the lib, but wanted to point it out as an option).

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -60,8 +60,14 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     /// Dependency registry managed by Art Blocks
     address public artblocksDependencyRegistryAddress;
 
-    /// randomizer contract
+    /// current randomizer contract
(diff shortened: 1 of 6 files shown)


jakerockland (approved):
🙏 🔥 TY ser.

PR #236, opened 2022-08-11
Add Art Blocks secondary royalty data to V3 core

ryley-o:
## Only merge after #235 

Add Art Blocks secondary royalty data to the V3 core.

Specifically, this adds an `artblocksSecondarySalesAddress` and `artblocksSecondarySalesBPS` public, configurable (by admin) field. The values are applied across the entire contract (i.e. no specific values per-project).

Chose to also update legacy names `artblocksAddress` -> `artblocksPrimarySalesAddress` and `artblocksPercentage` -> `artblocksPrimarySalesPercentage`, to make variable names more intuitive. The impact of breaking this portion of the legacy interface is minimized thanks to the new primary payment splitter function on the V3 core contract, as well as a follow-on PR where we will support Manifold's Royalty Registry directly (i.e. we won't need to deploy a new shim layer for breaking the legacy interface).

Also, I put a constraint for admin to never increase our basis points above 250 (2.5%). Input from @jakerockland would be much appreciated on that constraint!

## Follow-on work
A follow-on PR will be added to support Manifold's Royalty Registry directly

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -20,10 +20,10 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     uint256 constant FOUR_WEEKS_IN_SECONDS = 2_419_200;
 
     // generic platform event fields
-    bytes32 constant FIELD_ARTBLOCKS_ADDRESS_PRIMARY_SALES =
-        "artblocksAddressPrimarySales";
-    bytes32 constant FIELD_ARTBLOCKS_ADDRESS_SECONDARY_SALES =
-        "artblocksAddressSecondarySales";
+    bytes32 constant FIELD_ARTBLOCKS_PRIMARY_SALES_ADDRESS =
+        "artblocksPrimarySalesAddress";
+    bytes32 constant FIELD_ARTBLOCKS_SECONDARY_SALES_ADDRESS =
+        "artblocksSecondarySalesAddress";
     bytes32 constant FIELD_RANDOMIZER_ADDRESS = "randomizerAddress";
     bytes32 constant FIELD_ARTBLOCKS_CURATION_REGISTRY_ADDRESS =
         "curationRegistryAddress";
@@ -115,11 +115,11 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     mapping(uint256 => bytes32) public proposedArtistAddressesAndSplitsHash;
 
     /// Art Blocks payment address for all primary sales revenues
-    address payable public artblocksAddressPrimarySales;
+    address payable public artblocksPrimarySalesAddress;
     /// Percentage of primary sales revenue allocated to Art Blocks
     uint256 public artblocksPrimarySalesPercentage = 10;
     /// Art Blocks payment address for all secondary sales royalty revenues
-    address payable public artblocksAddressSecondarySales;
+    address payable public artblocksSecondarySalesAddress;
     /// Basis Points of secondary sales royalties allocated to Art Blocks
     uint256 public artblocksSecondarySalesBPS = 250;
 
@@ -214,8 +214,8 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         address _randomizerContract,
         address _adminACLContract
     ) ERC721(_tokenName, _tokenSymbol) {
-        _updateArtblocksAddressPrimarySales(msg.sender);
-        _updateArtblocksAddressSecondarySales(msg.sender);
+        _updateArtblocksPrimarySalesAddress(msg.sender);
+        _updateArtblocksSecondarySalesAddress(msg.sender);
         _updateRandomizerAddress(_randomizerContract);
         // set AdminACL management contract as owner
         _transferOwnership(_adminACLContract);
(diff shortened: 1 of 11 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -1049,15 +1087,15 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+        artblocksRevenue_ = (_price * artblocksPrimarySalesPercentage) / 100;
         uint256 projectFunds = _price - artblocksRevenue_;
         additionalPayeePrimaryRevenue_ =
             (projectFunds *
                 projectIdToAdditionalPayeePrimarySalesPercentage[_projectId]) /
             100;
         artistRevenue_ = projectFunds - additionalPayeePrimaryRevenue_;
         // set addresses from storage
-        artblocksAddress_ = artblocksAddress;
+        artblocksAddress_ = artblocksPrimarySalesAddress;
```
ryley-o, line 1098: @jakerockland - do you prefer that I update this return var name to be `artblocksAddress` -> `artblocksPrimarySalesAddress`? would go into the minter contracts as well. I'm fine either way; since this function is getting primary sales splits, designating something as primary is a bit redundant, but we are doing it that way for additional payee, so could also do it for Art Blocks.

jakerockland: In isolation, I don't think I have a strong opinion on this one–though in alignment my comment here https://github.com/ArtBlocks/artblocks-contracts/pull/237#issuecomment-1213652892 I think we may want to maintain backwards compatibility for these royalty reading functions, in which case perhaps slight preference to not changing the naming convention here.

jakerockland:
> Also, I put a constraint for admin to never increase our basis points above 250 (2.5%). Input from @jakerockland would be much appreciated on that constraint!

I don't feel strongly about this, but would suggest that we probably don't need to add this constraint. I say this not because I think Art Blocks will ever raise our secondary royalty rate above 2.5%, but because ultimately secondary royalties enforced via social consensus, and because of this adding this constraint in code is a bit unnecessary here. That said, I think in isolation just thinking about Art Blocks, it does make sense. 

However, it likely doesn't make sense for our contracts in general (the eventual upgraded PBAB/Art Blocks Engine fork of this contract that will follow when we finish V3) as different partners may have different royalties based on the setup/release plans for their projects (e.g. I can imagine a partner project with a platform secondary royalty that starts very high and decays over time for a pfp type project). 

Because our partners may have different use cases here, and this is something that is a nice-to-have-socially-enforced thing anyway, my inclination is to not have this minor difference in contracts that is something we will need to remember to remove later on.

In any case, this is a long winded response for something that I don't feel super strongly about–but tl;dr I would say we probably can remove this...

jakerockland (approved):
Generally LGTM but a couple points that probably warrant more conversation.

contracts/GenArt721CoreV3.sol
```diff
@@ -20,13 +20,19 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
+        "artblocksSecondarySalesAddress";
     bytes32 constant FIELD_RANDOMIZER_ADDRESS = "randomizerAddress";
     bytes32 constant FIELD_ARTBLOCKS_CURATION_REGISTRY_ADDRESS =
         "curationRegistryAddress";
     bytes32 constant FIELD_ARTBLOCKS_DEPENDENCY_REGISTRY_ADDRESS =
         "dependencyRegistryAddress";
-    bytes32 constant FIELD_ARTBLOCKS_PERCENTAGE = "artblocksPercentage";
+    bytes32 constant FIELD_ARTBLOCKS_PRIMARY_SALES_PERCENTAGE =
+        "artblocksPrimaryPercentage";
+    bytes32 constant FIELD_ARTBLOCKS_SECONDARY_SALES_BPS =
```
mchrupcala, line 34: Tiny nit: should we use either Percentage or BPS just for consistency btwn the primary & secondary fields?

ryley-o:
> say this not because I think Art Blocks will ever raise our secondary royalty rate above 2.5%, but because ultimately secondary royalties enforced via social consensus, and because of this adding this constraint in code is a bit unnecessary here. 

@jakerockland Thanks for the thoughts, I've updated the constraint to be <= 100% (just for math reasons). I agree that keeping this open is okay, since royalties are enforced via social consensus.

contracts/GenArt721CoreV3.sol
mchrupcala: Oh I see, so there's a decimal difference between % and BPS. In that case the naming change wouldn't make sense here, nvm! But I'm curious why there's a need for BPS vs. % decimals and if this is worth choosing one over the other for V3 and standardizing?

contracts/GenArt721CoreV3.sol
ryley-o: @jakerockland 

Good point!. After not implementing this, then implementing this, I have a minor preference to keep the new naming to improve readability of V3 (mostly because everything for AB & artist additional payee is labeled as either primary or secondary).

re [#237 comment](https://github.com/ArtBlocks/artblocks-contracts/pull/237#issuecomment-1213652892), if we opt to keep the legacy getRoyaltiesData function, WDYT about adding a backwards-compatible getter function for `artblocksAddress` that returns `artblocksPrimarySalesAddress`, and document the getter function accordingly?

mchrupcala (approved):
Approved with a quick thought about the % and BPS formatting, but otherwise this looks awesome!

contracts/GenArt721CoreV3.sol
ryley-o: per discussion, will add a getter for `artblocksAddress`, then resolve this 😄

contracts/GenArt721CoreV3.sol
ryley-o: hmm, yes, I've stuck with percentage everywhere internally, except had to use BPS since our typical AB royalties are defined as 2.5%, which is not able to be expressed as a uint. Externally, when we support the royalty registry, the `getRoyalties` function converts everything to BPS.

I think my current implementation logic here is something like:
 - let's keep the numbers percentage where possible, since we already do that, and it is more easily understandable by artists/typical people, and also already integrates with our current artist/admin dashboard pages
 - let's use BPS for AB secondary sales royalties so we can defined it as 2.5% (i.e. 250 BPS)
 
I don't have super strong opinions here, other than want to balance internal conventions with backwards-compatibility. Also, we can always update names in follow-on PRs, so nothing is permanent until we deploy.

ryley-o: @mchrupcala - how do you feel about merging this as-is, and if we want some further discussion, we can add an issue?

ryley-o: per discussion, opened #240 for follow-on discussion of this topic 🚀

mchrupcala: > merging this as-is

Yes I totally agree, this shouldn't block merging in your PR 🚀 🚀  thanks Ryley!

PR #237, opened 2022-08-12
Integrate with Manifold's Royalty Registry directly on V3 core

ryley-o:
## Only merge after #236 

Integrate with Manifold's Royalty Registry directly on V3 core.

This adds a new function conforming to [IManifold](https://github.com/manifoldxyz/royalty-registry-solidity/blob/main/contracts/specs/IManifold.sol) on the V3 core that returns appropriate royalty recipient addresses and basis points (BPS). Note that we no longer need the `IArtBlocks` because we do not require a contract address input like our V1 and PBAB overrides do.

## Design Decisions
Opted to only returns recipients that have a non-zero BPS allocation. This is not required, and behaves differently than our current shim layer does. Overall, this simplifies code for integrators (no longer have to check for and skip a zero-address to remain efficient), at the cost of some added logic (still relatively simple) in the `getRoyalties` function. It also avoids an SLOAD of any payment addresses that aren't due royalties, and minimizes return data size, so is likely a bit more gas efficient.

Opted to remove legacy `getRoyaltyData` function from V3 core. The reasoning was to eliminate confusion for integrators (since that function silently did not include Art Blocks portion of royalties).

## Other
Source code for our V1 Royalty Registry shim layer was using the IV3 interface, and broke because the legacy `getRoyaltyData` function was removed from the V3 core. Dev comments were made, and the IV1 core interface is now used instead.

## Tests
- [x] Update existing tests
- [x] Write new tests for coverage

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -1018,32 +1018,57 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     }
 
     /**
-     * @notice Gets royalty data for token ID `_tokenId`.
+     * @notice Gets royalties for token ID `_tokenId`. This conforms to the
+     * IManifold interface designated in the Royalty Registry's
+     * RoyaltyEngineV1.sol contract.
+     * ref: https://github.com/manifoldxyz/royalty-registry-solidity
      * @param _tokenId Token ID to be queried.
-     * @return artistAddress Artist's payment address
-     * @return additionalPayee Additional payee's payment address
-     * @return additionalPayeePercentage Percentage of artist revenue
-     * to be sent to the additional payee's address
-     * @return royaltyFeeByID Total royalty percentage to be sent to
-     * combination of artist and additional payee
+     * @return recipients Array of royalty payment recipients
+     * @return bps Array of Basis Points (BPS) allocated to each recipient,
+     * aligned by index.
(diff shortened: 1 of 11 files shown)


jakerockland:
This generally LGTM, but I'd like to push on this a little bit.

> Opted to remove legacy getRoyaltyData function from V3 core. The reasoning was to eliminate confusion for integrators (since that function silently did not include Art Blocks portion of royalties).

I believe that there are other integrators that potentially depend on `getRoyaltyData` and will find it easier to rely on this method and its functionality being the same backwards-compatible, vs. having to handle a slightly different `getRoyalties`. I _think_ that this may be the case for Rarible IIRC.

Given that if we keep both, both `getRoyalties` and `getRoyaltyData` will return the same information _functionally_ (even if one omits zero-percentage info), I think your concerns around potential integrator confusion can just be mitigated via developer comments and if it is the case that an integrator choose the "wrong option" (the legacy one) this is generally still fine if they both provide the same functional information. 

For this reason, I'd suggest that we just keep both–as it seems inexpensive to do so IIUC?

jakerockland (approved):

ryley-o:
>I believe that there are other integrators that potentially depend on getRoyaltyData and will find it easier to rely on this method and its functionality being the same backwards-compatible, vs. having to handle a slightly different getRoyalties

per discussion, this seems reasonable to keep the backwards-compatible `getRoyaltyData` function, flagging it with @dev/@notice of what is and is not being included in the function. Prime use-case of this is probably Rarible.

PR #238, opened 2022-08-12
 Move V3 core's events to interface.

ryley-o:
## Only merge after #237 

Move V3 core's events to interface.

per [this discussion](https://github.com/ArtBlocks/artblocks-contracts/pull/204#discussion_r944847756) as part of #204 's review, having events in the interface is ideal if we think they will remain in future version updates.

This move the `ProposedArtistAddressesAndSplits` and `AcceptedArtistAddressesAndSplits` events to the V3 core's `IGenArt721CoreContractV3` interface.

PR #239, opened 2022-08-13
configure and update solidity-coverage hardhat plugin

ryley-o:
## Only merge after #238 

Configure and update solidity-coverage hardhat plugin.

This:
- updates our solidity-coverage plugin to the most recent version
- adds a coverage command to generate a coverage report
- configures our coverage via `.solcover.js`
  - skip appropriate .sol files
  - skip tests flagged with `@skip-on-coverage` [reference solution used](https://github.com/sc-forks/solidity-coverage/blob/master/docs/advanced.md#skipping-tests)
- flags tests related to gas-costs to be skipped
  - (solidity-coverage recommends this because it returns incorrect gas costs and causes invalid test failures)
    - see "distorts gas consumption" [on lib's repo readme](https://github.com/sc-forks/solidity-coverage)

Currently:
- V3 core at 97.2% coverage 🎉 
- Latest minters generally around 85-95% coverage

Overall, we don't have any major holes in our testing, but would like to improve to 100% prior to audit.
  
## Follow-on
- Auto-generate coverage reports for PRs (need to figure out the best way to do this)
- PR to reach 100% coverage on V3 core and most recent minters

### DIFF
--- a/.solcover.js
+++ b/.solcover.js
@@ -1,0 +1,28 @@
+module.exports = {
+  skipFiles: [
+    "interfaces/",
+    "legacy/",
+    "libs/",
+    "mock/",
+    "PBAB+Collabs/art-blocks-x-pace/",
+    "PBAB+Collabs/artcode/",
+    "PBAB+Collabs/bright-moments/",
+    "PBAB+Collabs/colors-and-shapes/",
+    "PBAB+Collabs/crypto-citizens/",
+    "PBAB+Collabs/doodle-labs/",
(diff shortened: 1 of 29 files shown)


ryley-o:
Cc @mchrupcala @lyaunzbe (Michael - I know we talked about adding this recently 😄 )

jakerockland (approved):
This is fantastic 🙏 🙏 🔥 🔥 

LGTM!

PR #242, opened 2022-08-16
Update details regarding minter projectMaxInvocations

ryley-o:
Update details regarding the optionally-populated minter `projectMaxInvocations`.

This updates minters regarding how they describe and populate their local mapping of `projectMaxInvocations`:
- mapping is described as optionally synced to the core contract
- `function setProjectMaxInvocations(projectId)` is no longer gated to any specific caller, because it no longer checks to see if maxInvocations has been increased on the core contract (V3 core does not ever allow maxInvocations to ever increase).
  - This is only enabled by the V3 core change of never allowing maxInvocations to increase, and should help reduce overhead (function must be gated to onlyWhitelisted on pre-V3 minters)
- comment about unchecked underflow improved

## Design Choices
- Opted to not change the name of mapping `projectMaxInvocations` on minters because it is an existing public mapping on V1-compatible minters. Better name would probably be `localProjectMaxInvocations` or something, but adding comments around the variable to better describe that it is optionally synced to the core contract for gas optimization purposes seemed sufficient.
- Opted to not change the function name `setProjectMaxInvocations` to `syncProjectMaxInvocations`, because existing function already integrates with our frontend today (I think, Cc @yoshiwarab if keeping this consistent in any way would help us).

### DIFF
--- a/contracts/V3_CHANGELOG.md
+++ b/contracts/V3_CHANGELOG.md
@@ -54,3 +54,5 @@ _This document is intended to document and explain the Art Blocks Core V3 change
   - Minimize costly SLOAD operations & re-organize logic to minimize gas usage
   - Optimize mint function signature to reduce gas usage
 - Update randomizer interface for V3 core contract
+- Only allow artists to reduce the number of maximum invocations
+  - Remove any unnecessary minter logic accordingly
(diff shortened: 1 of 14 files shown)


jakerockland (approved):
LGTM–agree with the rationale you outlined for not changing naming conventions. 👍 🙏

PR #243, opened 2022-08-16
Broadcast support of IManifold interface (royalty registry)

ryley-o:
## only merge after #242 

Broadcast support of IManifold interface (used for the Royalty Registry) via the IERC165 `supportsInterface` function.

This is not required for integration with the Royalty Registry. It is, however, good practice to broadcast support of interfaces that are expected points of integration, especially for third-party integrations.

### DIFF
--- a/.solcover.js
+++ b/.solcover.js
@@ -1,0 +1,28 @@
+module.exports = {
+  skipFiles: [
+    "interfaces/",
+    "legacy/",
+    "libs/",
+    "mock/",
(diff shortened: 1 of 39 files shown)


jakerockland:
Just curious, was this one actually dependent on #242 ?

jakerockland (approved):
LGTM

ryley-o:
> Just curious, was this one actually dependent on #242 ?

Ah, yeah these were probably quite independent from each other 👍

PR #247, opened 2022-08-17
Improve gas tests, improve optimizer settings

ryley-o:
Improve gas tests and improve compiler optimizer settings.

This does not affect the audit because no contracts are touched.

This
- improves gas tests by averaging mint costs over 15 transactions to eliminate noise.
- changes our solc compiler optimizer settings from 100 runs to 25 runs to find some gas savings for purchases. 25 runs was selected based on the table below, because it has 80% of the reduction of 10%, but isn't on the edge of the 0.43% increase.

>note: these results are somewhat unintuitive, as one would generally expect higher runs to reduce costs of calling. In this case, we are only looking at the mint transaction cost, so global optimizer settings could have unexpected results.

| Optimizer Runs | Contract Size (kb) | Typical V3 gas to mint\*  | % gas reduced relative to Optimizer: 100 |
| -------------- | ------------------ | ------------------------- | ---------------------------------------- |
| 0              | 36.005             | N/A - too large to deploy | \-                                       |
| 1              | 19.521             | 147165                    | 0.43%                                    |
| 5              | 19.521             | 147165                    | 0.43%                                    |
| 8              | 19.537             | 146309                    | \-0.15%                                  |
| 10             | 19.537             | 146309                    | \-0.15%                                  |
| 25             | 19.573             | 146345                    | \-0.13%                                  |
| 50             | 19.639             | 146414                    | \-0.08%                                  |
| 100            | 19.778             | 146529                    | 0.00%                                    |
| 200            | 19.998             | 146460                    | \-0.05%                                  |
| 500            | 21.381             | 146529                    | 0.00%                                    |
| 1000           | 23.199             | 146574                    | 0.03%                                    |
| 2500           | 23.832             | 146544                    | 0.01%                                    |

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -41,7 +41,7 @@ module.exports = {
         settings: {
           optimizer: {
             enabled: true,
-            runs: 100,
+            runs: 25,
           },
         },
       },
--- a/test/core/GenArt721CoreV1_Integration.tests.ts
+++ b/test/core/GenArt721CoreV1_Integration.tests.ts
@@ -104,7 +104,7 @@ describe("GenArt721CoreV1 Integration", async function () {
       ).to.equal(ethers.utils.parseEther("0.8971167"));
       expect(
         (await this.accounts.user.getBalance()).sub(ownerBalance)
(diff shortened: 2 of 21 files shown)


jakerockland (approved):
Fantastic–TY @ryley-o

PR #248, opened 2022-08-17
Pack Project struct to reduce gas costs

ryley-o:
Pack Project struct where possible to reduce gas costs

## Gas impacts
Mint costs are slightly reduced. Representative minter cost updates are shown in the table below.

| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 146345 | 142557 | -2.6% (cheaper) |
| MinterDAExpV2_V3Core | 158183 | 154395 | -2.4% (cheaper) |

## Audit impacts
- No interfaces are affected by this code
- Would still be good to Cc ongoing auditors of this update

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -82,6 +82,7 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
         uint24 invocations;
         uint24 maxInvocations;
         uint24 scriptCount;
+        // max uint64 ~= 1.8e19 sec ~= 570 billion years
         uint64 completedTimestamp;
         bool active;
         bool paused;
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -18,6 +18,7 @@ import "@openzeppelin-4.7/contracts/token/ERC721/ERC721.sol";
  */
 contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     uint256 constant ONE_MILLION = 1_000_000;
+    uint24 constant ONE_MILLION_UINT24 = 1_000_000;
     uint256 constant FOUR_WEEKS_IN_SECONDS = 2_419_200;
 
     // generic platform event fields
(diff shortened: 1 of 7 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -78,6 +79,12 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
     IAdminACLV0 public adminACLContract;
 
     struct Project {
+        uint24 invocations;
+        uint24 maxInvocations;
+        uint24 scriptCount;
+        uint64 completedTimestamp;
```
jakerockland, line 85: nit: can we please make an inline comment clarifying that `64` bytes is plenty for the life of all of our existences and many generations (all of them) to come?

jakerockland (approved):
This LGTM.

On a related note, curious your thoughts on storing some of the `string` values as raw `bytes32` instead of `string` for similar byte packing/gas optimization reasons?

contracts/GenArt721CoreV3.sol
ryley-o: 570 billion years 😄 updated!

jakerockland: Haha thank you.

ryley-o:
>On a related note, curious your thoughts on storing some of the string values as raw bytes32 instead of string for similar byte packing/gas optimization reasons?

Packing will only save gas when combining multiple sub-32-byte items into the same 32-byte storage slot. So we would only pack if we used smaller bytes, like two bytes16 in a row.

We could definitely change any of the string parameters to any bytes size, but we would limit the length of the field. If, for example, we changed license from `string` to `bytes32`, the artist would be limited to roughly 31 characters (depending on the characters, some take >1 byte). Any time a `bytes32` is able to be used, there is an associated efficiency gain because strings have to use an additional storage slot to record the number of slots they use (since they are unbounded in size, essentially a bytes array).

tldr - if we think any of the fields that are currently strings can always be < ~30 characters, we could switch to reduce gas costs when an artist is populating the items. If we think we can limit several fields to < ~15 characters, we could pack those values into the same slot, potentially finding more gas reductions. These types of changes may impact the calls coming from the frontend.

PR #249, opened 2022-08-17
Remove Enumerable conformance for PBAB and PRTNR

jakerockland:
~~This PR is very much just a draft/proposal.~~

~~That said,~~ I don't think that there is a meaningful enough divergence in functionality when removing `Enumerable` conformance that we need to treat this as a new version for our templates, but open to feedback here to the contrary.

There is a big enough gas savings from removing this in V3 that I am inclined to remove from PBAB/PRTNR templates **now** rather than waiting until V4 (the new PBAB/PRTNR standard/template that we should pursue as a follow on once V3 is 🚢 'd)~~, but would love additional input from @ryley-o @yoshiwarab @lyaunzbe before moving this out of draft mode.~~

~~WDYT folks?~~

### DIFF
--- a/contracts/PBAB+Collabs/GenArt721CoreV2_PBAB.sol
+++ b/contracts/PBAB+Collabs/GenArt721CoreV2_PBAB.sol
@@ -6,7 +6,7 @@ import "../interfaces/0.8.x/IGenArt721CoreV2_PBAB.sol";
 
 import "@openzeppelin-4.5/contracts/utils/Strings.sol";
 
-import "@openzeppelin-4.5/contracts/token/ERC721/extensions/ERC721Enumerable.sol";
+import "@openzeppelin-4.5/contracts/token/ERC721/ERC721.sol";
 
 pragma solidity 0.8.9;
 
@@ -14,7 +14,7 @@ pragma solidity 0.8.9;
  * @title Powered by Art Blocks ERC-721 core contract.
  * @author Art Blocks Inc.
  */
-contract GenArt721CoreV2_PBAB is ERC721Enumerable, IGenArt721CoreV2_PBAB {
+contract GenArt721CoreV2_PBAB is ERC721, IGenArt721CoreV2_PBAB {
     /// randomizer contract
     IRandomizer public randomizerContract;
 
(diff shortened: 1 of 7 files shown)


jakerockland:
Other rationale for doing this now (vs. waiting for V4) is coalescing with changes in https://github.com/ArtBlocks/artblocks-contracts/pull/204, which will be merged soon. 😄

ryley-o:
Yeah, I don't think this affects anything downstream in our pipeline, so kind of an easy yes from me.
If we have to make any non-negligible changes to our frontend, this might be not worth it, but I suspect nothing will change at all.
If we want to support existing partners upgrading to this core, we will need a constructor arg or something to indicate first project ID.

jakerockland:
> If we want to support existing partners upgrading to this core, we will need a constructor arg or something to indicate first project ID. (would also like input from the Bens on this!)

Great point, we probably should do that either in this PR or follow up. Going to wait to get more feedback from Ben or Ben before updating this PR but that SGTM.

yoshiwarab:
Yeah, I don't see any issue with removing enumerable conformance straight away. The only thing that feels somewhat useful is being able to check the total supply to sanity check our indexing quickly. Still, there are definitely plenty of other sanity checks we can do instead.

yoshiwarab:
> If we want to support existing partners upgrading to this core, we will need a constructor arg or something to indicate first project ID. (would also like input from the Bens on this!)

Also definitely think we should add this and wondering if we'd also like to add some mechanism to v3 to prevent adding new projects once we've switched over to a new core contract.

jakerockland:
> wondering if we'd also like to add some mechanism to v3 to prevent adding new projects once we've switched over to a new core contract.

This seems like a good idea–low cost to do so.

ryley-o:
> Also definitely think we should add this and wondering if we'd also like to add some mechanism to v3 to prevent adding new projects once we've switched over to a new core contract.

Ready for your review ser! https://github.com/ArtBlocks/artblocks-contracts/pull/255 😄

jakerockland:
Tests are updated 😄 

Ready for review/merge @ryley-o

ryley-o (approved):
LGTM 🎉

PR #250, opened 2022-08-17
pack artist financials into struct

ryley-o:
Documenting this to note that I don't see any opportunity for gas optimization by packing artist financials.

Basically, we only need to load a single value when splitting payments for either primary or secondary, so putting everything into the same slot doesn't really buy us anything except a bunch of casting between uint256 and uint8.

@jakerockland please close this PR once you take a peek at the summary, for posterity.

## Gas impacts
Mint costs are not reduced. Representative minter cost updates are shown in the table below.

| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 142557 | 142563 | +/- |
| MinterDAExpV2_V3Core | 154395 | 154401 | +/- |

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -100,18 +100,21 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
 
     mapping(uint256 => Project) projects;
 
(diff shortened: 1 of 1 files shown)


jakerockland:
Thank you very much for filing this PR for posterity–agree with your rationale and closing this out.

PR #252, opened 2022-08-18
Allow V3 to initialize nextProjectId to non-zero value

ryley-o:
Allow V3 to initialize nextProjectId to non-zero value.

We expect the first V3 core's projectId to be immediately after the last V1 core value. This allows the initial value of `nextProjectId` to be initialized in the core contract's constructor.

## Design Choices
- Initialize `nextProjectId` via a constructor input arg `startingProjectId` to avoid hard-coding the initial value of `nextProjectId` in the contract source code
  - This helps reduce the complexity of tests, allows for easy version rolling, etc.
- A `PlatformUpdated` event is emitted in the constructor indicating that `nextProjectId` has been updated
  - This seems more concise vs. adding a more-generic event with a field like "contractDeployed" and implying the subgraph should know what to do
  - Not including the value of nextProjectId in the event opens us up to a state mis-match if a project were added in the same block that the contract is deployed in
    - Recommend that we update `nextProjectId` in the subgraph any time a new project is added (since the new projectId is included in that event), which prevents state mis-match from ever persisting after an entire block is processed.
    
## Audit Impacts
Recommend that this is included in our audit.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -200,14 +200,14 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
      * @param _randomizerContract Randomizer contract.
      * @param _adminACLContract Address of admin access control contract, to be
      * set as contract owner.
-     * @param _nextProjectId The initial next project ID.
+     * @param _startingProjectId The initial next project ID.
      */
     constructor(
         string memory _tokenName,
         string memory _tokenSymbol,
         address _randomizerContract,
         address _adminACLContract,
-        uint256 _nextProjectId
+        uint256 _startingProjectId
     ) ERC721(_tokenName, _tokenSymbol) {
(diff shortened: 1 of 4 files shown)


ryley-o:
>Recommend that we update nextProjectId in the subgraph any time a new project is added (since the new projectId is included in that event), which prevents state mis-match from ever persisting after an entire block is processed
Cc @yoshiwarab

jakerockland:
Would suggest that `initialProjectId` may be a more idiomatic nomenclature than `nextProjectId`.

jakerockland (approved):

ryley-o:
>Would suggest that initialProjectId or startingProjectId may be a more idiomatic nomenclature than nextProjectId

Good call! Updated the constructor arg to be `startingProjectId`. Kept the event field as `nextProjectId`, since that will be the public variable the subgraph will have to reach out to to sync.

yoshiwarab:
Awesome, I much prefer this to hardcoding the initial nextProjectId! Event names/arg names all sound good to me

PR #255, opened 2022-08-18
Enable admin to forbid new projects on V3 core

ryley-o:
Enable admin to forever forbid new projects from being added to V3 core.

A new function is added, gated to admin, that can forever forbid new projects from ever being added to the V3 core contract.

## Design Choices
This also packs `_nextProjectId` and the new bool into the same storage slot by internally limiting `_nextProjectId` to a uint248. Previous interface maintained via a getter function that returns `nextProjectId` as uint256.


jakerockland (approved):
🙏 🔥

PR #259, opened 2022-08-19
V2 PBAB/PRTNR non-zero startingProjectId

ryley-o:
Update V2 PBAB and PRTNR to enable non-zero startingProjectId.

This is follow-on work somewhat related to https://github.com/ArtBlocks/artblocks-contracts/pull/249/files, and also related to https://github.com/ArtBlocks/artblocks-contracts/pull/252.

This keeps startingProjectId as a uint256 to minimize change relative to the existing V2 PBAB and PRTNR contracts. There is also no opportunity to reduce gas by packing, because we don't have any bools nearby (different than the new PBAB flex contract, where there is opportunity).

Tests added for new functionality, PBAB deployer script templates updated for additional constructor arg.


jakerockland (approved):
TYVM for sending this–will plan to merge this and then fix the gas values in #249 after re-basing (they were correct and then got broken after re-basing to `main`, so going to let this PR land first and then will re-base mine and fix).

PR #261, opened 2022-08-19
bytes16 hash

ryley-o:
(for posterity only) Update V3 core to use bytes16 hash instead of bytes32 hash.

## Result
the EVM likes working in 32-byte chunks. I could only get this to add gas instead of reduce it. Not recommended unless someone can work more magic than my somewhat haste attempts (maybe some assembly would work here?). Overall, does not seem worth our effort.

Before: Fixed price minter cost: 137855 gas
After: Fixed price minter cost: 138353 gas

Cc @jakerockland , immediately closing this PR.

### DIFF
--- a/contracts/BasicRandomizerV2.sol
+++ b/contracts/BasicRandomizerV2.sol
@@ -18,17 +18,19 @@ contract BasicRandomizerV2 is IRandomizerV2, Ownable {
     }
 
     // When `genArt721Core` calls this, it can be assured that the randomizer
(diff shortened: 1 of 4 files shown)


jakerockland:
For posterity: @ryley-o and I had some fun explorative convo about trying another packing approach, using `bytes12` for token hash and squeezing it to sit alongside the 20-byte `address` for token owners. He's going to explore a PR or two that take this approach for deeper consideration, but from our initial digging this seems quite promising.

ryley-o:
update: see improved version with much gas savings by packing hash with token owner in #250

PR #263, opened 2022-08-20
Pack hash seed with owner

ryley-o:
Pack hash seed with token owner to eliminate storage slot accessed during mint transaction (assuming atomic randomizer callback).

Based on brainstorming session w/ @jakerockland, this uses a packed struct to store token owner and a bytes12 hashSeed into the same storage slot. The net result is an entire storage slot is not longer accessed or written to during the mint flow, resulting in gas cost reduction for users.

Note that a bytes12 hashSeed provides sufficient theoretical entropy for our token generation; the chance of a hash seed collision of any two tokens in a 1-million token project (maximum possible for the V3 core contract) is less than 1 in 100 million.

This specific implementation forks OpenZeppelin's ERC721 v4.7.1 contract to provide an internally-accessible packed struct. See further discussion in the following section for alternate designs considered to achieve the same end result of packing token hash and token owner into the same storage slot.

## Design Considerations
Using a struct to pack owner address and bytes12 hash seed is a safe way to write/read/parse the two values in storage. Assembly could also be used to access this storage slot and read/write hash seed directly, but our team currently isn't completely sure if some unintended effects could be encountered unless we also managed reading/writing owner via assembly. The simpler implementation for now is to simply fork OpenZeppelin's ERC721 and utilize struct packing.

Inheriting from ERC721 directly, defining a mapping to a packed struct in the V3 core contract, and overriding methods that interact with ERC721's `_owners` method would be another strategy of achieving the gas reductions in this PR. Unfortunately, ERC721's `_mint` function also increments the ERC721 `private _balances` variable, and `private` variables are not accessible from inheriting contracts. This means that a variable representing balances would also have to be defined on the V3 core, and all functions on ERC721 that interact with `_balances` would also have to be overridden and defined on the V3 core contract. All things considered, it was simpler to just fork the ERC721 contract and design `ERC721_PackedHashSeed` directly.

## Gas impacts
Mint costs are reduced, as shown in the table below.

_(All gas costs assume project 3 mints avg(501-516) of 1000)_
| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 137855 | 116106 | -15.8% |
| MinterSetPriceERC20V2_V3Core | 138014 | 116265 | -15.8% (cheaper) |
| MinterDAExpV2_V3Core | 147255 | 125506 | -14.8% (cheaper) |
| MinterDALinV2_V3Core | 147353 | 125604 | -14.8% (cheaper) |
| MinterMerkleV1_V3Core | 153003 | 131254 | -14.2% (cheaper) |
| MinterHolderV1_V3Core | 144497 | 122748 | -15.1% (cheaper) |

## Audit
This is required to be included in third party audits of the V3 contract.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -307,11 +307,11 @@ contract GenArt721CoreV3 is
      * May only be called for tokens that have not already been assigned a
      * non-zero hash.
      * @param _tokenId Token ID to set the hash for.
-     * @param _hash Hash seed set for the token ID. Only last 12 bytes will be
-     * used.
+     * @param _hashSeed Hash seed to set for the token ID. Only last 12 bytes
+     * will be used.
      * @dev gas-optimized function name because called during mint sequence
      */
-    function setTokenHash_8PT(uint256 _tokenId, bytes32 _hash)
+    function setTokenHash_8PT(uint256 _tokenId, bytes32 _hashSeed)
         external
         onlyValidTokenId(_tokenId)
     {
@@ -326,7 +326,7 @@ contract GenArt721CoreV3 is
             ownerAndHashSeed.hashSeed == bytes12(0),
             "Token hash already set"
         );
-        ownerAndHashSeed.hashSeed = bytes12(_hash);
+        ownerAndHashSeed.hashSeed = bytes12(_hashSeed);
     }
 
     /**
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -10,13 +10,17 @@ import "./interfaces/0.8.x/IManifold.sol";
 
 import "@openzeppelin-4.7/contracts/utils/Strings.sol";
 import "@openzeppelin-4.7/contracts/access/Ownable.sol";
-import "@openzeppelin-4.7/contracts/token/ERC721/ERC721.sol";
+import "./libs/0.8.x/ERC721_PackedHashSeed.sol";
 
 /**
  * @title Art Blocks ERC-721 core contract, V3.
  * @author Art Blocks Inc.
  */
-contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
+contract GenArt721CoreV3 is
+    ERC721_PackedHashSeed,
+    Ownable,
+    IGenArt721CoreContractV3
+{
     uint256 constant ONE_MILLION = 1_000_000;
     uint24 constant ONE_MILLION_UINT24 = 1_000_000;
     uint256 constant FOUR_WEEKS_IN_SECONDS = 2_419_200;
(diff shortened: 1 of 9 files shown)


ryley-o:
@jakerockland diffchecker here: https://www.diffchecker.com/ONezw5Y0

contracts/GenArt721CoreV3.sol
```diff
@@ -300,27 +302,31 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
             "Only randomizer may set"
         );
         require(
-            tokenIdToHash[_tokenId] == bytes32(0),
-            "Token hash already set."
+            ownerAndHashSeed.hashSeed == bytes12(0),
+            "Token hash already set"
         );
-        tokenIdToHash[_tokenId] = _hash;
+        ownerAndHashSeed.hashSeed = bytes12(_hash);
```
jakerockland, line 329: I think we may want to output an event here that contains the output of `tokenIdToHash(_hash)`, so that we can more readily index the token hash itself, unless we do so somewhere else that I'm overlooking?

cc @yoshiwarab @lyaunzbe

contracts/GenArt721CoreV3.sol
```diff
@@ -300,27 +302,31 @@ contract GenArt721CoreV3 is ERC721, Ownable, IGenArt721CoreContractV3 {
...
      * May only be called by the current randomizer contract.
      * May only be called for tokens that have not already been assigned a
      * non-zero hash.
      * @param _tokenId Token ID to set the hash for.
-     * @param _hash Hash to set for the token ID.
+     * @param _hash Hash seed set for the token ID. Only last 12 bytes will be
+     * used.
      * @dev gas-optimized function name because called during mint sequence
      */
     function setTokenHash_8PT(uint256 _tokenId, bytes32 _hash)
```
jakerockland, line 314: WDYT re: renaming `_hash` to be the more explicit/idiomatic `_hashSeed` instead given this updated paradigm?

jakerockland (approved):
Couple areas for a bit more discussion (happy to hop on a call to hash this out as well in the interest of expedience), but generally LGTM! 💜 

Diffchecker link is super super appreciated–TYVM @ryley-o

contracts/GenArt721CoreV3.sol
ryley-o: I did consider this, but opted to not emit the event from our core contract via the following logic:
- When using a randomizer that assigns token hash during the mint tx, we can key our subgraph off of the `Mint` event to populate token hash on that block
- If we were to change to an async randomizer in the future, the randomizer could emit a `TokenHashPopulatedAsync` event from the future minter, giving our subgraph something else to key off of.

The benefit of this solution is that we don't have to pay the extra gas during the mint tx for an event. There isn't much of a downside, other than some minor subgraph complexity if we upgrade in the future (although, would be a very minor amount of complexity). I suppose some could feel strongly that the event should be something emitted by the core contract instead of the randomizer. I don't have a strong preference for where events are emitted from though, they are really only intended for off-chain indexing.

jakerockland: This rationale tracks to me–so long as we're comfortable with the complexity on the subgraph side needed for launching this V3 contract with our infra+frontend integration. Sounds like you're comfortable with this, which I think is totally fair/reasonable, so SGTM to move forward without added event on Core 👍

contracts/GenArt721CoreV3.sol
ryley-o: good call! updated the parameter name to `_hashSeed`. Note that I did consider re-naming the function, but I like keeping this name (and the bytes32 input) as an interface, allowing the core contract to choose how to store and turn the hash seed into a given token's hash.

PR #264, opened 2022-08-22
Pack financial data on V3 core

ryley-o:
Pack financial data on V3 core to improve gas efficiency.

This minimizes the qty of cold storage slots accessed by packing primary and royalty percentages with addresses. This is an improved version of #250, based on things I've learned in the last week.

The percentages quoted below assume no additional payee for primary sales is set up. There would be slightly more benefit if our benchmark included additional payee payments (so this is a conservative estimate of the benefit).

## Gas impacts
Mint costs are reduced. Representative minter cost updates are shown in the table below.

| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 116106 | 114112 | -1.7% |
| MinterDAExpV2_V3Core | 125506 | 123512 | -1.6% |

## TESTS

- [x] Gas tests updated
- [x] 100% coverage on V3 core

## Audit
This should be included in third party audits of the V3 core contract.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -107,26 +107,30 @@ contract GenArt721CoreV3 is
 
     mapping(uint256 => Project) projects;
 
-    // All financial functions are stripped from struct for visibility
-    mapping(uint256 => address payable) public projectIdToArtistAddress;
-    mapping(uint256 => address payable)
-        public projectIdToAdditionalPayeePrimarySales;
-    mapping(uint256 => uint256)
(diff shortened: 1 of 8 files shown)


jakerockland (approved):
This is fantastic, and also something I think is worth noting in the PR description that should be "in scope of audit". I think we basically just want to pick a new commit hash to work off of re: auditing with the latest and greatest, as nothing has functionally changed since scoping out audit, all just internal details–though ones we do def want to capture in our audit review.

ryley-o:
>I think is worth noting in the PR description that should be "in scope of audit"
 
Good call, updated!

PR #265, opened 2022-08-22
Storage pointers on V3 core

ryley-o:
Add more storage pointers on V3 core to improve code readability and gas efficiency.

This is mostly for improved readability of code, and gas efficiency of minting is just an additional benefit.

## Gas impacts
Mint costs are very slightly reduced. Representative minter cost updates are shown in the table below.

| --- | --- | --- | --- |
| MinterSetPriceV2_V3Core | 114112 | 113916 | -0.2% |
| MinterDAExpV2_V3Core | 123512 | 123316 | -0.2% |

## Tests
This is a refactor - no new tests required. Gas tests updated.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -249,35 +249,35 @@ contract GenArt721CoreV3 is
     ) external returns (uint256 _tokenId) {
         // CHECKS
         require(msg.sender == minterContract, "Must mint from minter contract");
-
+        Project storage project = projects[_projectId];
         // load invocations into memory
-        uint24 invocationsBefore = projects[_projectId].invocations;
+        uint24 invocationsBefore = project.invocations;
         uint24 invocationsAfter;
         unchecked {
             // invocationsBefore guaranteed <= maxInvocations <= 1_000_000,
             // 1_000_000 << max uint24, so no possible overflow
             invocationsAfter = invocationsBefore + 1;
         }
-        uint24 maxInvocations = projects[_projectId].maxInvocations;
+        uint24 maxInvocations = project.maxInvocations;
 
         require(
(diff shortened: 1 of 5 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -249,35 +249,35 @@ contract GenArt721CoreV3 is
     ) external returns (uint256 _tokenId) {
         // CHECKS
         require(msg.sender == minterContract, "Must mint from minter contract");
-
+        Project storage project = projects[_projectId];
```
lindsgil, line 252: Is it possible to get into a state where project is undefined here?

lindsgil (approved):

contracts/GenArt721CoreV3.sol
ryley-o: Kind of a funny quirk about solidity, mappings can be thought of as "virtually initialized" such that every possible key exists and is mapped to a value whose byte-representation is all zeros (a types default value). So we won't ever have an undefined key.

That being said, what if our minter tried to mint on a project that has not yet been added? When the code below loads `project.invocations` from storage, it will return the default value for `uint24`, which is zero. Similarly, when `project.maxInvocations` is loaded from storage, it will return the default value of zero. Finally, the requirement to have `invocationsBefore < maxInvocations` will fail, preventing the uninitialized project from being minted. Note that the default value for ` bool` is false, so later checks (e.g. `require(project.active,...)` would fail.

Here is a good reference about mappings: https://docs.soliditylang.org/en/develop/types.html?highlight=mapping#mapping-types

lindsgil: awesome, thank you for the explanation and the resources! excited to learn more :)

jakerockland (approved):
LGTM

PR #269, opened 2022-08-23
Combine scriptType and scriptVersion on V3 core

ryley-o:
Combine scriptType and scriptVersion on V3 core.

This aligns the V3 core with our dependency registry plans.

## Design Choices
Opted to validate that the artist or admin's input library@version string contains exactly one `@` symbol when calling `updateProjectScriptType` function. This will help ensure the on-chain data include version for all projects.

The costs associated with the on-chain validation of the `@` symbol are:
With validation: 
```
"p5js@v1.2.3" = 54k gas (updated from previously 52k now that we check for exactly 1 occurrence)
"long-library-name-search@v1.2.3" = 55k gas (updated from previously 59k now that we check for exactly 1 occurrence, and added some unchecked maths)
```
Without validation:
```
"p5js@v1.2.3" = 50k gas
```

Additionally, since most of the helper functions dealt with bytes32, a utility library was created and used on bytes32. This will allow future versions of our contracts to also implement the utility functions without having to copy and paste the functions in more than a single place. It also makes code in the V3 core contact very clean and easy to read.

## Tests
Tests are updated for new function interfaces, and also added to test the new on-chain validation functionality.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -312,11 +312,11 @@ contract GenArt721CoreV3 is
      * May only be called for tokens that have not already been assigned a
      * non-zero hash.
      * @param _tokenId Token ID to set the hash for.
-     * @param _hash Hash seed set for the token ID. Only last 12 bytes will be
-     * used.
+     * @param _hashSeed Hash seed to set for the token ID. Only last 12 bytes
+     * will be used.
      * @dev gas-optimized function name because called during mint sequence
      */
-    function setTokenHash_8PT(uint256 _tokenId, bytes32 _hash)
+    function setTokenHash_8PT(uint256 _tokenId, bytes32 _hashSeed)
         external
         onlyValidTokenId(_tokenId)
     {
@@ -331,7 +331,7 @@ contract GenArt721CoreV3 is
             ownerAndHashSeed.hashSeed == bytes12(0),
             "Token hash already set"
         );
-        ownerAndHashSeed.hashSeed = bytes12(_hash);
+        ownerAndHashSeed.hashSeed = bytes12(_hashSeed);
     }
 
     /**
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -249,35 +249,35 @@ contract GenArt721CoreV3 is
     ) external returns (uint256 _tokenId) {
         // CHECKS
         require(msg.sender == minterContract, "Must mint from minter contract");
-
+        Project storage project = projects[_projectId];
         // load invocations into memory
-        uint24 invocationsBefore = projects[_projectId].invocations;
+        uint24 invocationsBefore = project.invocations;
         uint24 invocationsAfter;
         unchecked {
             // invocationsBefore guaranteed <= maxInvocations <= 1_000_000,
             // 1_000_000 << max uint24, so no possible overflow
             invocationsAfter = invocationsBefore + 1;
         }
-        uint24 maxInvocations = projects[_projectId].maxInvocations;
+        uint24 maxInvocations = project.maxInvocations;
 
         require(
             invocationsBefore < maxInvocations,
             "Must not exceed max invocations"
         );
         require(
-            projects[_projectId].active ||
+            project.active ||
                 _by == projectIdToFinancials[_projectId].artistAddress,
             "Project must exist and be active"
         );
         require(
-            !projects[_projectId].paused ||
+            !project.paused ||
                 _by == projectIdToFinancials[_projectId].artistAddress,
             "Purchases are paused."
         );
 
         // EFFECTS
         // increment project's invocations
-        projects[_projectId].invocations = invocationsAfter;
+        project.invocations = invocationsAfter;
         uint256 thisTokenId;
         unchecked {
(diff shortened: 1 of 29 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -808,34 +808,30 @@ contract GenArt721CoreV3 is
...
+        bytes32 _scriptTypeAndVersion
     )
         external
         onlyUnlocked(_projectId)
         onlyArtistOrAdminACL(_projectId, this.updateProjectScriptType.selector)
     {
-        projects[_projectId].scriptType = _scriptType;
-        projects[_projectId].scriptTypeVersion = _scriptTypeVersion;
+        Project storage project = projects[_projectId];
+        // TODO - require @ symbol in _scriptTypeAndVersion
```
jakerockland, line 833: This PR generally LGTM but we probably shouldn't merge until this part is implemented as well–WDYT?

jakerockland (changes requested):

contracts/GenArt721CoreV3.sol
ryley-o: @jakerockland Yes this one was not quite ready for review yet. The on-chain validation and PR description are updated, and are ready for review now! 😄

ryley-o: ohh I think something automatically triggered your review request? 🤔 sorry it was early

jakerockland: Ah–I think it is because this wasn't a draft maybe, so auto-reviewer assigned?

contracts/GenArt721CoreV3.sol
```diff
@@ -817,21 +819,24 @@ contract GenArt721CoreV3 is
...
     {
         Project storage project = projects[_projectId];
-        project.scriptType = _scriptType;
-        project.scriptTypeVersion = _scriptTypeVersion;
+        // require @ symbol in _scriptTypeAndVersion
+        require(
+            _scriptTypeAndVersion.containsCharacter(AT_CHARACTER_CODE),
+            "must contain @"
+        );
+        project.scriptTypeAndVersion = _scriptTypeAndVersion;
```
jakerockland, line 839: I wonder if we want to require that it contains _only one_ `@`?

jakerockland (approved):

contracts/GenArt721CoreV3.sol
ryley-o: yeah must have been not a draft, oops!

contracts/GenArt721CoreV3.sol
ryley-o: Would this potentially conflict with npm's convention? e.g. `@mattheath/p5js-node` is a valid npm package@1.3.7

jakerockland: Hrm–that is valid... though don't know we'd ever support an arbitrary NPM package such that this would be relevant. I'm worried if we don't enforce _just one `@`_, that we put ourselves in a weird spot wrt to parsing out the type and version separately if/as we need to in our backend infra. Though based on recent convos with @lyaunzbe , this may be less important as a function of time in the future.

ryley-o: true true... Unless we make "allowed qty of @ symbols" an admin configurable qty, I'm slightly inclined to lean on the more permissive side of things for on-chain validation since it is not updatable.  We could always handle any one-off quirks in subgraph. That being said, I don't have very strong opinions on this, and we can work around any one-off quirks different ways (for several reasons, e.g. we overflow 32-byte limit).

Want to make this require only a single `@`?

jakerockland: Hrm–yeah my thinking is to just make require a single `@` to have this map to our actual current assumptions–but would love a gut check from @yoshiwarab or @lyaunzbe as well before we ship (or both).

lyaunzbe: i think requiring one `@` is fine here. as mentioned above, npm uses that first `@` for scoping and I do think that's only relevant in the context of npm ultimately

lyaunzbe: here's a specific example also: 
```
https://www.npmjs.com/package/@babylonjs/core 
https://cdnjs.cloudflare.com/ajax/libs/babylonjs/5.20.0/babylon.js
```
even though this dependency does use scoping, it only matters within the context of the npm pkg

yoshiwarab: Yeah I think one @ makes sense to me and will make it a little safer to break the dependency type into type and version components.

ryley-o: SGTM! updated 🙏

ryley-o:
(just to note, I did fix a bug that could have returned false positives. Not sure why I was thinking two characters per byte 😓 but it is good now 👍 )

ryley-o:
okay, another minor update to ensure multi-byte utf-8 characters are properly handled. by checking <0x80 we are guaranteed that a byte in a solidity byte-encoded string is a single-byte character.

ryley-o:
@lyaunzbe gentle ping for an approval on this so I can merge 👍

lyaunzbe (approved):
looks good 🚀

PR #270, opened 2022-08-24
Prototype contract as script storage

ryley-o:
⚠️ This is a prototype, do not merge. Much more investigation required.

Many things to do before we decide if we want to implement this.

## Gas impacts
Very large script upload gas improvement. Example script upload cost impacts are shown in the table below.

| Gas Test | Previous gas to upload script | New gas used to upload script | % change |
| --- | --- | --- | --- |
| GenArt721CoreV3: Upload Squiggle Script | 2_822_470 | 943_967 | -66.6% |
| GenArt721CoreV3: Upload Skulptuur-like Script | 4_569_796 | 1_493_682 | -67.3% |

---
  - https://app.asana.com/0/0/1202817061735237


jakerockland:
Very interesting–would be awesome to profile how this impacts max upload size as well per "chunk".

ryley-o:
>Very interesting–would be awesome to profile how this impacts max upload size as well per "chunk".

Do we have a way that we currently calculate the max size per "chunk"?

jakerockland:
> > Very interesting–would be awesome to profile how this impacts max upload size as well per "chunk".
> 
> Do we have a way that we currently calculate the max size per "chunk"?

I don't think so explicitly, but it will be basically defacto limited to what the gas limit is per block–as IIUC this is what impacts the amount that can be written into storage w/ a string in one "chunk"?

ryley-o:
>I don't think so explicitly, but it will be basically defacto limited to what the gas limit is per block–as IIUC this is what impacts the amount that can be written into storage w/ a string in one "chunk"?

Ah, that makes sense!

ryley-o:
I guess if this saves roughly 2/3, we should be able to have 3x larger chunks.

jakerockland:
> I guess if this saves roughly 2/3, we should be able to have 3x larger chunks. I can find the limits via test though.

Yeah good point! Wasn't sure if there were other limits on contract upload size other than gas limits.

ryley-o:
>Yeah good point! Wasn't sure if there were other limits on contract upload size other than gas limits.

This is true - 24kb is thee limit. Looks like our current chunks are about 10kb max, so probably closer to ~2.4x larger chunks.

ryley-o:
closing in favor of #299

PR #273, opened 2022-08-26
Contract base URI as fallback

ryley-o:
Add a contract base URI that projects without an artist-populated project URI fall back to.

based on conversation with @Jbern16 

This:
- adds a new, admin-configurabele `contractBaseURI` field, automatically populated to `https://token.artblocks.io/` during deployment
- updates `tokenURI(...)` logic to fall back to use `contractBaseURI` if a `projectBaseURI` is not populated
- leaves `projectBaseURI` as an artist-configurable field, keeping artists in control of their project's preferred token endpoint

## Design Choices
- Upgraded to latest supported hardhat and solidity v0.8.16 so we can use the new solidity `string.concat(...)` method (instead of manually concatenating strings in v0.8.9)
  - If we don't like updating things, we could go back to using abi.encodePacked instead of string.concat
- `contractBaseURI` is admin-configurable, and updatable post-deployment if we want to change our token API endpoint in the future
  
## Gas Impacts
This has a negligible impact to mint gas costs.

## Audit
Would be nice to include in-scope if still possible.

---
  - https://app.asana.com/0/0/1202871381462797

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -1464,11 +1464,9 @@ contract GenArt721CoreV3 is
             _projectBaseURILen := mload(_projectBaseURI)
         }
         if (_projectBaseURILen > 0) {
-            return
-                string(abi.encodePacked(_projectBaseURI, _tokenId.toString()));
+            return string.concat(_projectBaseURI, _tokenId.toString());
         } else {
-            return
-                string(abi.encodePacked(contractBaseURI, _tokenId.toString()));
+            return string.concat(contractBaseURI, _tokenId.toString());
         }
     }
 
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -23,7 +23,6 @@ contract GenArt721CoreV3 is
     IGenArt721CoreContractV3
 {
     using Bytes32Strings for bytes32;
-    using Strings for address;
     using Strings for uint256;
     uint256 constant ONE_MILLION = 1_000_000;
     uint24 constant ONE_MILLION_UINT24 = 1_000_000;
@@ -238,13 +237,7 @@ contract GenArt721CoreV3 is
         // set AdminACL management contract as owner
         _transferOwnership(_adminACLContract);
         // initialize contract base URI
-        _updateContractBaseURI(
-            string.concat(
-                "https://token.artblocks.io/",
-                address(this).toHexString(),
-                "/"
-            )
-        );
+        _updateContractBaseURI("https://token.artblocks.io/");
         // initialize next project ID
         _nextProjectId = uint248(_startingProjectId);
         emit PlatformUpdated(FIELD_NEXT_PROJECT_ID);
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -1458,16 +1458,11 @@ contract GenArt721CoreV3 is
     {
         string memory _projectBaseURI = projects[_tokenId / ONE_MILLION]
             .projectBaseURI;
-        uint256 _projectBaseURILen;
-        assembly {
-            // load string length directly from first word
-            _projectBaseURILen := mload(_projectBaseURI)
-        }
-        if (_projectBaseURILen > 0) {
+        if (bytes(_projectBaseURI).length > 0) {
             return string.concat(_projectBaseURI, _tokenId.toString());
-        } else {
-            return string.concat(contractBaseURI, _tokenId.toString());
         }
+        // fallback to contract base URI
+        return string.concat(contractBaseURI, _tokenId.toString());
     }
 
     /**
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-only
-pragma solidity 0.8.9;
+pragma solidity 0.8.16;
 
 // Created By: Art Blocks Inc.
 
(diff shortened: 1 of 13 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -230,6 +237,14 @@ contract GenArt721CoreV3 is
...
         // set AdminACL management contract as owner
         _transferOwnership(_adminACLContract);
+        // initialize contract base URI
+        _updateContractBaseURI(
+            string.concat(
+                "https://token.artblocks.io/",
+                address(this).toHexString(),
+                "/"
+            )
+        );
```
jakerockland, line 247: Is this codifying an assumption that V3 needs to have a base-URI that includes the contract address rather than one that just supports `https://token.artblocks.io/{TOKEN_ID}`?

I don't love that, and think that's not a good path forward (I think that `https://token.artblocks.io/{TOKEN_ID}` fallback _needs_ to work properly for Art Blocks flagship).

If that isn't the case, I would suggest here that the default assignment on construction should be just `https://token.artblocks.io/` (no concatenating the address).

contracts/GenArt721CoreV3.sol
```diff
@@ -1425,13 +1456,20 @@ contract GenArt721CoreV3 is
...
-                    Strings.toString(_tokenId)
-                )
-            );
+        string memory _projectBaseURI = projects[_tokenId / ONE_MILLION]
+            .projectBaseURI;
+        uint256 _projectBaseURILen;
+        assembly {
+            // load string length directly from first word
+            _projectBaseURILen := mload(_projectBaseURI)
+        }
```
jakerockland, line 1465: Do we need the fancy-ness of doing this in assembly?

ryley-o:
I think this should help reduce unnecessary overhead of artists always populating the same project base URI. I will update our design document with these changes upon merge.

jakerockland (changes requested):
Would like to discuss some of the assumptions/implications around this change before we merge.

contracts/GenArt721CoreV3.sol
ryley-o: The field is admin-updateable, so the value during deployment is not permanent.

Including the address was based on quick convo w/ @Jbern16 yesterday, but definitely can remove the addr here if we don't want to go in that direction for flagship. There is a slight benefit of including the addr in the default token URI during deployment because it would be common between flagship and AB engine contracts, but also totally reasonable to have difference in the contracts here.

jakerockland: Makes sense to me–I think if we just make this `https://token.artblocks.io/`, I'm less concerned about the assumptions of this PR because it forces us to handle this backwards compatibility in our infra.

If we want to have address injected, then I think we need to discuss this a bit more...

> There is a slight benefit of including the addr in the default token URI during deployment because it would be common between flagship and AB engine contracts, but also totally reasonable to have difference in the contracts here.

I'd say this should just be one of the differences between V3 and V4 (new partner/engine standard).

contracts/GenArt721CoreV3.sol
ryley-o: This is typically called as a read-only, so yeah may as well cast as bytes and strip out the assembly. pushed an update!

jakerockland: Fantastic–TYVM 🙏 🙏

contracts/GenArt721CoreV3.sol
ryley-o: SGTM - updated to just be `https://token.artblocks.io/` in constructor, no address.

Cc @Jbern16 @lyaunzbe

jakerockland (approved):
Fantastic–with updated changes LGTM/SGTM to merge. If we want to discuss more the token API implications for flagship–please pull me into that conversation makes sense and happy to share opinions about my concerns with original setup of this PR. I should have decent amount of openness on my calendar for next week for this convo if it is time sensitive 🙏 

TY for putting this PR together and updating so quickly.

ryley-o:
_updated description based off of feedback iterations_

contracts/GenArt721CoreV3.sol
```diff
@@ -884,6 +894,18 @@ contract GenArt721CoreV3 is
         emit ProjectUpdated(_projectId, FIELD_PROJECT_BASE_URI);
```
Jbern16, line 894: Right now Hasura tracks this data as `token_base_url` IIRC. This is used within the token api for some logic to how to construct some pieces of data (external_url). Should we sync contract base uri as default as well? Or will this happen automatically with inheritance here as structured?

Jbern16 (approved):
this LGTM. Just tagging @lyaunzbe for visibility. Also had one question!

contracts/GenArt721CoreV3.sol
ryley-o: Very much believe we should also sync contract base URI as well, and plan to update the V3 Integration design doc with a `contract.baseURI` field in our subgraph, and we will want to sync to our DB as well 😄

ryley-o:
Note an update to contractBaseUri would cause us to potentially update all tokenURIs on a given project in a single event handler. That seems like a bad idea for our subgraph. per some discussion w/ @yoshiwarab , some of our tokenURIs are not updated in our current subgraph deployment (unintentionally, due to a bug/missing update on mint), so tokenURI field is not being actively used on any of our public subgraphs today. Thus, we feel like we can update our subgraph's `Token.tokenURI` field to be a template, and we can handle the fallback logic when populating our calculating values in our DB.

ryley-o:
after some more discussion on this w/ @yoshiwarab - Going to close this PR in favor of a slightly different option. Instead, we will have a default contractBaseURI that gets automatically populated when adding new projects (for the operational benefit). However, going to keep projectBaseURI as the only source of a token's base URI.

This avoids the potential of having to iterate through every single token on a contract, and also avoids having to explicitly set the project base URI. If we were to change what our base URI should be in the future, we would have to update that for every project, which seems acceptable.

Cc @jakerockland @Jbern16

contracts/GenArt721CoreV3.sol
lyaunzbe: this decision sgtm fwiw 👍

PR #276, opened 2022-08-29
Default contract base uri

ryley-o:
Add a default base URI for contract that **NEW** projects automatically are populated with, but artists may update at any time.

This is an update, preferred implementation of #273, which used a fallback-style contract-level baseURI instead of an automatically populated version here. Note that keeping base URIs defined at the project level keeps update iterations at or below 1e6 iterations in our subgraph event handlers, whereas a contract-level fallback baseURI update could cause a subgraph event handler to have to iterate over all tokens on a contract (considered infeasible, so would tip us into only providing tokenURI templates on our subgraph, which is not preferred).

This:
- adds a new, admin-configurabele `defaultBaseURI` field, automatically populated to `https://token.artblocks.io/` during deployment
- automatically populates **new** projects' `projectBaseURI` to the contract's `defaultBaseURI` in the `addProject` function
- does NOT update `tokenURI(...)` logic (i.e. still uses `projectBaseURI`)
- leaves `projectBaseURI` as an artist-configurable field, keeping artists in control of their project's preferred token endpoint

## Design Choices
- Upgraded to latest supported hardhat and solidity v0.8.16 so we can use the new solidity `string.concat(...)` method (instead of manually concatenating strings in v0.8.9)
- `defaultBaseURI` is admin-configurable, and updatable post-deployment if we want to change our atomically-populated token API endpoint in the future
- If we want to change our base URIs in the future, we will have to have all existing projects update their projectBaseURI (via artist call)
  
## Gas Impacts
This has a negligible impact to mint gas costs.

## Audit
Would be nice to include in-scope if still possible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
   lint-and-test:
     docker:
       - image: "cimg/base:stable"
-    resource_class: medium
+    resource_class: large
     steps:
       - checkout
       - node/install:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,8 @@ orbs:
 jobs:
   lint-and-test:
     docker:
-      - image: 'cimg/base:stable'
+      - image: "cimg/base:stable"
+    resource_class: medium
     steps:
       - checkout
       - node/install:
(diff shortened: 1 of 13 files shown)


jakerockland (approved):
Disregard my comment in https://github.com/ArtBlocks/artblocks-contracts/pull/273#issuecomment-1230834026, I misunderstood what you meant by https://github.com/ArtBlocks/artblocks-contracts/pull/273#issuecomment-1230792934 and this change LGTM–I like this approach.

ryley-o:
Cc @yoshiwarab

Jbern16 (approved):

PR #279, opened 2022-09-01
Clear proposals upon acceptance, encode instead of encodePacked

ryley-o:
Update V3 core logic to clear artist proposed addresses and payment splits upon admin acceptance, and improve edge case by using `abi.encode` (instead of `abi.encodePacked`) when calculating/verifying `proposedArtistAddressesAndSplitsHash`.

Two main updates in this PR:
- Clear artist proposed addresses and payment splits upon admin acceptance
  - This is more intuitive from a process perspective, and prevents stale approvals from remaining valid
- Calculates `proposedArtistAddressesAndSplitsHash` via `abi.encode` of the values, instead of `abi.encodePacked`
  - This prevents a low-severity bug where admin moves decimals between addresses and values, potentially resulting in unexpected splits percentages and/or payment addresses
  - This bug would only happen if admin behaved somewhat maliciously, or our frontend did some really crazy stuff.

## Audit Impacts
Recommend Cc'ing audit team on this PR. They will probably identify these details, and will appreciate us letting them know we have some sort of solution already in-place.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -497,7 +497,7 @@ contract GenArt721CoreV3 is
         );
         // effects
         proposedArtistAddressesAndSplitsHash[_projectId] = keccak256(
-            abi.encodePacked(
+            abi.encode(
                 _artistAddress,
                 _additionalPayeePrimarySales,
                 _additionalPayeePrimarySalesPercentage,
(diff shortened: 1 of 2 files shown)


jakerockland (approved):
This LGTM! TY for this update!! 🙏 🙏 

> Recommend Cc'ing audit team on this PR. They will probably identify these details, and will appreciate us letting them know we have some sort of solution already in-place.

Could you flag this in the shared Slack pls ser? 🙏

ryley-o:
>Could you flag this in the shared Slack pls ser? 🙏

Done ✅  😄

PR #280, opened 2022-09-01
V3 Consistent PROJECT naming convention

ryley-o:
Based on one-off DMs w/ @yoshiwarab - use consistent PROJECT naming convention when defining ProjectUpdated field constants.

## Audit Impacts
This is a no-op and can be ignored with respect to audits of the V3 contract


ryley-o:
Cc @jakerockland for visibility

PR #283, opened 2022-09-13
Engine Flex Core Contract: adding coreType and coreVersion public constants

lyaunzbe:
adding coreType and coreVersion public constants to flex contract

```
    string public constant coreVersion = "v2.0.0";
    string public constant coreType = "GenArt721CoreV2_ENGINE_FLEX";
```


ryley-o (approved):
LGTM!

PR #284, opened 2022-09-14
Remove project IPFS hash from V3 core

ryley-o:
per conversation, remove V3's project IPFS hash.

Follow-on updates will be needed for V3 subgraph to change how view function is handled.

## Audit impacts
This only removes data/complexity from contracts, and I don't think needs audit review.


ryley-o:
Cc @lyaunzbe

jakerockland (approved):
🙏 🙏

PR #285, opened 2022-09-14
V3/Expose project completed timestamp

ryley-o:
Expose V3 core project `completedTimestamp` in `projectStateData` view.

Based off of some discussion a while ago with @yoshiwarab. To be clear, this is a nice-to-have, and not a must have. It does allow us to view the on-chain completedTimestamp value without relying on our subgraph's indexing of the event `ProjectUpdated(_projectId, FIELD_PROJECT_COMPLETED)`.

## Downstream Impacts
I don't think this requires any updates to the subgraph or frontend (since we are only adding more data to the view), other than updating the ABI files for the V3 core contract.

## Audit Impacts
I don't think this needs an audit review - the change is very minimal

---
  - https://app.asana.com/0/0/1202896474680854


jakerockland (approved):
LGTM! 🙏

PR #286, opened 2022-09-14
Add AdminACLV0 to be used in production

ryley-o:
Add a new AdminACLV0 that is suitable for use in production.

This adds an AdminACLV0 that doesn't emit unnecessary events (MockAdminACLV0 emits events for testing purposes only).

The new AdminACLV0 only has a single, transferrable SuperAdmin that is allowed to call any function on any contract. No other addresses are ever allowed to call anything.

Two functions are allowed to be called by the AdminACLV0 contract itself:
- `Ownable.renounceOwnership()`: enables ownership of a V3 core to be renounced
- `Ownable.transferOwnership()`: enables this AdminACLV0 to transfer AdminACL rights to a new AdminACL contract in the future

The two functions enabling the two calls above (`transferOwnershipOn` & `renounceOwnershipOn`) were added to the `IAdminACLV0` interface. Additionally, a check that requires the new AdminACL in `transferOwnershipOn` broadcasts support of IAdminACLV0 via ERC165 interface detection is included to prevent 🧱 .

## Test updates
Tests are updated to use this new AdminACLV0 contract instead of MockAdminACLV0 with events, except for the events tests that explicitly use events to validate desired behaviors. Additionally, new tests are added to ensure the check for ERC165 conformance of IAdminACLV0 works as intended.

## Audit impacts
This is a good Cc for the audit team, but changes from the audited `MockAdminACLV0.sol` contract are relatively straightforward.


contracts/AdminACLV0.sol
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * @dev Allows superAdmin to call renounceOwnership on other contract from
+     * this contract.
+     */
+    function renounceOwnershipOn(address _contract) external {
+        require(msg.sender == superAdmin, "Only superAdmin");
+        Ownable(_contract).renounceOwnership();
+    }
```
jakerockland, line 49: I'm not sure I'm understanding if/why we _need_ to expose these here at all.

ryley-o: The V3 core uses OpenZeppelin's Ownable standard, which exposes these two functions gated to `onlyOwner`. Since a V3 core's owner is the AdminACL contract itself, this implementation just allows the superAdmin to instruct the AdminACL contract to call the functions on the V3 core.

An alternate implementation would be to expose functions on our V3 core that transfer/revoke ownership, and then manage the calling of those functions in our AdminACL schema. That might be a bit cleaner than this, WDYT?

jakerockland: > An alternate implementation would be to expose functions on our V3 core that transfer/revoke ownership, and then manage the calling of those functions in our AdminACL schema. That might be a bit cleaner than this, WDYT?

Yeah this seems more right to me, that we'd expose this on the Core and then ACL-check the ability to do so on the AdminACL. They're functionally quite equivalent though..

ryley-o: per sync discussion, prefer to keep these transfer-of-adminship functions on the AdminACL contract, but also would like to have a check that the new AdminACL contract conforms to IAdminV0, where IAdminV0 includes the `transferOwnershipOn` and `renounceOwnershipOn` functions.

@jakerockland pushed updates, tagging you for a re-review with the updates 🙌

jakerockland (approved):
LGTM! 🙏 🚢 

TYVM for the iterations here.

PR #289, opened 2022-09-15
Enable indexing of AdminACL superAdmin

ryley-o:
Update AdminACLV0 to enable subgraph indexing of its `superAdmin` address.

Adds:
- event emitted when superAdmin is changed
- includes superAdmin in `IAdminACLV0`
  - Allows subgraph to use only `IAdminACLV0` ABI, minimizing subgraph code bloat if we roll versions of this in the future.
- tests for new event

## Design Choices
- `New SuperAdminTransferred` event includes an `address[] genArt721CoreAddressesToUpdate` for indexing purposes only. This is because when the subgraph views the new event, it needs to know which contracts it should attempt to the new superAdmin. It seemed most straightforward to include the addresses in the call to transfer ownership and the event.

## Audit Impacts
This change is straightforward and should not impact audit results.

### DIFF
--- a/contracts/AdminACLV0.sol
+++ b/contracts/AdminACLV0.sol
@@ -22,12 +22,23 @@ contract AdminACLV0 is IAdminACLV0, ERC165 {
 
     /**
      * @notice Allows superAdmin change the superAdmin address.
+     * @param _newSuperAdmin The new superAdmin address.
+     * @param _genArt721CoreAddressesToUpdate Array of genArt721Core
+     * addresses to update to the new superAdmin, for indexing purposes only.
+     * @dev this function is gated to only superAdmin address.
      */
-    function changeSuperAdmin(address _newSuperAdmin) external {
+    function changeSuperAdmin(
+        address _newSuperAdmin,
+        address[] calldata _genArt721CoreAddressesToUpdate
+    ) external {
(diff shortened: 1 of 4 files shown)


jakerockland (approved):
LGTM but I would think we probably also want to emit event then on transfers/revokes?

ryley-o:
>LGTM but I would think we probably also want to emit event then on transfers/revokes?

We definitely could, however we would expect that anything conforming to `Ownable` would emit it's own events, according to that standard. In this case, the V3 core emits the standard `Ownable.OwnershipTransferred` event on both of those calls, so emitting an event from the AdminACLV0 is a bit redundant. I can update if you prefer though.

jakerockland:
> > LGTM but I would think we probably also want to emit event then on transfers/revokes?
> 
> @jakerockland We definitely could, however we would expect that anything conforming to `Ownable` would emit it's own events, according to that standard. In this case, the V3 core emits the standard `Ownable.OwnershipTransferred` event on both of those calls, so emitting an event from the AdminACLV0 is a bit redundant. I can update if you prefer though.

Ah OK, my apologies I didn't realize that Ownable emitting events was a strong guarantee. Leaving it out SGTM then in that case.

PR #294, opened 2022-09-20
QSP-11-Fixed - Lock MockAdminACLV0 to explicit Solidiy pragma

jakerockland:
Update `MockAdminACLV0` to use locked pragma Solidity version.


ryley-o (approved):
@jakerockland - edit - see follow on #297

PR #297, opened 2022-09-20
Lock AdminACLV0 to solidity version

ryley-o:
Lock AdminACLV0 to specific solidity version.

Follow-on from #294


jakerockland (approved):
🙏 🙏 🙏

PR #299, opened 2022-09-20
Contract bytecode for script storage

ryley-o:
Update V3 core to use contract bytecode for persistent script storage to achieve significant gas reduction during artist script upload.

This implementation is tailored to cleanly save Art Blocks script string data. It does not involve any compression of the raw string data.

Perhaps in the future, this method of storing strings can be standardized, and blockchain explorers could include displaying the stored utf-8 strings.

To view compartmentalized diff for purge-logic, please see #304, which was folded into this PR.

## Script Chunk Size Limits
As shown in the table below, this implementation limits script chunks to the contract size limit of <24 KB, due to the current mainnet contract size limit. In today's baseline, contract chunks of up to ~41 KB are able to be uploaded, only being limited by the current mainnet block gas limit. If this is undesirable UX, a follow-on PR could automatically divide added scripts into 24 KB chunks. With that type of implementation, roughly `24 KB * (30,000,000 / 5,438,309) =  132 KB` of script string data could be uploaded in a single transaction, resulting in 6 "bytecode storage contracts" being deployed.

## Gas impacts
Script upload costs are reduced by roughly 67%, as shown in the table below.

_(All gas costs assume project 3, script index 0)_
| Gas Test | Previous gas to upload script | New gas used to upload script | % change |
| --- | --- | --- | --- |
| GenArt721CoreV3: Upload Squiggle Script | 2_801_726 | 930_685 | -66.8% |
| GenArt721CoreV3: Upload Skulptuur-like Script | 4_572_755 | 1_493_682 | -67.7% |
| GenArt721CoreV3: Upload 23.95 KB script | 17_423_468 | 5_438_309 | -68.8% |
| GenArt721CoreV3: Upload 24.00 KB script | ~17_424_000 | Not Possible (contract size limit) | - |
| GenArt721CoreV3: Upload 40.99 KB script | 29_685_403 | Not Possible (contract size limit) | - |
| GenArt721CoreV3: Upload script > ~41 KB | Not Possible (block gas limit) | Not Possible (contract size limit) | - |

## Other
Ethers has a difficult time estimating the `gasLimit` for script upload transactions. Our frontend may have to develop our own logic for recommended gas limits of script upload transactions.

## Credit

This follows a similar pattern as other libraries that use contract bytecode for storage of raw bytes:
- https://github.com/0xsequence/sstore2/blob/master/contracts/SSTORE2.sol
- https://github.com/transmissions11/solmate/blob/main/src/utils/SSTORE2.sol

---
  - https://app.asana.com/0/0/1202817061735237
  - https://app.asana.com/0/0/1203066943086879


contracts/GenArt721CoreV3.sol
```diff
@@ -871,7 +877,8 @@ contract GenArt721CoreV3 is
     {
         Project storage project = projects[_projectId];
         require(_scriptId < project.scriptCount, "scriptId out of range");
-        project.scripts[_scriptId] = _script;
+        // store script in contract bytecode
+        project.scriptBytecodeAddresses[_scriptId] = _script.writeToBytecode();
```
jakerockland, line 881: Making direct note of this for posterity as follow up from our call, this is one call-site where we'd want to `selfdestruct` the old script/script-chunk that is being replaced.

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,109 @@
...
+        // address must be non-zero if contract was deployed successfully
+        require(address_ != address(0), "ContractAsStorage: Write Error");
+    }
+
+    /**
+     * @notice Read a string from contract bytecode
+     * @param _address address of deployed contract with bytecode containing concat(0x00, data)
+     * @return data string read from contract bytecode
+     */
+    function readFromBytecode(address _address)
```
jakerockland, line 62: I may be overlooking something, but could we add a unit test that validates `readFromBytecode` (that bytes in == bytes out)?

jakerockland (changes requested):
Some initial questions/requested minor adjustments, generally though this LG and is really exciting but want to hash out some of the remaining pieces here before marking as approved. 🙏 🚀

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,109 @@
...
+            // 0x38    |  0x38               | CODESIZE     | codeSize codeOffset 0 codeOffset                               //
+            // 0x03    |  0x03               | SUB          | (codeSize - codeOffset) 0 codeOffset                           //
+            // 0x80    |  0x80               | DUP          | (codeSize - codeOffset) (codeSize - codeOffset) 0 codeOffset   //
+            // 0x92    |  0x92               | SWAP3        | codeOffset (codeSize - codeOffset) 0 (codeSize - codeOffset)   //
+            // 0x59    |  0x59               | MSIZE        | 0 codeOffset (codeSize - codeOffset) 0 (codeSize - codeOffset) //
+            // 0x39    |  0x39               | CODECOPY     | 0 (codeSize - codeOffset)                                      //
+            // 0xf3    |  0xf3               | RETURN       |                                                                //
+            //---------------------------------------------------------------------------------------------------------------//
+            hex"60_0B_59_81_38_03_80_92_59_39_F3", // returns all code in the contract except for the first 11 (0B in hex) bytes
```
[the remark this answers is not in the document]
ryley-o: here we are building the creation code, which is supposed to return the bytecode of the contract to be created. We want the deployed contract to contain the 0x00 at the start to prevent it from being called.

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,109 @@
...
+    uint256 internal constant DATA_OFFSET = 1;
+
+    /**
+     * @notice Write a string to contract bytecode
+     * @param _data string to be written to contract
+     * @return address_ address of deployed contract with bytecode containing concat(0x00, data)
+     */
+    function writeToBytecode(string memory _data)
+        internal
+        returns (address address_)
```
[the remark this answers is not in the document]
ryley-o: I think when working in assembly it is nice to define address_ up-front so we don't have to do it in assembly (and deal with moving the free-memory pointer).

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,109 @@
...
+    function readFromBytecode(address _address)
+        internal
+        view
+        returns (string memory data)
+    {
+        // get the size of the data
+        uint256 codeSize = _codeSizeAt(_address);
+        // handle case where address does not contain code
+        if (codeSize == 0) {
+            return "";
```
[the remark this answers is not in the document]
ryley-o: yeah, we should probably revert here, as this case is never expected to be possible.

contracts/libs/0.8.x/BytecodeStorage.sol
ryley-o: I don't know if hardhat supports testing this library directly, but we essentially test that in `test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts`, right?

contracts/libs/0.8.x/BytecodeStorage.sol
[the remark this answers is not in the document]
jakerockland: I see I see I see, thank you for explaining here and over call, my misunderstanding here.

contracts/libs/0.8.x/BytecodeStorage.sol
jakerockland: Ugh sorry, I had missed this I think because I didn't have that folder open or something on the side bar. Just took a look and LGTM and def does satisfy my question here. Adding one minor comment to the tests below.

test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
```diff
@@ -18,6 +18,11 @@ import {
   advanceEVMByTime,
 } from "../../util/common";
 import { FOUR_WEEKS } from "../../util/constants";
+import {
+  SQUIGGLE_SCRIPT,
+  SKULPTUUR_SCRIPT_APPROX,
+  CONTRACT_SIZE_LIMIT_SCRIPT,
```
jakerockland, line 24: Nit: I feel like we may want an assertion below that asserts that each of these imports is not the empty string, just to catch the case in the future where someone accidentally deletes those scripts or something but the above would still pass IIUC.

contracts/libs/0.8.x/BytecodeStorage.sol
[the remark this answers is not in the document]
jakerockland: SGTM, would say we can probably update to just handling this case and the `codeSize < DATA_OFFSET` case as the same type of error.

jakerockland (approved):

jakerockland:
One other thing, realized we should be updating the V3 changelog with this PR as well.

contracts/GenArt721CoreV3.sol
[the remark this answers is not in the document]
jakerockland: This is addressed now with folding in of #306 into this PR.

contracts/GenArt721CoreV3.sol
```diff
@@ -885,7 +892,7 @@ contract GenArt721CoreV3 is
     {
         Project storage project = projects[_projectId];
         require(project.scriptCount > 0, "there are no scripts to remove");
-        delete project.scripts[project.scriptCount - 1];
+        delete project.scriptBytecodeAddresses[project.scriptCount - 1];
```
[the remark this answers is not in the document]
jakerockland: This is addressed now with folding in of #306 into this PR.

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * @notice Read a string from contract bytecode
+     * @param _address address of deployed contract with bytecode containing concat(0x00, data)
+     * @return data string read from contract bytecode
+     */
+    function readFromBytecode(address _address)
+        internal
+        view
+        returns (string memory data)
```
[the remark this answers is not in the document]
jakerockland: Addressed w/ https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/4598127e62df21542bf7c9be1daef0526889879d

contracts/libs/0.8.x/BytecodeStorage.sol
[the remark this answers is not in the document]
jakerockland: Addressed w/ https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/4598127e62df21542bf7c9be1daef0526889879d

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,109 @@
...
+            extcodecopy(_address, add(data, 0x20), DATA_OFFSET, size)
+        }
+    }
+
+    /**
+        @notice Returns the size of the code at address `_address`
+        @param _address address that may or may not contain code
+        @return size size of code at `_address`
+    */
+    function _codeSizeAt(address _address) private view returns (uint256 size) {
```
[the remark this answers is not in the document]
jakerockland: Addressed w/ https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/4598127e62df21542bf7c9be1daef0526889879d

test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
jakerockland: This is addressed w/ https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/e71fae326601bcb009f9c03814e303ebfea765ab

contracts/GenArt721CoreV3.sol
ryley-o: This is addressed now with folding in of https://github.com/ArtBlocks/artblocks-contracts/pull/306 into this PR.

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,183 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title Art Blocks Script Storage Library
+ * @author Art Blocks Inc.
+ * @author Modified from 0xSequence (https://github.com/0xsequence/sstore2/blob/master/contracts/SSTORE2.sol)
+ * @author Modified from Solmate (https://github.com/transmissions11/solmate/blob/main/src/utils/SSTORE2.sol)
```
jakerockland, line 10: TODO: I'm going to add a quick explanation here under these that clarifies what modifications this library makes relative to the two SSTORE2 libraries above?

contracts/libs/0.8.x/BytecodeStorage.sol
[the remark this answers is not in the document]
jakerockland: This is addressed now here: https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/a260e9f81868683e8caebe468016322e2cbbb787

contracts/libs/0.8.x/BytecodeStorage.sol
jakerockland: This has also been addressed, and documentation cleaned up with a couple more passes.

test/core/V3/GenArt721CoreV3_Views.test.ts
```diff
@@ -1421,11 +1421,11 @@ describe("GenArt721CoreV3 Views", async function () {
...
-    it("returns empty string by default", async function () {
-      const emptyProjectScript = await this.genArt721Core
-        .connect(this.accounts.user)
-        .projectScriptByIndex(this.projectZero, 0);
-      expect(emptyProjectScript).to.be.equal("");
+    it("reverts on invalid reads", async function () {
+      await expectRevert(
+        this.genArt721Core.projectScriptByIndex(this.projectZero, 0),
+        "ContractAsStorage: Read Error"
+      );
```
jakerockland, line 1428: @ryley-o want to explicitly double-click on this here and make sure you feel this behavior is acceptable, WDYT?

ryley-o: hmm, we should probably actually check that they aren't checking a script index that is out of bounds in our V3 core function, and give a more informative error here. good call!

jakerockland: Yeah that makes sense to me–do you think that belongs in this PR or a different follow up one? 

Alternatively, we could make behavior of `BytecodeStorage` to return `""` always in the empty-case rather than revert. 

I'm not sure I have a strong opinion here really..

test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
```diff
@@ -714,6 +727,173 @@ describe("GenArt721CoreV3 Project Configure", async function () {
...
+        "ContractAsStorage: Read Error"
+      );
+
+      const removedScriptByteCode = await ethers.provider.getCode(
+        scriptAddress
+      );
+      expect(removedScriptByteCode).to.equal("0x");
+    });
+
+    it("uploads chromie squiggle script and attempts to purge from non-allowed address", async function () {
```
ryley-o, line 798: Thinking on this - we should add a similar test to this that specifically contains a bunch of valid utf-8 characters >1 byte

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                          INTERNAL HELPER LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+        @notice Returns the size of the code at address `_address`
+        @param _address address that may or may not contain code
+        @return size size of code at `_address`
```
ryley-o, line 211:
```suggestion
        @notice Returns the size of the bytecode at address `_address`
        @param _address address that may or may not contain bytecode
        @return size size of the bytecode code at `_address`
```

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,218 @@
...
+     * @notice Read a string from contract bytecode
+     * @param _address address of deployed contract with bytecode containing concat(gated-cleanup-logic, data)
+     * @return data string read from contract bytecode
+     */
+    function readFromBytecode(address _address)
+        internal
+        view
+        returns (string memory data)
+    {
+        // get the size of the data
```
ryley-o, line 156:
```suggestion
        // get the size of the bytecode
```

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,218 @@
...
+        uint256 codeSize = _codeSizeAt(_address);
```
ryley-o, line 157: WDYT about calling this `bytecodeSize`?

ryley-o (commented):
I cannot approve (since I own), but these are the only remaining thoughts I have on this for now 😃

contracts/libs/0.8.x/BytecodeStorage.sol
jakerockland: Makes sense to me–updating!

test/core/V3/GenArt721CoreV3_Views.test.ts
ryley-o: I suppose we might want to be consistent with our prior behavior, so having the contract return "" when an index is out of bounds makes sense. I thinking having the library revert is okay, since empty string should be when the bytecode of a contract represents an empty string, not when the bytecode is non-existent.

jakerockland: Yeah that makes sense and SGTM–will update our implementation in the Core contract as part of this PR.

jakerockland:
Added additional commit to address this best practice in our purge calls: https://docs.guardrails.io/docs/vulnerabilities/solidity/use_of_low_level_call

test/core/V3/GenArt721CoreV3_Views.test.ts
jakerockland: This should be addressed now here: https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits

PTAL @ryley-o and let me know if this looks off at all.

ryley-o: LGTM (pretty sure you mean https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/c1ce3b85bda390647e0a003654d37fa85c1163ed) 🎉

test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
jakerockland: Updated to add a specific test case for this that uses a newly introduced `RANDOM_UTF_EIGHT_SCRIPT` mocked up script with a bunch of utf-8 characters.

jakerockland:
Note to any OSS viewers who are looking at this changeset based on the latest commit, from some additional discussion @ryley-o and I have determined it is a much safer/better pattern to have the `selfdestruct` logic not _just be gated_ by a raw `.call` but to have the bytecode in `BytecodeStorage` gate this `selfdestruct`ion on an explicit message passed as call data into the `.call` to tighten the security assumptions of this approach. We're planning to send out a follow on PR that makes this adjustment which will fold into this one.

jakerockland:
> Note to any OSS viewers who are looking at this changeset based on the latest commit, from some additional discussion @ryley-o and I have determined it is a much safer/better pattern to have the `selfdestruct` logic not _just be gated_ by a raw `.call` but to have the bytecode in `BytecodeStorage` gate this `selfdestruct`ion on an explicit message passed as call data into the `.call` to tighten the security assumptions of this approach.
> 
> We're planning to send out a follow on PR that makes this adjustment which we will fold into this PR before this PR is considered `main`-merge ready.

This has been accomplished with https://github.com/ArtBlocks/artblocks-contracts/pull/308

test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
```diff
@@ -714,6 +729,221 @@ describe("GenArt721CoreV3 Project Configure", async function () {
...
+        .connect(this.accounts.artist)
+        .addProjectScript(this.projectZero, SKULPTUUR_SCRIPT_APPROX);
+      const script = await this.genArt721Core.projectScriptByIndex(
+        this.projectZero,
+        0
+      );
+      expect(script).to.equal(SKULPTUUR_SCRIPT_APPROX);
+    });
+
+    it("uploads and recalls 23.95 KB script", async function () {
```
shantanubala, line 849: Do we expect specific behavior when the script exceeds the size limit?

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,257 @@
...
+    function writeToBytecode(string memory _data)
+        internal
+        returns (address address_)
+    {
+        // prefix bytecode with
+        bytes memory creationCode = abi.encodePacked(
+            //---------------------------------------------------------------------------------------------------------------//
+            // Opcode  | Opcode + Arguments  | Description  | Stack View                                                     //
+            //---------------------------------------------------------------------------------------------------------------//
+            // (0) creation code returns all code in the contract except for the first 11 (0B in hex) bytes
```
shantanubala, line 53: I was doing some digging, and I don't fully understand why this is offset by 11 bytes specifically. Why 11?

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,257 @@
...
+ *      - stores the "writer" address (library user) in the deployed contract bytes, which is useful for both:
+ *         a) providing necessary information for safe deletion; and
+ *         b) allowing this to be introspected on-chain
+ *      Also, given that much of this library is written in assembly, this library makes use of a slightly
+ *      different convention (when compared to the rest of the Art Blocks smart contract repo) around
+ *      pre-defining return values in order to simplify need to directly memory manage these return values.
+ */
+library BytecodeStorage {
+    //---------------------------------------------------------------------------------------------------------------//
+    // Offset Amount | Offset Aggregate | Description                                                                //
```
shantanubala, line 27: Nit: I personally prefer documenting the starting offset and the size of the value, since those are the two most relevant parts for pointer math, but Solidity doesn't really do pointers so maybe that's not too important.

test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
jakerockland: Great point, adding a test to cover this case.

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -0,0 +1,257 @@
...
+    //---------------------------------------------------------------------------------------------------------------//
+    // 53            | 53               | allow contract to be `selfdestruct`-able via gated-cleanup-logic           //
+    // 20            | 73               | reserve 20 bytes for storing deploying-contract's address                  //
+    //---------------------------------------------------------------------------------------------------------------//
+    // define the offset for where the "logic bytes" end, and the "data bytes" begin
+    uint256 internal constant DATA_OFFSET = 73;
```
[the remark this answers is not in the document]
jakerockland: Unfortunately, I don't think that this is possible without writing some sort of custom semantic analysis tool or assembly-generator based on config files or something like that which is out of scope and a bit scary heh – but yes this would definitely be very aggressively caught by current test coverage if we messed this offset up.

@ryley-o any thoughts on additional readily-usable ways to enforce this?

contracts/libs/0.8.x/BytecodeStorage.sol
jakerockland: These 11 bytes are excluded as we don't want to include the contract creation code itself in the deployed contract data, which is the length of this contract creation code (11 bytes)–will add comment to clarify this inline.

contracts/libs/0.8.x/BytecodeStorage.sol
jakerockland: I think I may be misunderstanding the feedback here, but in this case these offsets are refering to the deployed contract code that sits in storage vs. memory pointers/offsets, so the feedback may not exactly apply here? 

That said, perhaps we could do something like this, which I believe is in alignment with your nit here, if this seems clearer:

```
Starting Index | Size | Ending Index | Description
0                      | N/A  | 0                     | N/A, starting index is at 0.
0                      | 53    | 53                  | the bytes of the gated-cleanup-logic allowing for `selfdestruct`ion
53                    | 20    | 73                  | the 20 bytes used for storing the deploying-contract's address
```

WDYT @ryley-o @shantanubala ?

jakerockland:
@ryley-o @shantanubala FYI I've added a few commits, based on @shantanubala's feedback and also additional test coverage I realized was worth adding after introduction of #304.

PTAL at https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/afa18bbfa8a0f3ba4516b948b9e0d3d481b2f6c1, https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/77d4843996ab7d8c806b8c8c840b1d75c1f259c8, and https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/66eebd81af1841f9f5971082553953d35d8d6835 if you want to get into each diff'd commit separately.

contracts/libs/0.8.x/BytecodeStorage.sol
shantanubala: That sgtm, but also not a huge deal, I was able to understand what was happening with the current documentation 👍

shantanubala (approved):
This looks great to me! Amazing job on documenting and testing 👏

test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
```diff
@@ -714,6 +734,246 @@ describe("GenArt721CoreV3 Project Configure", async function () {
...
+          gasLimit: 30000000, // hard-code gas limit because ethers sometimes estimates too high
+        });
+      const script = await this.genArt721Core.projectScriptByIndex(
+        this.projectZero,
+        0
+      );
+      expect(script).to.equal(CONTRACT_SIZE_LIMIT_SCRIPT);
+    });
+
+    it("fails to upload 26 KB script", async function () {
```
lindsgil, line 880: If a script is > 24KB should it be chunked and each chunk uploaded separately?

jakerockland:
> This looks great to me! Amazing job on documenting and testing 👏

Thank you very much for TAL! 🙏

contracts/libs/0.8.x/BytecodeStorage.sol
ryley-o: @jakerockland i like your suggested code in your comment above. Seems more clear.

ryley-o:
>PTAL at https://github.com/ArtBlocks/artblocks-contracts/commit/afa18bbfa8a0f3ba4516b948b9e0d3d481b2f6c1, https://github.com/ArtBlocks/artblocks-contracts/commit/77d4843996ab7d8c806b8c8c840b1d75c1f259c8, and https://github.com/ArtBlocks/artblocks-contracts/commit/66eebd81af1841f9f5971082553953d35d8d6835 if you want to get into each diff'd commit separately.

@jakerockland - I've updated the test in https://github.com/ArtBlocks/artblocks-contracts/commit/afa18bbfa8a0f3ba4516b948b9e0d3d481b2f6c1 to avoid the ethers-induced gas-estimation error, and instead pick up our library's write error, in this commit: https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/8d4c2b06e99aafb645f5137785e5ef71fbc6db4a

contracts/libs/0.8.x/BytecodeStorage.sol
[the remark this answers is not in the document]
ryley-o: I've updated comments in this commit: https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/fd2bbf836e8834b5526829dbd2afecccf0863b3d

Cc @jakerockland @shantanubala

contracts/libs/0.8.x/BytecodeStorage.sol
ryley-o: I've updated the table to match @jakerockland 's suggested format above in this commit: https://github.com/ArtBlocks/artblocks-contracts/pull/299/commits/1f4e0a2d5a82e099e6b39685544e6d8fc9528059

jakerockland: TYVM @ryley-o 🙏

jakerockland:
> > PTAL at [afa18bb](https://github.com/ArtBlocks/artblocks-contracts/commit/afa18bbfa8a0f3ba4516b948b9e0d3d481b2f6c1), [77d4843](https://github.com/ArtBlocks/artblocks-contracts/commit/77d4843996ab7d8c806b8c8c840b1d75c1f259c8), and [66eebd8](https://github.com/ArtBlocks/artblocks-contracts/commit/66eebd81af1841f9f5971082553953d35d8d6835) if you want to get into each diff'd commit separately.
> 
> @jakerockland - I've updated the test in [afa18bb](https://github.com/ArtBlocks/artblocks-contracts/commit/afa18bbfa8a0f3ba4516b948b9e0d3d481b2f6c1) to avoid the ethers-induced gas-estimation error, and instead pick up our library's write error, in this commit: [8d4c2b0](https://github.com/ArtBlocks/artblocks-contracts/commit/8d4c2b06e99aafb645f5137785e5ef71fbc6db4a)

This SGTM ser, TYVM! :pra

shantanubala (approved):
Took another look and the changes lgtm 👍

ryley-o:
Update: We have merged #307 into this PR. This adds the ability to directly introspect via the BytecodeStorage library the writer of a given deployed contracts-as-storage contract. This functionality is not actually used by the Art Blocks GenArt721CoreV3 contract, but seemed appropriate to include in the library. In addition to the new introspection capability, more tests were added to improve testing of the BytecodeStorage library.

ryley-o:
just to note - folded in #320 to this PR. It was a functional no-op that improved the readability of the bytecode logic.

PR #303, opened 2022-09-21
Hotfix: Sending correct dependencyType uint with the UpdatedGateway event

lyaunzbe:
The dependencyType being sent by the GatewayUpdated event was mistakenly swapped. Also added tests specifically around the gateway updates (should've been there anyway).


jakerockland (approved):

jakerockland:
> Also added tests specifically around the gateway updates (should've been there anyway).

It looks like these tests may not have been checked in to initial commit?

jakerockland:
> > Also added tests specifically around the gateway updates (should've been there anyway).
> 
> It looks like these tests may not have been checked in to initial commit?

Ignore me, I see them.

ryley-o:
🙏 🙏

PR #304, opened 2022-09-21
Allow for cleanup of unused contract bytecode for script storage

jakerockland:
Co-authored w/ @ryley-o, this PR builds off of #299 (incorporating #306) in order to support the ability to purge from blockchain state bytecode storage contracts deployed by the `BytecodeStorage.sol` class introduced in #299 when they are no longer used.

This is done by using triggering the deployed bytecode contract to itself call [`SELFDESTRUCT`](https://www.evm.codes/#ff) op-code, which will self-delete the bytecode storage contract and remove it entirely from blockchain state. This is beneficial from the perspective of being responsible custodians of EVM state bloat.

Note: this functionality of executing [`SELFDESTRUCT`](https://www.evm.codes/#ff) is being added in such a way that it is gated to only be call-able by the contract that deployed the bytecode storage contract, any other caller who attempts to trigger this logic by direct interaction with the bytecode storage contract will have their call reverted with the [`INVALID`](https://www.evm.codes/#fe) op-code.A

Unit tests have been added that cover the functionality here to cover: a) the delete-script flow, b) the update-script flow, and c) the attempt of directly triggering [`SELFDESTRUCT`](https://www.evm.codes/#ff) via a non-deployer address (which is not possible).

### DIFF
--- a/test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
+++ b/test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
@@ -754,6 +754,40 @@ describe("GenArt721CoreV3 Project Configure", async function () {
       expect(script).to.equal(SQUIGGLE_SCRIPT);
     });
 
+    it("uploads chromie squiggle script, and then purges it", async function () {
+      await this.genArt721Core
+        .connect(this.accounts.artist)
+        .addProjectScript(this.projectZero, SQUIGGLE_SCRIPT);
+      const script = await this.genArt721Core.projectScriptByIndex(
+        this.projectZero,
+        0
+      );
+      expect(script).to.equal(SQUIGGLE_SCRIPT);
+
+      const scriptAddress =
+        await this.genArt721Core.projectScriptBytecodeAddressByIndex(
+          this.projectZero,
+          0
+        );
+
+      const scriptByteCode = await ethers.provider.getCode(scriptAddress);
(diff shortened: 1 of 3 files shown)


ryley-o:
tacked onto this PR  in #306 🥳

test/core/V3/GenArt721CoreV3_ProjectConfigure.test.ts
```diff
@@ -754,6 +754,40 @@ describe("GenArt721CoreV3 Project Configure", async function () {
       expect(script).to.equal(SQUIGGLE_SCRIPT);
     });
 
+    it("uploads chromie squiggle script, and then purges it", async function () {
```
ryley-o, line 757: Reminder: TODO - add additional tests that ensure external accounts cannot trigger selfdestruct

jakerockland: Done in https://github.com/ArtBlocks/artblocks-contracts/pull/304/commits/0d6d941a92507311e39d839129c2ca9c4d4cd61f

Still adding test coverage for one other critical path here and then will take this PR out of draft mode.

ryley-o (approved):
LGTM! Agree that folding this into #299 makes sense 👍

jakerockland:
> LGTM! Agree that folding this into #299 makes sense 👍

SGTM, ahhhhhh I'm folddinngggg

PR #306, opened 2022-09-22
working version of bytecode storage w/selfdestruct

ryley-o:


jakerockland:
🔥 🙌 🥳 💯 🙏 

SO EXCITING. I'm going to merge this into #304 and then can add additional tests (or you can add them if you've already started) but you should be able to push commits to [cleaup-bytecode-storage](https://github.com/ArtBlocks/artblocks-contracts/tree/cleaup-bytecode-storage) @ryley-o so let's just share that PR if that SG?

PR #307, opened 2022-09-23
Introduce `getWriterAddressForBytecode` to `BytecodeStorage` library

jakerockland:
This PR builds on top of #299 and will likely be merged directly into that PR rather than re-based off of main, but is currently split off to make it easier to de-couple from a review perspective.

This adds the ability to directly introspect via the `BytecodeStorage` library the writer of a given deployed contracts-as-storage contract.

Note: this information is introspect-able via the on-chain data written by `BytecodeStorage.sol` regardless of whether or not the `getWriterAddressForBytecode` method of `BytecodeStorage` is used, as this information is written to chain-state in a directly publicly visible way, this convenience reader simply provides a straightforward interface for doing so.

The introspect-ability of this author/writer data is interesting because it enables some interesting use-cases for systems-of-applications (dApps that use multiple smart contracts). For example, in the case of the Art Blocks ecosystem, this introspect-ability allows us to potentially build in the future a tool that allows a collector to input into a given tool the contract address for a deployed script's data, and have that tool use this introspection to reverse-lookup the core-contract that the said script is associated with (e.g. is it an Art Blocks flagship core-contract, an Art Blocks Engine core-contract, an auxiliary registry Art Blocks contract that writes some other provenance related data to chain-state via `BytecodeStorage`, etc.).

As part of this PR, a mock has been added that wraps the `BytecodeStorage.sol`  implementation may be more deeply tested than would be possible by just the usage of this class within `GenArt721CoreV3` alone. Added alongside this new mock is a set of unit tests which prod at the full set of functionality exposed by `BytecodeStorage.sol`.


contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -31,6 +31,8 @@ library BytecodeStorage {
     //---------------------------------------------------------------------------------------------------------------//
     // define the offset for where the "logic bytes" end, and the "data bytes" begin
     uint256 internal constant DATA_OFFSET = 47;
+    // define the offset for where the purely-logic aspects of the the "logic bytes" end
+    uint256 internal constant DATA_OFFSET_MINUS_ADDRESS = 27;
```
ryley-o, line 35: WDYT about calling this `DEPLOYER_ADDRESS_OFFSET`?

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -175,6 +177,40 @@ library BytecodeStorage {
...
+        returns (address writerAddress)
+    {
+        // get the size of the data
+        uint256 codeSize = _codeSizeAt(_address);
+        // handle case where address contains code < DATA_OFFSET
+        // note: the first check here also captures the case where codeSize == 0
+        //       implicitly, but we add the second check of (codeSize == 0)
+        //       as a fall-through that will never execute unless `DATA_OFFSET`
+        //       is set to 0 at some point.
+        if ((codeSize < DATA_OFFSET) || (codeSize == 0)) {
```
ryley-o, line 197: can we drop the second check after || here?

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -175,6 +177,40 @@ library BytecodeStorage {
...
+        assembly {
+            // allocate free memory
+            writerAddress := mload(0x40)
+            // copy the 20-byte address of the data contract writer to memory
+            extcodecopy(
+                _address,
+                add(writerAddress, 0x20),
+                DATA_OFFSET_MINUS_ADDRESS,
+                20
+            )
```
ryley-o, line 210: I think destOffset should just be `writerAddress` + `12 bytes` (to have left-zero-padding of the address in memory). I think the 0x20 is copied from the other read function, but we only do that because the first byte is the length of the bytes array.
```suggestion
            extcodecopy(
                _address,
                add(writerAddress, 12),
                DATA_OFFSET_MINUS_ADDRESS,
                20
            )
```

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -175,6 +177,40 @@ library BytecodeStorage {
...
+        //       implicitly, but we add the second check of (codeSize == 0)
+        //       as a fall-through that will never execute unless `DATA_OFFSET`
+        //       is set to 0 at some point.
+        if ((codeSize < DATA_OFFSET) || (codeSize == 0)) {
+            revert("ContractAsStorage: Read Error");
+        }
+
+        assembly {
+            // allocate free memory
+            writerAddress := mload(0x40)
```
ryley-o, line 203: writerAddress should already be allocated since it is in the return statement. If we did allocate here, we would also need to increment the free memory pointer by 0x20, but I think we can skip altogether.

ryley-o (approved):
Couple of suggestions - we should probably mock tests for this read capability if we are going to include in our contract.

contracts/libs/0.8.x/BytecodeStorage.sol
ryley-o: NVM - this makes sense based on the dev comments!

contracts/libs/0.8.x/BytecodeStorage.sol
jakerockland: Yeah that makes more sense, minor change so just updating now :)

contracts/libs/0.8.x/BytecodeStorage.sol
jakerockland: 👑 🙏 🐐

jakerockland:
> Couple of suggestions - we should probably mock tests for this read capability if we are going to include in our contract.

Very much agreed, I want to spend more time trying to help w/ QSP before getting to this part, but made your other suggestions here, appreciate your insight ser 🙏 🐐 

Marking this as "do not merge" for now.

jakerockland:
@ryley-o re-requesting review from you if you're willing to do another pass now that mock+tests have been added 🙏 💜

contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -31,9 +32,11 @@ library BytecodeStorage {
     // 53             | 20   | 73           | the 20 bytes used for storing the deploying contract's address         //
     //---------------------------------------------------------------------------------------------------------------//
     // Define the offset for where the "logic bytes" end, and the "data bytes" begin. Note that this is a manually
-    // calculated value, and must be updated if the above table is changed. It is expected that tests will fail if
-    // this value is not updated.
+    // calculated value, and must be updated if the above table is changed. It is expected that tests will fail
+    // loudly if these values are not updated in-step with eachother.
     uint256 internal constant DATA_OFFSET = 73;
+    uint256 internal constant ADDRESS_LENGTH = 20;
+    uint256 internal constant ADDRESS_OFFSET = 53; // DATA_OFFSET - ADDRESS_LENGTH
```
ryley-o, line 39: nit: we could calculate DATA_OFFSET in a constructor now, and declare it immutable.

ryley-o (changes requested):
This is awesome, really great tests for this!

See a couple notes on this - happy to discuss if you think it would help!

contracts/libs/0.8.x/BytecodeStorage.sol
ryley-o: oops - forgot this is a library and no constructors allowed. ignore this comment!

ryley-o (approved):
LGTM!

PR #308, opened 2022-09-24
BytecodeStorage - Require intentional selfdestruct

ryley-o:
Update BytecodeStorage.sol to only selfdestruct if deploying address calls with calldata equal to `0xFF`.

This is a security improvement because the deployed bytecode storage contracts must intentionally be called with specific calldata to invoke a selfdestruct. Without this added logic, if the deployed bytecode were to send funds to a deployed bytecode storage contract, the bytecode storage contract would selfdestruct.

Note that this does not patch any existing hole in the Art Blocks GenArt721CoreV3 contract security model, but it is intended to help encourage security best practices across the Ethereum developer community.


jakerockland (approved):
This LGTM @ryley-o – folding this into #299

jakerockland:
Ah, I _think_ gas tests need to be updated slightly as a result of this @ryley-o ?

PR #311, opened 2022-09-26
QSP-7 - Mitigated - Owner can renounce ownership

ryley-o:
Override `Ownable`'s `renounceOwnership` to enable improved documentation of the function, and to more clearly indicate new projects are forbidden after renouncing ownership.

Note that this change is not intended to change the functionality of adding new projects after an owner renounces ownership (i.e. before this change, the owner address == 0 means new projects will never be added), but rather more clearly indicates that new projects will never be added by setting forbidNewProjects to true (and emitting the corresponding event log).

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -75,6 +75,10 @@ import "./libs/0.8.x/Bytes32Strings.sol";
  * unlocked, and only callable by Admin ACL contract when a project is locked:
  * - updateProjectDescription
  * ----------------------------------------------------------------------------
+ * The following function is restricted to owner calling directly:
+ * - transferOwnership
+ * - renounceOwnership
+ * ----------------------------------------------------------------------------
  * Additional admin and artist privileged roles may be described on minters,
  * registries, and other contracts that may interact with this core contract.
  */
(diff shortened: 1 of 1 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -407,6 +411,29 @@ contract GenArt721CoreV3 is
...
+     * Note that the contract is intended to continue to function after the
+     * owner renounces ownership, but no new projects will be able to be added.
+     * Renouncing ownership will leave the contract without an owner,
+     * thereby removing any functionality that is only available to the
+     * owner/AdminACL contract. The same is true for any dependent contracts
+     * that also integrate with the owner/AdminACL contract (e.g. potentially
+     * minter suite contracts, registry contracts, etc.).
+     * After renouncing ownership, artists will be in control of updates to
+     * their payment addresses and splits (see modifier
+     * onlyAdminACLOrRenouncedArtist`)
```
jakerockland, line 425: I think that this description is great as-is. I wonder if we should even more explicitly add something like:

"While there is no currently intended reason to call this method based on defined Art Blocks business practices, this method exists to allow artists to continue to maintain the limited set of contract functionality that exists post-project-lock in an environment in which there is no longer an admin maintaining this smart contract." 

WDYT, this too verbose/not helpful?

jakerockland (approved):

jakerockland:
I know this is a Draft still, but it _does_ LGTM hehe.

contracts/GenArt721CoreV3.sol
ryley-o: I like it! added and pushed, will merge with these updates.

PR #312, opened 2022-09-26
QSP-8 - Fixed - Update constructor for GenArtCoreV3 to support safe-casting

jakerockland:
This PR addresses audit feedback by updating constructor requirements such that it enforces `nextProjectId` to always fit within an `uint248` before assignment, and perform safe up-casting rather than unsafe down-casting.

## Audit

Response to QSP-10, addresses issue with suggested fix.

---
  - https://app.asana.com/0/0/1203034890733938/f


ryley-o (approved):
LGTM!

PR #315, opened 2022-09-27
QSP-9 - Acknowledged - update documentation about generic events

ryley-o:
Update documentation to improve description of generic event strategy.

---
  - https://app.asana.com/0/0/1203034890733939

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -94,7 +94,11 @@ contract GenArt721CoreV3 is
     uint256 constant FOUR_WEEKS_IN_SECONDS = 2_419_200;
     uint8 constant AT_CHARACTER_CODE = uint8(bytes1("@")); // 0x40
 
-    // generic platform event fields
+    // This contract emits generic events that contain fields that indicate
+    // which parameter has been updated. This is sufficient for application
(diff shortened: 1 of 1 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -94,7 +94,11 @@ contract GenArt721CoreV3 is
     uint256 constant FOUR_WEEKS_IN_SECONDS = 2_419_200;
     uint8 constant AT_CHARACTER_CODE = uint8(bytes1("@")); // 0x40
 
-    // generic platform event fields
+    // This contract emits generic events that contain fields that indicate
+    // which parameter has been updated. This is sufficient for application
+    // state management, while also simplifying the contract and indexing code.
```
jakerockland, line 99: May be worth being even more idiomatic here and specifying: "This was done as an alternative to having custom events that emit what field-values have changed for each event, given that changed values can be introspected by indexers due to the design of this smart contract exposing these state changes via publicly viewable fields." or something like that?

jakerockland (approved):

contracts/GenArt721CoreV3.sol
ryley-o: SGTM! - updated!

PR #319, opened 2022-09-29
Delegation registry integration with merkle-proof minter

jakerockland:
Integration the delegate.cash delegation registry with our merkleminter – co-authored with @mchrupcala, who added the integration testing support here.

Diff between this version (V2) and the V1 merkle minter that it branches from: https://www.diffchecker.com/nZgT92ry

Additional context: 
- https://0xfoobar.substack.com/p/delegatecash 
- https://twitter.com/0xfoobar/status/1575290861570740226?s=20&t=llZHT6mXd_RDYB6CrSrgpw 
- https://github.com/0xfoobar/delegation-registry


jakerockland:
cc @yoshiwarab

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
```diff
@@ -298,7 +309,24 @@ contract MinterMerkleV1 is ReentrancyGuard, IFilteredMinterMerkleV0 {
...
+     * @param _proof Merkle proof.
+     * @param _delegee Delegee being purchased on behalf of.
+     * @return tokenId Token ID of minted token
+     */
+    function purchase(
+        uint256 _projectId,
+        bytes32[] calldata _proof,
+        address _delegee
+    ) external payable returns (uint256 tokenId) {
+        tokenId = purchaseTo_K1L(msg.sender, _projectId, _proof, _delegee);
```
ryley-o, line 329: I think minting to `msg.sender` is okay here, but we should make this behavior very clear on the frontend, because I can definitely see some users thinking the mint will be sent directly to their vault. We might even want to only accept vault to be specified on calls to `purchaseTo` (and assume delegee of zero here).
Honestly, this whole concept of "should a delegate claiming an airdrop/benefit of any kind be the receiver of said benefit" might be a good "best practice" question for the registry dev chat.

ryley-o (commented):
This approach generally LGTM! Very straightforward 😄 

Couple of notes:
- You will have some merge conflicts after #321 is merged
- Our super cool `purchaseTo_K1L` will have to be be updated 😓

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
```diff
@@ -298,7 +309,24 @@ contract MinterMerkleV1 is ReentrancyGuard, IFilteredMinterMerkleV0 {
...
+     *         `msg.sender`) on behalf of an explicitly defined delegee.
+     * @param _projectId Project ID to mint a token on.
+     * @param _proof Merkle proof.
+     * @param _delegee Delegee being purchased on behalf of.
+     * @return tokenId Token ID of minted token
+     */
+    function purchase(
+        uint256 _projectId,
+        bytes32[] calldata _proof,
+        address _delegee
```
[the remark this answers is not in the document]
mchrupcala: +1 I think it's easy to mix up delegee vs. delegate 😅

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
```diff
@@ -365,24 +413,47 @@ contract MinterMerkleV1 is ReentrancyGuard, IFilteredMinterMerkleV0 {
...
+            // than optimistically-minting with original `msg.sender`.
+            // Note, we do not check `checkDelegateForAll` as well, as it is known
+            // to be implicitly checked by calling `checkDelegateForContract`.
+            bool isValidDelegee = delegationRegistryContract
+                .checkDelegateForContract(
+                    msg.sender, // delegate
+                    delegee, // vault
+                    genArt721CoreAddress // contract
+                );
+            require(isValidDelegee, "Invalid delegate-delegee pairing");
```
mchrupcala, line 433: Is this error message something that we'll surface to the end-user? Maybe something like "No match was found for this delegate/vault pair".

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
mchrupcala: The funds for minting are still coming from the delegate, right? If so then I agree it feels to me like our existing purchaseTo() and the delegate.cash purchasing are very similar user flows (difference being the delegation would automatically pull the registered vault address for the user instead of a copy/paste). As such I'm not sure if I see the use case where purchase() would need a `delegee` because implicitly this flow is always meant for sending $$ & receiving art to/from the same wallet.

^This has me wondering if we want to re-work the purchase modal along the lines of "purchase" and then a new screen or section that asks "are you sending this to another wallet?" --> if yes, then a copy/paste uses `purchaseTo()` or if getDelegatesForContract returns 1+ addresses, the delegate sees a list of vaults they can click & send to.

jakerockland:
> You will have some merge conflicts after #321 is merged

Yes yes! Was planning to wait until #321 is landed to do another real pass over updating this PR, and to add test coverage, etc.

> Our super cool purchaseTo_K1L will have to be be updated 😓

Indeed 🥲 I think I will need your assistance (or direction towards the tooling to use) to determine most efficient signature to use here ser.

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
jakerockland: Generally would say that we probably should not design our contract assertion messages to be as detailed/verbose as error messages should be for user-facing frontend applications, as there are reasons to keep bytes used for assertions messages smol-er. 

In the case of surfacing explanation of this type of revert scenario to users, would say that we should catch the error and output more detailed error message as frontend handling itself. :)

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
```diff
@@ -4,12 +4,13 @@
...
 import "../../../interfaces/0.8.x/IMinterFilterV0.sol";
 import "../../../interfaces/0.8.x/IFilteredMinterMerkleV0.sol";
+import "../../../interfaces/0.8.x/IDelegationRegistry.sol";
 
 import "@openzeppelin-4.7/contracts/utils/cryptography/MerkleProof.sol";
 import "@openzeppelin-4.7/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin-4.7/contracts/security/ReentrancyGuard.sol";
 
-pragma solidity 0.8.9;
+pragma solidity 0.8.17;
```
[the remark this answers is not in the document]
jakerockland: Agreed–will bump to 0.8.16 when I do another edit-pass over this PR (leaving this comment unresolved until I address), which I'll do after #321 is merged 👍 💜

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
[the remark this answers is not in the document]
jakerockland: SGTM–will update to rename when I do another edit-pass over this PR (leaving this comment unresolved until I address), which I'll do after #321 is merged 👍 💜

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
jakerockland: Yeah this is a great point–wonder in this case if we may even want to just disallow `purchase` to be used when using this minter via the `delegate-vault` approach, instead _only allowing_ `purchaseTo`, just to simplify the set of possible interactions with this minter to be **only 3 options**:

1) minting as a directly-allowlisted minter (on the merkle list directly) using `purchase` (`msg.sender` receives NFT)
2) minting as a directly-allowlisted minter (on the merkle list directly) using `purchaseTo` (provided `_to` address receives NFT)
3) minting as the delegate for an allowlisted minter (vault is on merkle list, minter is not but is a delegate) using `purchaseTo` (provided `_to` address receives NFT)

>  Honestly, this whole concept of "should a delegate claiming an airdrop/benefit of any kind be the receiver of said benefit" might be a good "best practice" question for the registry dev chat.

This is a great point–going to post this convo there for any additional thoughts, and will also let this bop around in my brain a bit more over the weekend before I loop back and update this PR at all to account for this discussion and the others mentioned here.

ryley-o:
>Indeed 🥲 I think I will need your assistance (or direction towards the tooling to use) to determine most efficient signature to use here ser.

happy to help, feel free to play around at: https://emn178.github.io/solidity-optimize-name/

(I always double check the signature in remix though because that frontend allows some entry of invalid solidity 🤷 e.g. use `uint256` not `uint`, but that frontend incorrectly allows `uint`)

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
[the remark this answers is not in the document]
jakerockland: Resolving now that we're on 0.8.17 anyway 😄

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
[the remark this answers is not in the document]
jakerockland: Updated.

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
jakerockland: Updating to just support these 3 routes: `purchase`, `purchaseTo`, and `purchaseTo` as a delegate–I think this is the most explicit pattern and allows flexibility on the frontend for us to iterate on UX to make it as clear as possible.

jakerockland:
> > Indeed 🥲 I think I will need your assistance (or direction towards the tooling to use) to determine most efficient signature to use here ser.
> 
> happy to help, feel free to play around at: https://emn178.github.io/solidity-optimize-name/
> 
> (I always double check the signature in remix though because that frontend allows some entry of invalid solidity 🤷 e.g. use `uint256` not `uint`, but that frontend incorrectly allows `uint`)

Huzzah! Thank you for the tip! Updating to `purchaseTo_kem` based on tweakin w/ that tool 🙌 💜

jakerockland:
@mchrupcala @ryley-o still have not yet had a chance to update to add tests, but did take a pass over incorporating the other feedback in this PR–if I have any extra time during Devcon week next week will try to get test coverage added so that we can get this in a more merge-ready state for real review here.

jakerockland:
Also, if either of you would like to take on getting this across the finish line in that regard, please do not block on me–that said, please also do not let this derail other priorities that are non-Merkle related, this is a stretch goal after all 💜 😄

jakerockland:
@mchrupcala as part of taking this PR on, I would recommend that we bump the version on the MerkleMinter from V1 to V2, given that we've now deployed something to mainnet with the V1 nomenclature that does not include this functionality. 

cc @ryley-o for his thoughts as well to see if there is a difference of opinion from his perspective.

jakerockland:
Additional context re: MerkleMinterV1 being live on mainnet here: https://github.com/ArtBlocks/artblocks-contracts/pull/337

🙏 💜

test/minter-suite-minters/MinterMerkle/MinterMerkleV1.test.ts
```diff
@@ -165,6 +171,11 @@ describe("MinterMerkleV1", async function () {
     // mock ERC20 genArt721Core
     const ERC20Factory = await ethers.getContractFactory("ERC20Mock");
     this.ERC20Mock = await ERC20Factory.deploy(ethers.utils.parseEther("100"));
+
+    fakeDelegationRegistry = await smock.fake('IDelegationRegistry', {
+      address: '0x00000000000076A84feF008CDAbe6409d2FE638B'
```
mchrupcala, line 176: @ryley-o found a great pattern using [Smock](https://smock.readthedocs.io/en/latest/fakes.html#options) so that we can mock `checkDelegateForContract`. In the future we can prob use this to mock method calls in our unit tests whenever we have an external contract dependency.

package.json
```diff
@@ -51,7 +52,7 @@
     "eslint": "^8.25.0",
     "ethereum-waffle": "^3.4.0",
     "ethers": "^5.4.1",
-    "hardhat": "^2.10.2",
+    "hardhat": "^2.12.0",
```
mchrupcala, line 55: Wanted to call out the Hardhat version upgrade - I had to switch versions to fix a typeerror that Smock was giving me (`Cannot read properties of undefined (reading 'events')`) but please lmk if we wanted to stay consistent with the earlier version 🙏🏻

test/libs/BytecodeStorage_BytecodeTextCR_DMock.test.ts
```diff
@@ -316,17 +316,19 @@ describe("BytecodeStorage + BytecodeTextCR_DMock Library Tests", async function
       expect(deployedBytecode).to.not.equal("0x");
 
       // Non-writer addresses should **not** be able to purge bytecode storage.
-      await expectRevert.unspecified(
+      await expectRevert(
```
mchrupcala, line 319: One more callout- `expectRevert.unspecified()` stopped working for me (I'm guessing that might be due to the Hardhat upgrade, although this util's from OpenZeppelin). Adding a revert error seemed to do the trick!

mchrupcala:
> @mchrupcala as part of taking this PR on, I would recommend that we bump the version on the MerkleMinter from V1 to V2, given that we've now deployed something to mainnet with the V1 nomenclature that does not include this functionality.
> 
> cc @ryley-o for his thoughts as well to see if there is a difference of opinion from his perspective.

Bumped! I switched this to `MinterMerkleV2` everywhere I could find a reference.

hardhat.config.ts
```diff
@@ -54,6 +54,15 @@ module.exports = {
...
       },
+      {
+        version: "0.8.17",
+        settings: {
+          optimizer: {
+            enabled: true,
+            runs: 25,
+          },
+        },
+      },
```
jakerockland, line 65: I don't think this is needed.

README.md
```diff
@@ -84,7 +84,7 @@ The following table shows which Minters and MinterFilters are compatible with wh
 | V0                       | -              | (legacy minter)                                                                                                 | -                                                                           |
 | V1, V1_PRTNR             | MinterFilterV0 | MinterSetPriceV1<br>MinterSetPriceERC20V1<br>MinterDAExpV1<br>MinterDALinV1<br>MinterMerkleV0<br>MinterHolderV0 | MinterSetPriceV0<br>MinterSetPriceERC20V0<br>MinterDAExpV0<br>MinterDALinV0 |
 | V2 (PBAB)                | -              | (various PBAB minters)                                                                                          | -                                                                           |
-| V3                       | MinterFilterV1 | MinterSetPriceV2<br>MinterSetPriceERC20V2<br>MinterDAExpV2<br>MinterDALinV2<br>MinterMerkleV1<br>MinterHolderV1 | -                                                                           |
+| V3                       | MinterFilterV1 | MinterSetPriceV2<br>MinterSetPriceERC20V2<br>MinterDAExpV2<br>MinterDALinV2<br>MinterMerkleV2<br>MinterHolderV1 | -                                                                           |
```
jakerockland, line 87: This diff seems off to me–could you clarify please?

package.json
jakerockland: This SGTM 🙏

README.md
mchrupcala: Oh sorry, I was assuming since we haven't deployed core V3 to prod then our V1 minter should be replaced by V2, instead of listing both in the ReadMe. Is it better if I list both V1 and V2 instead of replacing this one?

hardhat.config.ts
jakerockland: (reason being that this is already duplicated above)

test/minter-suite-minters/MinterMerkle/MinterMerkleV1.test.ts
jakerockland: 🙌

hardhat.config.ts
jakerockland: Removed.

test/minter-suite-minters/MinterMerkle/MinterMerkleV2.test.ts
```diff
@@ -165,6 +174,11 @@ describe("MinterMerkleV1", async function () {
     // mock ERC20 genArt721Core
     const ERC20Factory = await ethers.getContractFactory("ERC20Mock");
     this.ERC20Mock = await ERC20Factory.deploy(ethers.utils.parseEther("100"));
+
+    fakeDelegationRegistry = await smock.fake("IDelegationRegistry", {
+      address: "0x00000000000076A84feF008CDAbe6409d2FE638B",
+    });
+    fakeDelegationRegistry.checkDelegateForContract.returns(true);
```
jakerockland, line 181: @mchrupcala rather than adding this as a shared before-each assumption, can we please just set this assumption on a per-test level to make things more explicit / more in alignment with assumptions of each individual test? I think this will improve read-ability.

jakerockland (commented):
One outstanding item @mchrupcala and then I think we're g2g re: merging this and beginning to follow up on the minor frontend pieces here.

test/minter-suite-minters/MinterMerkle/MinterMerkleV2.test.ts
mchrupcala: Agreed! And this way the next time somebody writes a delegation test they won't inherit the `.returns(true)`. One more note- I added a comment to mention the deployed address we're mocking for the delegation contract.

Would you like me to go ahead and merge this? I'll request another round of review from Ryley in the meantime for when he's back.

mchrupcala (approved):

test/minter-suite-minters/MinterMerkle/MinterMerkleV2.test.ts
jakerockland: Fantastic–yep this LGTM to merge. 🙏

jakerockland:
Merging this @mchrupcala – I think we're ready to take on frontend changes necessary for this and to test this end-to-end on Goerli once that is done 🔥 💜 

Great work here!!

PR #320, opened 2022-09-29
Contract bytecode for script storage – simplify purge logic

jakerockland:
This PR is an implementation-detail adjustment to #299 (and intended to be merged into it), which simplifies the purge-gating logic to use fewer assembly instructions (see: https://ethereum.org/en/developers/docs/evm/opcodes/) at the expense of a very minor (25-byte) increase to the storage used by the deployed bytecode-as-storage contract.

The main delta here in the purge-gating logic is in (3a), where this implementation is updated to use a `PUSH32` op to push the value of `0xFF...00` (32-byte word) to the stack directly when comparing this against the expected call-data of a single-byte `0xFF`, rather than loading and reading the value of `0xFF...00` in and out of memory. 

This is necessary as the `EQ` operator performs a "(u)int256 equality" check and `CALLDATALOAD` loads a 32-byte word from current execution contract storage, meaning that if one tries to push a single-byte value of `0xFF` to the stack, and then compare it against the single-byte expected call-data, the value of `0x00...FF` (pushed to stack) will be compared to `0xFF...00` (added to stack from `CALLDATALOAD`) and not give us the comparison that we would like to make.

As such, it is important that we validate in our logic separately that **both** the `CALLDATASIZE` is _exactly `1`_ and that the call-data is `0xFF` (which requires that we check against `0xFF...00`). By pushing `0xFF...00` directly onto the stack, this logic becomes easier to rationalize about than it was with the previous implementation which shifted around `0xFF` in memory to be able to be loaded out of memory with `0x00`-padding.

Additionally, this PR adds an additional optimization in the first check, (1a), in the purge-gating logic, to just reference the `CALLER` directly from the stack when making the initial `(msg.sender == byteDeployerAddress)` comparison rather than `MSTORE`ing this into memory and then `MLOAD`ing it back out, which shaves off 6 op-codes.

This PR is entirely a functional no-op, which is validated by unit tests now that we have comprehensive mock/test coverage as of #307. Note that I validated that tests failed as expected during the development of this PR, as offsets were shifted around, which is a nice confidence-add to the changes in #307.


contracts/libs/0.8.x/BytecodeStorage.sol
```diff
@@ -70,79 +70,93 @@ library BytecodeStorage {
...
-            // 0x60    |  0x60_00            | PUSH1 0      | 0                                                              //
-            // 0x51    |  0x51               | MLOAD        | byteDeployerAddress                                            //
-            // 0x60    |  0x60_20            | PUSH1 32     | 32 byteDeployerAddress                                         //
-            // 0x51    |  0x51               | MLOAD        | msg.sender byteDeployerAddress                                 //
-            // 0x14    |  0x14               | EQ           | (msg.sender == byteDeployerAddress)                            //
-            hex"60_14_60_35_60_0C_39_33_60_20_52_60_00_51_60_20_51_14",
+            // 0x60    |  0x60_14            | PUSH1 20           | 20                                                       //
+            // 0x60    |  0x60_4E            | PUSH1 76 (*)       | contractOffset 20                                        //
+            // 0x60    |  0x60_0C            | PUSH1 12           | 12 contractOffset 20                                     //
```
ryley-o, line 82: Should we do the same thing here? (i.e. zero-pad the address in bytecode instead of using an offset)

ryley-o (approved):
LGTM! one comment about doing this for addresses - whatever you think SGTM!

contracts/libs/0.8.x/BytecodeStorage.sol
jakerockland: I don't know that we can do that IIUC, given how `CODECOPY` works (we need to copy into memory) but I am realizing that there is a place here where we can further optimize the assembly logic here. 

Pushing another update shortly here and will re-request a review – would love another look with the additional tweaks.

jakerockland: Added additional optimization to just reference the `CALLER` directly from the stack when making the initial `(msg.sender == byteDeployerAddress)` comparison rather than `MSTORE`ing this into memory and then `MLOAD`ing it back out, which shaves off 6 op-codes.

jakerockland:
https://github.com/ArtBlocks/artblocks-contracts/pull/320/commits/24a3475c0ac914d07c39e4ad1b1e2b1244013f4d has been added to adjust for conversation that spun out of https://github.com/ArtBlocks/artblocks-contracts/pull/320#discussion_r984110263

Ready for another look from your end when you get a chance @ryley-o (cc @shantanubala PTAL if you have bandwidth/interest, but no pressure to review).

ryley-o (approved):
LGTM!

PR #321, opened 2022-09-30
Merkle minter - Configurable qty per wallet address

ryley-o:
Update V0 and V1 of our Merkle minters to allow the artist to define a qty of allowed mints per address.

Values on the V1 minter are packed to keep gas impacts of this change minimal.

## Tests
Tests are updated for equivalent coverage of the new functionality:
- **MinterMerkleV1**: 100% statement, 100% branches, 100% functions, 100% lines
- **MinterMerkleV0**: 95% statement, 79% branches, 100% functions, 95% lines
>MinterMerkleV0 not planned to be used in production environment

## Other
V0 minter pulls an update from the V1 minter that validates a Merkle root != zero when artist is assigning.

## Follow-on
- Subgraph should be updated with the new MerkleMinter ABIs, and additional routing for generic event emitting a single uint256 should be added to the subgraph config.
  - see https://github.com/ArtBlocks/artblocks-subgraph/pull/113
- Frontend updates:
  - New MerkleMinter ABIs
  - Display qty of mints used and allowed for connected user
    - see `projectUserMintInvocations ` and `projectMaxInvocationsPerAddress` view functions
  - Update artist project configure to allow artists to configure the new mint limit per address value

### DIFF
--- a/contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV0.sol
+++ b/contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV0.sol
@@ -36,18 +36,22 @@ contract MinterMerkleV0 is ReentrancyGuard, IFilteredMinterMerkleV0 {
 
     /// project minter configuration keys used by this minter
     bytes32 private constant CONFIG_MERKLE_ROOT = "merkleRoot";
-    bytes32 private constant CONFIG_MINT_LIMITER_DISABLED =
-        "mintLimiterDisabled";
+    bytes32 private constant CONFIG_USE_MAX_INVOCATIONS_PER_ADDRESS_OVERRIDE =
+        "useMaxMintsPerAddrOverride"; // shortened to fit in 32 bytes
+    bytes32 private constant CONFIG_MAX_INVOCATIONS_OVERRIDE =
+        "maxMintsPerAddrOverride"; // shortened to match format of previous key
 
     uint256 constant ONE_MILLION = 1_000_000;
 
+    uint256 public constant DEFAULT_MAX_INVOCATIONS_PER_ADDRESS = 1;
+
     /// projectId => merkle root
     mapping(uint256 => bytes32) public projectMerkleRoot;
-    /// projectId => purchaser address => has purchased one or more mints
-    mapping(uint256 => mapping(address => bool)) public projectMintedBy;
-    /// projectId => may a single address mint multiple times?
-    /// (default behavior is limit one mint per address)
-    mapping(uint256 => bool) public projectMintLimiterDisabled;
+    /// projectId => are max invocations per address overrides enabled
+    /// (default is false, so DEFAULT_MAX_INVOCATIONS_PER_ADDRESS per address)
+    mapping(uint256 => bool) public projectUseMaxInvocationsPerAddressOverride;
+    /// projectId => max invocations per address override
+    mapping(uint256 => uint256) public projectMaxInvocationsPerAddressOverride;
     /// projectId => has project reached its maximum number of invocations?
     mapping(uint256 => bool) public projectMaxHasBeenInvoked;
     /// projectId => project's maximum number of invocations
@@ -56,6 +60,10 @@ contract MinterMerkleV0 is ReentrancyGuard, IFilteredMinterMerkleV0 {
     mapping(uint256 => uint256) private projectIdToPricePerTokenInWei;
     /// projectId => price per token has been configured on this minter
     mapping(uint256 => bool) private projectIdToPriceIsConfigured;
+
+    /// projectId => purchaser address => qty of mints purchased for project
+    mapping(uint256 => mapping(address => uint256))
+        public projectUserInvocations;
 
     modifier onlyCoreWhitelisted() {
         require(
@@ -143,21 +151,36 @@ contract MinterMerkleV0 is ReentrancyGuard, IFilteredMinterMerkleV0 {
     }
 
     /**
-     * @notice Toggles mint limit of one per address for project `_projectId`.
-     * If mint limit is disabled, unlimited mints per address are allowed.
+     * @notice Sets maximum allowed invocations per allowlisted address for
+     * project `_project` to `limit`. If `limit` is set to 0, infinite mints
+     * will be allowed.
+     * Default is a value of 1 if never configured by artist.
      * @param _projectId Project ID to toggle the mint limit.
+     * @param _maxInvocationsPerAddress Maximum allowed invocations per
+     * allowlisted address.
+     * @dev default value stated above must be updated if the value of
+     * CONFIG_USE_MAX_INVOCATIONS_PER_ADDRESS_OVERRIDE is changed.
      */
-    function toggleProjectMintLimiter(uint256 _projectId)
-        external
-        onlyArtist(_projectId)
-    {
-        projectMintLimiterDisabled[_projectId] = !projectMintLimiterDisabled[
+    function setProjectInvocationsPerAddress(
+        uint256 _projectId,
+        uint24 _maxInvocationsPerAddress
+    ) external onlyArtist(_projectId) {
+        // use override value instead of the contract's default
+        projectUseMaxInvocationsPerAddressOverride[_projectId] = true;
+        // update the override value
+        projectMaxInvocationsPerAddressOverride[
             _projectId
-        ];
+        ] = _maxInvocationsPerAddress;
+        // generic events
         emit ConfigValueSet(
             _projectId,
-            CONFIG_MINT_LIMITER_DISABLED,
-            projectMintLimiterDisabled[_projectId]
+            CONFIG_USE_MAX_INVOCATIONS_PER_ADDRESS_OVERRIDE,
+            true
+        );
+        emit ConfigValueSet(
+            _projectId,
+            CONFIG_MAX_INVOCATIONS_OVERRIDE,
+            uint256(_maxInvocationsPerAddress)
         );
     }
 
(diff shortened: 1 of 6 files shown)


ryley-o:
Cc @mchrupcala - see the Follow-on work for the frontend updates in the PR's description 🎉

mchrupcala (approved):
Looks great to me!

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV0.sol
```diff
@@ -277,15 +301,20 @@ contract MinterMerkleV0 is ReentrancyGuard, IFilteredMinterMerkleV0 {
...
+        uint256 _maxProjectInvocationsPerAddress = projectUseMaxInvocationsPerAddressOverride[
+                _projectId
+            ]
+                ? projectMaxInvocationsPerAddressOverride[_projectId]
+                : DEFAULT_MAX_INVOCATIONS_PER_ADDRESS;
+
+        // FINAL CHECK, WITH FOLLOW-ON EFFECTS
+        require(
+            // _++ returns current invocations, then increments
+            projectUserInvocations[_projectId][msg.sender]++ <
```
jakerockland, line 313: Nit: Does doing incrementing and reading in this way having meaningful gas savings associated with it? I find this more confusing to read than separating out the ops re: order of operations here and wonder if it is worth keeping more readable.

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
```diff
@@ -452,6 +471,28 @@ contract MinterMerkleV1 is ReentrancyGuard, IFilteredMinterMerkleV0 {
...
+    /**
+     * @notice projectId => maximum invocations per allowlisted address. If a
+     * a value of 0 is returned, there is no limit on the number of mints per
+     * allowlisted address.
+     * Default behavior is limit 1 mint per address.
+     * This value can be changed at any time by the artist.
+     * @dev default value stated above must be updated if the value of
+     * CONFIG_USE_MAX_INVOCATIONS_PER_ADDRESS_OVERRIDE is changed.
+     */
+    function projectMaxInvocationsPerAddress(uint256 _projectId)
```
jakerockland, line 483: I think we for convenience probably want to also provide a similar view method that takes in an address as an input and returns the # of remaining for a given specific address?

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV0.sol
```diff
@@ -360,6 +389,27 @@ contract MinterMerkleV0 is ReentrancyGuard, IFilteredMinterMerkleV0 {
...
+    /**
+     * @notice projectId => maximum invocations per allowlisted address. If a
+     * a value of 0 is returned, there is no limit on the number of mints per
+     * allowlisted address.
+     * Default behavior is limit 1 mint per address.
+     * This value can be changed at any time by the artist.
+     * @dev default value stated above must be updated if the value of
+     * CONFIG_USE_MAX_INVOCATIONS_PER_ADDRESS_OVERRIDE is changed.
+     */
+    function projectMaxInvocationsPerAddress(uint256 _projectId)
```
jakerockland, line 401: I think we for convenience probably want to also provide a similar view method that takes in an address as an input and returns the # of remaining for a given specific address?

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
```diff
@@ -372,15 +387,19 @@ contract MinterMerkleV1 is ReentrancyGuard, IFilteredMinterMerkleV0 {
...
-        }
+        uint256 _maxProjectInvocationsPerAddress = _projectConfig
+            .useMaxInvocationsPerAddressOverride
+            ? _projectConfig.maxInvocationsPerAddressOverride
+            : DEFAULT_MAX_INVOCATIONS_PER_ADDRESS;
+
+        // FINAL CHECK, WITH FOLLOW-ON EFFECTS
+        require(
+            // _++ returns current invocations, then increments
+            projectUserInvocations[_projectId][msg.sender]++ <
```
jakerockland, line 398: Nit: Does doing incrementing and reading in this way having meaningful gas savings associated with it? I find this more confusing to read than separating out the ops re: order of operations here and wonder if it is worth keeping more readable.

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
```diff
@@ -149,21 +161,36 @@ contract MinterMerkleV1 is ReentrancyGuard, IFilteredMinterMerkleV0 {
     }
 
     /**
-     * @notice Toggles mint limit of one per address for project `_projectId`.
-     * If mint limit is disabled, unlimited mints per address are allowed.
+     * @notice Sets maximum allowed invocations per allowlisted address for
+     * project `_project` to `limit`. If `limit` is set to 0, infinite mints
```
jakerockland, line 165: Nit: probably don't want to use the word "infinite" without more clarification (only up to allowed # invocations for the project can be minted).

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
```diff
@@ -68,8 +78,10 @@ contract MinterMerkleV1 is ReentrancyGuard, IFilteredMinterMerkleV0 {
 
     /// projectId => merkle root
     mapping(uint256 => bytes32) public projectMerkleRoot;
-    /// projectId => purchaser address => has purchased one or more mints
-    mapping(uint256 => mapping(address => bool)) public projectMintedBy;
+
+    /// projectId => purchaser address => qty of mints purchased for project
+    mapping(uint256 => mapping(address => uint256))
+        public projectUserInvocations;
```
jakerockland, line 84: Nit: consider naming this: projectUserMintInvocations

jakerockland (approved):

jakerockland:
Some nits / topics for bit of minor conversation, but generally LGTM @ryley-o

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV0.sol
```diff
@@ -57,6 +61,10 @@ contract MinterMerkleV0 is ReentrancyGuard, IFilteredMinterMerkleV0 {
     /// projectId => price per token has been configured on this minter
     mapping(uint256 => bool) private projectIdToPriceIsConfigured;
 
+    /// projectId => purchaser address => qty of mints purchased for project
+    mapping(uint256 => mapping(address => uint256))
+        public projectUserInvocations;
```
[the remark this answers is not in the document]
ryley-o: updated!

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
ryley-o: updated!

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV0.sol
```diff
@@ -142,21 +151,36 @@ contract MinterMerkleV0 is ReentrancyGuard, IFilteredMinterMerkleV0 {
     }
 
     /**
-     * @notice Toggles mint limit of one per address for project `_projectId`.
-     * If mint limit is disabled, unlimited mints per address are allowed.
+     * @notice Sets maximum allowed invocations per allowlisted address for
+     * project `_project` to `limit`. If `limit` is set to 0, infinite mints
```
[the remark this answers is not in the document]
ryley-o: good call, updated!

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
ryley-o: good call, updated!

mchrupcala:
Quick question for you guys: is there any benefit to setting the default to no limit? Right now I'm defaulting to 1 with some helper text (but it feels slightly counter-intuitive). But i could see another flow where there's no limit by default, and then artists see an optional "set limit" field that they can ignore if they don't want one.

Then again if the most common use-case is 1 mint then we'll save the most people money with a default of 1. I don't have a strong opinion either way- just wanted to get your thoughts!

Screen Shot 2022-09-30 at 5 47 50 PM

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV0.sol
```diff
@@ -142,21 +151,36 @@ contract MinterMerkleV0 is ReentrancyGuard, IFilteredMinterMerkleV0 {
...
-            projectMintLimiterDisabled[_projectId]
+            CONFIG_USE_MAX_INVOCATIONS_PER_ADDRESS_OVERRIDE,
+            true
+        );
+        emit ConfigValueSet(
+            _projectId,
+            CONFIG_MAX_INVOCATIONS_OVERRIDE,
+            uint256(_maxInvocationsPerAddress)
         );
     }
```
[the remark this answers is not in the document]
ryley-o: added some inline documentation indicating that the value never changes from true to false, and the default value is only used if the artist has never configured their project invocations per address

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV0.sol
ryley-o: might be some gas savings here, but very small. broke out for improved readability, and also put increment in unchecked block, with comment `// this will never overflow since user's invocations on a project are limited by the project's max invocations`, which should be a net-reduction in mint gas cost.

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
ryley-o: (same response as above - broke out to separate statement for readability)

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
ryley-o: I've added a new function `projectMaxInvocationsPerAddress`. It's a little... weird, just because we have to return a bool indicating if the project limits the qty of mints allowed per allowlisted address, then also has to return qty of mints remaining. I think still worth it though.

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV0.sol
ryley-o: (see other response for same comment in MinterMerkleV1)

ryley-o:
>Quick question for you guys: is there any benefit to setting the default to no limit? Right now I'm defaulting to 1 with some helper text (but it feels slightly counter-intuitive). But i could see another flow where there's no limit by default, and then artists see an optional "set limit" field that they can ignore if they don't want one.
>
>Then again if the most common use-case is 1 mint then we'll save the most people money with a default of 1. I don't have a strong opinion either way- just wanted to get your thoughts!

I think defaulting to 1 is probably less footgun-ish than any other default, since it is the most restrictive. I also think it will be a pretty commonly selected option, but 🤷 . Going to go ahead and ship this with a default value of 1 to get this across the finish line on testnet for our community testing, but can always change our minds down the line!

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV1.sol
jakerockland: This seems like the right approach to me–TY ser! 🙏

PR #325, opened 2022-10-03
Add initial flex deployment details.

jakerockland:
Adds deployment details for https://github.com/ArtBlocks/artblocks/issues/275 and https://github.com/ArtBlocks/artblocks/issues/287, our first 2 partner Engine Flex deployments (on Goerli).


lyaunzbe (dismissed):
overall looks good, but I think  to avoid confusion down the line it would make to continue using the format of previous script deployment folders where we add the network to the filename for non mainnet deploys, ie `1_bright_moments_flex_deployer_goerli.ts`

https://github.com/ArtBlocks/artblocks-contracts/tree/main/scripts/PBAB%2BCollabs/bright-moments

jakerockland:
> overall looks good, but I think to avoid confusion down the line it would make to continue using the format of previous script deployment folders where we add the network to the filename for non mainnet deploys, ie `1_bright_moments_flex_deployer_goerli.ts`
> 
> https://github.com/ArtBlocks/artblocks-contracts/tree/main/scripts/PBAB%2BCollabs/bright-moments

Actually was intentionally not using this convention, as we have logic within the scripts themselves that are network specific. IIRC, the reason we have some `_goerli` appends, was from the fact that we had to do a bulk reset for testnets when we switched to Goerli, but before that we had just used a single script and updated config params within it based on network.

jakerockland:
Happy to coalesce around a new convention here though if we like the appending, I just think it may make things a bit more complicated if we go that route as I would say that if we do so we should stop having network-specific logic within the scripts themselves (they should assume a network if it is defined in the file). WDYT sers? @lyaunzbe @ryley-o

ryley-o (approved):
Agree that we should come up with a better organization/naming convention for our scripts. The scripts are network independent, but they have hardcoded randomizer addresses, payment addresses, and sometimes different token/ticker names, which often are network-specific-values.

Don't think it needs to be solved here though 👍

jakerockland:
Agree @ryley-o that here we don't need to fully solve convention issue, but after chatting more w/ @lyaunzbe I do think that using `_goerli` convention in file names for the deployer scripts being added here is probably the clearer pattern to stick w/ in terms of _current convention_. 

Punting the redefining conventions conversation to a later date, I've updated this PR to add the `_goerli` to have internal consistency about things for now based on recent similar things we've done. 

Also, from chatting more w/ Ben on this, I do think this is also just clearly a better option of the two between a) my old arbitrary convention and b) the more recent convention that is still a bit arbitrary. Until we have a better-yet fully thought out convention, I think we're picking the better of two quick and dirty options with this convention 😅 😂 🙏

PR #326, opened 2022-10-04
Bytecode script storage memory safety

jakerockland:
This adds onto #299 by adding a tighter guarantee around making sure of the fact that the first memory slot used (first 32 bytes) is fully empty (not containing any arbitrary data stored in this scratch memory somehow) before using it for storing the address read from the deployed bytecode-as-storage contract, which is used for comparison purposes against the `CALLER` to gate purging.

Note that this has a minor impact on the size of the deployed contract, but that this is reasonable given that this is relatively trivial (see the _very minor_ test impact) and gives a stronger guarantee around how the deletion logic in this library works.


jakerockland:
@ryley-o adding you for review here for visibility, but I actually have a different idea here that I think is a better approach, which I will send out later tonight when I get home from some outstanding commitments this evening, which I think is a better path forward here that achieves the same result at a lower storage cost and also allows us to simplify our generated bytecode further–approach being to just store the deployer address as a 12-`0x00`-byte padded value 😄

jakerockland:
> –approach being to just store the deployer address as a 12-`0x00`-byte padded value 😄

This will allow us to skip this extra memory writing logic altogether while also simplifying our `CODECOPY` logic in a very minor way.

jakerockland:
Converting back to draft mode while I work on these changes: https://github.com/ArtBlocks/artblocks-contracts/pull/326#issuecomment-1266236444

jakerockland:
Closing in favor of #327.

PR #328, opened 2022-10-05
Core ENGINE_FLEX Interface

lyaunzbe:
this PR:

- creates unified interface for Engine Flex contracts
- uses this newly created interface in current core engine flex contract


jakerockland:
@lyaunzbe @ryley-o do we still need this or can we close as obsolete at this point given we've shifted to V3?

lyaunzbe:
i think should probably be okay to close given we have a newer v3 interface for flex and don't expect to create additional contracts that inherit/share this v2 flex interface (clients wanting to interface with existing v2 flex deployments can just continue use the v2 flex contract ABI directly)

jakerockland:
> i think should probably be okay to close given we have a newer v3 interface for flex and don't expect to create additional contracts that inherit/share this v2 flex interface (clients wanting to interface with existing v2 flex deployments can just continue use the v2 flex contract ABI directly)

Roger roger – closing for now!

PR #335, opened 2022-10-08
Auto accept artist proposals that do not change payment addresses

ryley-o:
Update core V3 contract to automatically accept artist proposed updates that do not change artist payment addresses.

This allows an artist to immediately update (only) their payment address percent splits without admin approval.
Thank you @sarahrossien for bringing this to my attention today!

## Other Changes
On-chain validation is added that prevents artists from directing funds to the zero address. This should be done regardless of if auto-approval is implemented.

## Tests
Tests added to maintain 100% coverage for V3 core contract.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -712,16 +712,13 @@ contract GenArt721CoreV3 is
             _additionalPayeeSecondarySalesPercentage
         );
         // automatically accept if no proposed addresses modifications
-        bool automaticAccept; // default false
-        if (
+        bool automaticAccept = (
             (_artistAddress == projectFinance.artistAddress) &&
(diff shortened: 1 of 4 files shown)


jakerockland (approved):
This LGTM Ryley!

I had some minor suggested changes, but in interest of time (re: deployment) I've added them via commit rather than commenting inline. 

All of my changes should be functionally a no-op, just restructuring things a bit in a way that is more condensed and hopefully a bit more readable. 

PTAL and do flag if there are areas worth double clicking, but LGTM to merge ser. 

Thank you for the quick PR here.

ryley-o:
@jakerockland - TY for simplifying the logic + added comments. This LGTM, enabling auto-merge, and will ensure tests are updated to be merged 👍

PR #338, opened 2022-10-10
auto-approve proposals that only remove addresses

ryley-o:
Update V3 core to also auto-approve artist proposals that only remove additional payment addresses.

This is a nice-to-have for V3 core, but not required. Just thought of it this morning, so very last-minute.

This pattern handles an artist wanting to remove an additional payee instead of just setting the corresponding percentage to zero.

Tests are updated for full coverage.

### DIFF
--- a/contracts/GenArt721CoreV3.sol
+++ b/contracts/GenArt721CoreV3.sol
@@ -59,7 +59,7 @@ import "./libs/0.8.x/Bytes32Strings.sol";
  *   to the Admin ACL contract, or the artist if the core contract owner has
  *   renounced ownership. Also note that a proposal will be automatically
  *   accepted if the artist only proposes changed payee percentages without
- *   modifying any payee addresses.)
+ *   modifying any payee addresses, or is only removing payee addresses.)
  * - toggleProjectIsPaused (note the artist can still mint while paused)
  * - updateProjectSecondaryMarketRoyaltyPercentage (up to
      ARTIST_MAX_SECONDARY_ROYALTY_PERCENTAGE percent)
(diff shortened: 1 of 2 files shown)


contracts/GenArt721CoreV3.sol
```diff
@@ -711,23 +713,31 @@ contract GenArt721CoreV3 is
...
         bool automaticAccept = ((_artistAddress ==
             projectFinance.artistAddress) &&
             (_additionalPayeePrimarySales ==
-                projectFinance.additionalPayeePrimarySales) &&
+                projectFinance.additionalPayeePrimarySales ||
+                _additionalPayeePrimarySales == address(0)) &&
             (_additionalPayeeSecondarySales ==
-                projectFinance.additionalPayeeSecondarySales));
+                projectFinance.additionalPayeeSecondarySales ||
+                _additionalPayeeSecondarySales == address(0)));
```
jakerockland, line 726: Nit: At this point, I'm tempted to break each of these out into their own bool, for readability purposes, as _I would think_ that this can be easily optimized away to be the same thing by the compiler anyway.

E.g.:

```
bool artistUnchanged = _artistAddress == projectFinance.artistAddress;
bool additionalPrimaryPayeeUnchanged = 
    (_additionalPayeePrimarySales == projectFinance.additionalPayeePrimarySales) ||
        (_additionalPayeePrimarySales == address(0));
bool additionalSecondaryPayeeUnchanged = 
     (_additionalPayeeSecondarySales == projectFinance.additionalPayeeSecondarySales) ||
        (_additionalPayeeSecondarySales == address(0));
bool automaticAccept = 
    artistUnchanged && additionalPrimaryPayeeUnchanged && additionalSecondaryPayeeUnchanged;
```

jakerockland (approved):
LGTM ser 🙏

contracts/GenArt721CoreV3.sol
ryley-o: Agree on this, much clearer!
Did cause a compiler stack-to-deep error when implemented as written above, so added block scoping to avoid. Also means optimizer is probably not doing a great job here, but I still think well worth it!

PR #339, opened 2022-10-10
Improve CI

ryley-o:
Improve CI by using parallelism in circleci, and improved linting to ensure no-floating-promises in tests (+ exposed test bug fixes)

This is a fallout of investigating how best to publish docs during CI. It will help us scale and keep CI times down when adding more V3 derivative contracts to our repo.

Improve CI:
- speed up tests via circleci parallelism (8 instances) 
  - brings down CI from >21 minutes to ~5 min 30 sec
- improve linting
  - eslint to ensure no floating promises in tests
    - this exposed many non-deterministic tests, some of which could potentially hide failing tests
  - all tests fixed to pass
  - one bug was exposed, but corrected, and no issues with the contracts (was using an incorrect contract, changed to desired contract, and ensured no coverage impacts)

## Design Choices
8 instances of parallelism were selected on CircleCI. 10 instances took ~5.5 mins, 5 instances took ~7 minutes, so settled on using 8 instances, which takes ~5.5 mins. If we have credit issues on CircleCI, we can use slightly less compute time by reducing parallelism, but for now it seems well worth it.

## Alternatives Considered
Using `hardhat test --parallel` was investigated, but due to [non-deterministic test ordering and long timeouts](https://hardhat.org/hardhat-runner/docs/guides/test-contracts#running-tests-in-parallel), the tests became flakey. Running hardhat in parallel mode also did not speed up tests much (~10%), so the added issues of variability were not worth the repeated issues that kept popping up.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
           name: Hardhat Test
           command: |
             TEST_FILES="$(circleci tests glob "./test/**/*.ts" | circleci tests split)"
-            yarn test ${TEST_FILES}
+            yarn test:parallel ${TEST_FILES}
   lint-and-test:
     docker:
       - image: "cimg/base:stable"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ jobs:
       - run:
           name: Hardhat Test
           command: |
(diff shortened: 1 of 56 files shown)


jakerockland (approved):
This is really really awesome @ryley-o – TYVM 🙏 🙏

PR #340, opened 2022-10-11
Auto-publish docs

ryley-o:
Add workflow to publish docs to github pages on pushes to `main`.

This adds a github workflow to publish the automatically generated docs by hardhat-docgen to this project's github pages.
It also updates the docgen config to ignore some of the files that we don't need docs for, to make things a bit more readable.

## Permissions
⚠️  I don't have permission to publish to our pages from this branch, so not able to confirm if deploy step works. build is confirmed to be working though


ryley-o:
@jakerockland - if you have a different preferred publishing route, feel free to suggest an alternative. I'm not very tied to this exact implementation, it just seemed easiest for now.

jakerockland (approved):

PR #341, opened 2022-10-12
fix pages workflow pks to be valid

ryley-o:
fixes this bug: https://github.com/ArtBlocks/artblocks-contracts/actions/runs/3237852121/jobs/5305398318#step:6:19


jakerockland (approved):
🙏 🙏 🙏

PR #343, opened 2022-10-17
verse - mint with hash

AugustDev:
- mint with hash method
- mint batch with hash method

### DIFF
--- a/contracts/PBAB+Collabs/verse-flex/GenArt721CoreV2_VerseFlex.sol
+++ b/contracts/PBAB+Collabs/verse-flex/GenArt721CoreV2_VerseFlex.sol
@@ -181,8 +181,8 @@ contract GenArt721CoreV2_VerseFlex is ERC721, IGenArt721CoreV2_PBAB {
         address _to,
         uint256 _projectId,
         address _by,
-        bytes32 _hash,
-    ) external returns (uint256 _tokenId) {
+        bytes32 _hash
+    ) public returns (uint256 _tokenId) {
         require(
             isMintWhitelisted[msg.sender],
             "Must mint from whitelisted minter contract."
@@ -228,6 +228,35 @@ contract GenArt721CoreV2_VerseFlex is ERC721, IGenArt721CoreV2_PBAB {
         emit Mint(_to, tokenIdToBe, _projectId);
 
         return tokenIdToBe;
+    }
+
+    /**
+     * @notice Mints a batch of tokens from project `_projectId[i]` and sets the
+     * token's owner to `_to[i]`.
+     * @param _to array of Addresses to be the minted token's owner.
+     * @param _projectId array of Project IDs to mint a tokens on.
+     * @param _by array of Purchasers of minted token.
+     * @param _hash array of Token hashes.
+     * @dev sender must be a whitelisted minter
+     */
+    function mintBatchWithHash(
+        address[] memory _to,
+        uint256[] memory _projectId,
+        address[] memory _by,
+        bytes32[] memory _hash
+    ) external {
+        require(
+            _to.length == _projectId.length,
+            "_projectId array length missmatch."
+        );
+
+        require(_to.length == _by.length, "_by array length missmatch.");
+
+        require(_to.length == _hash.length, "_hash array length missmatch.");
+
+        for (uint256 i; i < _to.length; i++) {
+            mintWithHash(_to[i], _projectId[i], _by[i], _hash[i]);
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


contracts/PBAB+Collabs/verse-flex/GenArt721CoreV2_VerseFlex.sol
```diff
@@ -168,6 +168,97 @@ contract GenArt721CoreV2_VerseFlex is ERC721, IGenArt721CoreV2_PBAB {
         randomizerContract = IRandomizer(_randomizerContract);
     }
 
+    /**
+     * @notice Mints a token from project `_projectId` and sets the
+     * token's owner to `_to`.
```
jakerockland, line 173: Nit: can you please clarify here that this is using the provided hash to associate with the token (which may be centralized), rather than deriving this via on-chain mechanisms?

jakerockland (changes requested):
One documentation nit, and one broader suggestion, and then this PR LGTM for merge (and redeployment re: necessary infra deploy on our end).

I think given how this PR is structured, and the intention to use these mint methods directly in a centralized-platform way (IIUC) I would suggest that we just, with this PR, remove the existing `mint` and `_mintToken` methods entirely, as they provide a bit of a footgun if left around and then accidentally used in the future in a mistaken way (can potentially strip unused `Randomizer` construct as well.

That said, this does add less flexibility for the future, which may not be what you want, so please feel free to disregard this suggestion if you would like – I just think that as-is there is some misuse-due-to-confusion potential.

Another way to address this would be to updated the top-of-file documentation for this adjusted implementation to clarify that this is only intended to be used in one "mode" at a time – either with or without centralized minter, but _never using both modes at the same time_.

An additional clarifying question–is the intention that when this is re-deployed we deploy it _without_ an associated minter contract (with this centralized minting dynamic, I'm not sure that it makes sense to use a minter contract, and think probably doesn't to avoid risk of misuse IIUC the intention here correctly–otherwise I would expect y'all need to implement some sort of custom minter that talks to these new methods)?

contracts/PBAB+Collabs/verse-flex/GenArt721CoreV2_VerseFlex.sol
AugustDev: Thanks for spotting @jakerockland! Fixed the comment.

jakerockland:
> > One documentation nit, and one broader suggestion, and then this PR LGTM for merge (and redeployment re: necessary infra deploy on our end).
> > I think given how this PR is structured, and the intention to use these mint methods directly in a centralized-platform way (IIUC) I would suggest that we just, with this PR, remove the existing `mint` and `_mintToken` methods entirely, as they provide a bit of a footgun if left around and then accidentally used in the future in a mistaken way (can potentially strip unused `Randomizer` construct as well.
> > That said, this does add less flexibility for the future, which may not be what you want, so please feel free to disregard this suggestion if you would like – I just think that as-is there is some misuse-due-to-confusion potential.
> > Another way to address this would be to updated the top-of-file documentation for this adjusted implementation to clarify that this is only intended to be used in one "mode" at a time – either with or without centralized minter, but _never using both modes at the same time_.
> > An additional clarifying question–is the intention that when this is re-deployed we deploy it _without_ an associated minter contract (with this centralized minting dynamic, I'm not sure that it makes sense to use a minter contract, and think probably doesn't to avoid risk of misuse IIUC the intention here correctly–otherwise I would expect y'all need to implement some sort of custom minter that talks to these new methods)?
> 
> Thanks for the review and thoughts! You are completely right that we will handle minting logic by calling smart contract via centralised platform, therefore minter contract is not necessary at this point.
> 
> We would like to use contract exactly as you described - by having 2 modes: standard decentralised AB and centralised contract calls. We therefore would like to keep original `mint` method and remaining on-chain logic as it allows us to do both types of projects in the future without a need of new contract deployments. As you suggested I have added a comment about the modes as `@dev` comment for the contract.

That SGTM @AugustDev – thank you for updating the documentation here.

One additional question, how is the public `verseMinter` intended to be used – IIUC, this doesn't currently have functionality added to it, can this just be stripped out?

AugustDev:
> That SGTM @AugustDev – thank you for updating the documentation here.
> 
> One additional question, how is the public `verseMinter` intended to be used – IIUC, this doesn't currently have functionality added to it, can this just be stripped out?

Hi @jakerockland, I left `verseMinter` accidentally, it is now removed.

jakerockland:
> > That SGTM @AugustDev – thank you for updating the documentation here.
> > One additional question, how is the public `verseMinter` intended to be used – IIUC, this doesn't currently have functionality added to it, can this just be stripped out?
> 
> Hi @jakerockland, I left `verseMinter` accidentally, it is now removed.

SGTM–I'll get this merged and deployed ASAP.

PR #345, opened 2022-10-19
V3 Core Explorations

ryley-o:
Add new V3 Core Explorations as a derivative of V3 Core Flagship contract.

This adds a new V3 Core Explorations contract that is intended to integrate with our existing indexing layer with no updates required (other than including the new contracts in our subgraph config).

The V3 Core Explorations is the same as the V3 Core Flagship, other than the following:
note: see diff here: https://www.diffchecker.com/TyJbRlCJ

- Configuring `artblocksCurationRegistryAddress` is unsupported, and the public function will always return the zero address (since there is no curation of Explorations projects)
  - This was done instead of fully removing the function due to our subgraph indexing layer calling this function when populating our subgraph
- public constants `ART_BLOCKS_ERC721TOKEN_ADDRESS_V0` & `ART_BLOCKS_ERC721TOKEN_ADDRESS_V1` are no longer exposed on the contract
  - These values are not indexed in our subgraph, and were therefore fully removed
- `coreVersion` is updated to "v3.0.1" to differentiate it from flagship V3 core
- `coreType` is unchanged as `GenArt721CoreV3" because subgraph indexing layer may index it as if it is any other V3 core contract

## TODO
The following items are likely to not impact the review process for this PR, but must be completed prior to merging.

- [x] Add tests for full coverage of new Explorations contract
- [x] Add deployment script that deploys new Explorations contract + minter suite to dev environment (for end-to-end testing on Goerli)

---
  - https://app.asana.com/0/0/1203146182606976


ryley-o:
@jakerockland - This contract is ready for your review & thoughts, and I don't expect the remaining items in the TODO section to meaningfully impact your review.

scripts/core-suite-deployments/1_reference_mainnet_flagship_V3ExplorationsCore_deployer.ts
```diff
@@ -0,0 +1,288 @@
...
+ * It is intended to document the deployment process and provide a
+ * reference for the steps required to deploy the V3 core contract suite.
+ * IMPORTANT: This deploys a basic randomizer, which may be changed after
+ * deployment by the configured superAdmin.
+ */
+//////////////////////////////////////////////////////////////////////////////
+// CONFIG BEGINS HERE
+//////////////////////////////////////////////////////////////////////////////
+const tokenName = "Art Blocks Explorations";
+const tokenTicker = "BLOCKS_EXPLORATIONS";
```
jakerockland, line 34: Do you think we need a shorter ticker here?

jakerockland (approved):
Contract changes all LGTM–lets chat offline re: appropriate timing around 🚢 ing this.

scripts/core-suite-deployments/1_reference_mainnet_flagship_V3ExplorationsCore_deployer.ts
jakerockland: "EXPLORE" may be better–let me sanity check with a couple folks on leadership.

ryley-o: lmk and I will update this to whatever is best 👍

ryley-o: as discussed in DM, updated token symbol to "EXPLORE", kept "Art Blocks Explorations" 🎉

PR #351, opened 2022-10-21
Add details for additional contract deployment for Explorations.

jakerockland:
Expands on Explorations minter suite (set price addition).


ryley-o (approved):
LGTM!

PR #353, opened 2022-10-24
Add Solidity interfaces for methods used by the Minting API

shantanubala:
## Summary of changes

The Minting API for Art Blocks Engine partners makes use of several function calls that are expected on the core and minter contracts for projects minted using the API.

This PR defines Solidity interfaces that document the functions used by the Minting API.


contracts/interfaces/0.8.x/IMintingAPICoreV0.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+interface IEngineContractMintingAPIV0 {
+    /**
+     * @notice Token ID `_tokenId` minted on project ID `_projectId` to `_to`.
+     */
+    event Mint(
```
jakerockland, line 10: Just wanted to confirm explicitly, the minting API uses the `Mint` event? 🙏

contracts/interfaces/0.8.x/IMintingAPICoreV0.sol
```diff
@@ -0,0 +1,54 @@
...
+            address,
+            uint256,
+            uint256,
+            uint256,
+            bool,
+            address,
+            uint256,
+            string memory,
+            address
+        );
```
jakerockland, line 48: Makes sense that these are dependencies right now, but flagging as items that will need to be updated on minting API side when we upgrade Engine deployments to use V3 spec. cc @ryley-o

contracts/interfaces/0.8.x/IMintingAPIMinterV0.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+interface IEngineContractMintingAPIV0 {
```
jakerockland, line 6: Can we name this `IMintingAPIMinterV0` please?

contracts/interfaces/0.8.x/IMintingAPICoreV0.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+interface IEngineContractMintingAPIV0 {
```
jakerockland, line 6: Can we name this `IMintingAPICoreV0` please?

jakerockland:
@shantanubala this PR generally LGTM–added some various minor comments inline.

contracts/interfaces/0.8.x/IMintingAPICoreV0.sol
jakerockland, line 6: Could we add a comment block above the interface definition here that clarifies why this interface is needed (e.g. for compatibility verification with the minting API assumptions)?

contracts/interfaces/0.8.x/IMintingAPIMinterV0.sol
jakerockland, line 6: Could we add a comment block above the interface definition here that clarifies why this interface is needed (e.g. for compatibility verification with the minting API assumptions)?

jakerockland:
Adding @ryley-o here for additional visibility/input.

contracts/interfaces/0.8.x/IMintingAPICoreV0.sol
shantanubala: Yup, it does 👍

contracts/interfaces/0.8.x/IMintingAPICoreV0.sol
shantanubala: Done 👍

contracts/interfaces/0.8.x/IMintingAPIMinterV0.sol
shantanubala: Done 👍

contracts/interfaces/0.8.x/IMintingAPICoreV0.sol
shantanubala: Taking a look through the V3 spec right now 👍

contracts/interfaces/0.8.x/IMintingAPIMinterV0.sol
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * In order for a minter contract to be compatible with the REST API provided
+ * at https://minting-api.artblocks.io/ it must implement the interface defined
+ * below.
+ */
+interface IMintingAPIMinterV0 {
+    mapping(uint256 => bool) public projectMaxHasBeenInvoked;
+
+    function purchaseTo(address _to, uint256 _projectId)
+        private
```
ryley-o, line 16:
```suggestion
        external
```

contracts/interfaces/0.8.x/IMintingAPICoreV0.sol
```diff
@@ -0,0 +1,60 @@
...
+        address indexed _to,
+        uint256 indexed _tokenId,
+        uint256 indexed _projectId
+    );
+
+    // getter function of public variable
+    function nextProjectId() external view returns (uint256);
+
+    function projectDetails(uint256 _projectId)
+        public
```
ryley-o, line 26:
```suggestion
        external
```

contracts/interfaces/0.8.x/IMintingAPIMinterV0.sol
```diff
@@ -0,0 +1,28 @@
...
+pragma solidity ^0.8.0;
+
+/* Provides an interface definition for compatibility with the Minting API.
+ *
+ * In order for a minter contract to be compatible with the REST API provided
+ * at https://minting-api.artblocks.io/ it must implement the interface defined
+ * below.
+ */
+interface IMintingAPIMinterV0 {
+    mapping(uint256 => bool) public projectMaxHasBeenInvoked;
```
ryley-o, line 13: mappings aren't allowed in an interface, but a public mapping actually creates the following public getter function. We label it as external here since interfaces must have all functions defined external (but contracts can define a public function and still meet the requirement of the interface)
```suggestion
    function projectMaxHasBeenInvoked(uint256) external returns (bool);
```

ryley-o (approved):
Thanks @shantanubala!
This generally LGTM, with the understanding that this only works with V1 or V2 engine contracts, and only the subset of those contracts that implement ERC721Enumerable (which enables the `tokensOfOwner` function). For example, GenArt721CoreV2_ArtBlocksXPace does not implement that function. for that reason, I would recommend that we don't edit files such as `IGenArt721CoreContractV1.sol` to say `interface IGenArt721CoreContractV1 is IMintingAPICoreV0` (since we will get compile errors on the non-enumerable contracts)

IIUC, we will need to define new interfaces that V3 cores must conform to, since many of these functions have been changed or are no longer possible in V3 🙏 We will likely need to update some of the minting API to use our subgraph for token enumeration, and there are new view functions with a bit more intuitive, but different, organization.

Regarding the CI issue, not sure why we CircleCI is not running these commits. @jakerockland , any idea why CirclieCI is bugging out? I've never seen this issue before (appears to be different than a fork that needs to run, which I have seen before...)

PR #354, opened 2022-10-24
Update Merkle Minter V2 w/additional vault documentation

ryley-o:
Update Merkle Minter V2 with comments surrounding vault delegation details.

Note that I also updated a constant to be UPPERCASE, which may have frontend implications/impacts. That specific change is definitely optional, but I would like to follow solidity's style guide if possible: https://docs.soliditylang.org/en/develop/style-guide.html#constants Cc @mchrupcala if this is impacting frontend work at all.

I think most of these comments may be copy-pasted to the token holder V2, once it is ready.


jakerockland (approved):
Thank you very much ser 🙏

PR #357, opened 2022-10-24
V2 Token Holder Contract + Delegation Registry

mchrupcala:
Vault delegation is now supported in this V2 upgrade for the token-holder minter -- this PR includes unit tests and deployment updates to support that change.

Diff between V1 and V2 of the .sol contracts can be found here: https://www.diffchecker.com/BElKj7EX


contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,765 @@
...
+ * ----------------------------------------------------------------------------
+ * The following functions are restricted to a project's artist:
+ * - allowHoldersOfProjects
+ * - removeHoldersOfProjects
+ * - allowRemoveHoldersOfProjects
+ * - updatePricePerTokenInWei
+ * ----------------------------------------------------------------------------
+ * Additional admin and artist privileged roles may be described on other
+ * contracts that this minter integrates with.
+ */
```
jakerockland, line 43: Let's expand on the file comment with something about the vault-delegate functionality, similar to what we have here in the MerkleMinter: https://github.com/ArtBlocks/artblocks-contracts/blob/main/contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV2.sol#L36-L48

🙏

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,765 @@
...
+contract MinterHolderV2 is ReentrancyGuard, IFilteredMinterHolderV0 {
+    /// Delegation registry address
+    address public constant delegationRegistryAddress =
```
jakerockland, line 46: Let's use the ALL_CAPS convention, as added recently to the MerkleMinter in a quick follow up from @ryley-o in #354 

https://github.com/ArtBlocks/artblocks-contracts/blob/main/contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV2.sol#L53-L55

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,765 @@
...
+    /**
+     * @notice Purchases a token from project `_projectId` and sets
+     *         the token's owner to `_to`, as a delegate, (the `msg.sender`)
+     *         on behalf of an explicitly defined vault.
+     * @param _to Address to be the new token's owner.
+     * @param _projectId Project ID to mint a token on.
+     * @param _ownedNFTAddress ERC-721 NFT address owned by msg.sender being used to
+     * claim right to purchase.
+     * @param _ownedNFTTokenId ERC-721 NFT token ID owned by msg.sender being used
+     * to claim right to purchase.
```
jakerockland, line 543: In this case, the NFT token ID is the one owned by the `_vault` being provided, rather than `msg.sender`.

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,765 @@
...
+    /**
+     * @notice gas-optimized version of purchaseTo(address,uint256,address,uint256,address).
+     */
+    function purchaseTo_L69(
+        address _to,
+        uint256 _projectId,
+        address _ownedNFTAddress,
+        uint256 _ownedNFTTokenId,
+        address _vault // acceptable to be `address(0)` if no vault
+    ) public payable nonReentrant returns (uint256 tokenId) {
```
jakerockland, line 573: From checking against https://emn178.github.io/solidity-optimize-name/ I think we want to update our gas-optimized name from the previous `purchaseTo_L69` to be now `purchaseTo_dlc`^.

^You can verify this yourself by passing `purchaseTo(address,uint256,address,uint256,address)` into that tool–tool is a bit poorly documented in terms of what its expected input format is.

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,765 @@
...
+import "@openzeppelin-4.5/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin-4.5/contracts/security/ReentrancyGuard.sol";
+import "@openzeppelin-4.5/contracts/utils/structs/EnumerableSet.sol";
+import "@openzeppelin-4.5/contracts/utils/structs/EnumerableMap.sol";
+
+pragma solidity 0.8.17;
+
+/**
+ * @title Filtered Minter contract that allows tokens to be minted with ETH
+ * when purchaser owns an allowlisted ERC-721 NFT.
```
jakerockland, line 19: Realizing that we should update the documentation here to more explicitly highlight that this minter does not "keep track" of if a given minter has already minted or not, it just allows one to use another ERC721 as a gate for minting without attempting to enforce a "snapshot" of ERC721 ownership (which is likely better served by an allowlist minter in most use cases).

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,765 @@
...
+        // that there is not a true vault, or may be `_vault` if one is
+        // provided explicitly (and it is valid).
+        address vault = msg.sender;
+        if (_vault != address(0)) {
+            // If a vault is provided, it must be valid, otherwise throw rather
+            // than optimistically-minting with original `msg.sender`.
+            // Note, we do not check `checkDelegateForAll` as well, as it is known
+            // to be implicitly checked by calling `checkDelegateForContract`.
+            bool isValidVault = delegationRegistryContract
+                .checkDelegateForContract(
```
jakerockland, line 617: I think we want the more specific `checkDelegateForToken` check in this case – and then should update the documentation comment above clarifying that this will be known to implicitly check both `checkDelegateForContract` and `checkDelegateForAll` (cc @ryley-o for an additional sanity check on this).

jakerockland (commented):
This generally LGTM, awesome work @mchrupcala 🙌 🔥 🙌 🔥 

Handful of pretty minor requests flagged inline 🙏 🙏 

Please re-request review when ready for another pass on my end 💜

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
ryley-o: yes ser - `checkDelegateForToken` would be the appropriate call in this case. good catch!

jakerockland: !vibe

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
ryley-o: The 'ole Merkle minter is a pretty good one 😄 agree that over-explaining here is a good idea

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
ryley-o: ++ to the above, and would like to also have natspec comments on this public function (can re-use one from `purchaseTo`)

ryley-o:
@mchrupcala if you don't mind, I'll prefer to wait until @jakerockland 's comments are resolved, then I'll give my review on this one 🙏 🚀

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
mchrupcala: Sounds good! I added these two lines...what do ya'll think?

> * @title Filtered Minter contract that allows tokens to be minted with ETH
 * when purchaser owns an allowlisted ERC-721 NFT. This contract does NOT track
 * if a purchaser has/has not minted already -- it simply restricts purchasing
 * to anybody that holds one or more of a specified list of ERC-721 NFTs.

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
mchrupcala: Done and done! I updated the language to reflect that this contract only allows delegating at the token level and not the contract.

jakerockland: So contract-level delegations are still possible and supported by this minter–they get inherited when we perform a token-specific lookup check. I think we may want to adjust wording here slightly and perhaps reword the language a bit in https://github.com/ArtBlocks/artblocks-contracts/blob/main/contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV2.sol#L36-L48 as well now that I'm looking at it again, as I realizing that it can be misunderstood; cc @ryley-o

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
jakerockland: That is perfect @mchrupcala 🙏

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
ryley-o: @jakerockland gaaaa yeah idk what I was thinking when writing those lines in the Merkle Minter.
Merkle V2 should say contract-level or wallet-level delegations are alllowed, and this minter should say token-level, contract-level, or wallet-level delegations are allowed.

mchrupcala: I made a few more tweaks just to be really clear about what our minters do/don't handle as far as the delegation goes. But please feel free to tear these up if I'm offbase with any of my wording 😅 
https://github.com/ArtBlocks/artblocks-contracts/pull/357/commits/a01d69d00f0be4b837c6a2218db853af2a254a46

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,791 @@
...
+                _vault
+            );
+    }
+
+    /**
+     * @notice gas-optimized version of purchaseTo(address,uint256,address,uint256,address).
+     * @param _to Address to be the new token's owner.
+     * @param _projectId Project ID to mint a token on.
+     * @param _ownedNFTAddress ERC-721 NFT address owned by msg.sender being used to
+     *         claim right to purchase.
```
ryley-o, line 586: nit: For this function, the NFT is either owned by _vault if _vault is not zero address, otherwise owned by msg.sender.

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,791 @@
...
+    }
+
+    /**
+     * @notice Purchases a token from project `_projectId` and sets
+     *         the token's owner to `_to`, as a delegate, (the `msg.sender`)
+     *         on behalf of an explicitly defined vault.
+     * @param _to Address to be the new token's owner.
+     * @param _projectId Project ID to mint a token on.
+     * @param _ownedNFTAddress ERC-721 NFT address owned by msg.sender being used to
+     *         claim right to purchase.
```
ryley-o, line 558: nit: For this function, the NFT is either owned by _vault if _vault is not zero address, otherwise owned by msg.sender.

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,791 @@
...
+
+    /**
+     * @notice gas-optimized version of purchaseTo(address,uint256,address,uint256,address).
+     * @param _to Address to be the new token's owner.
+     * @param _projectId Project ID to mint a token on.
+     * @param _ownedNFTAddress ERC-721 NFT address owned by msg.sender being used to
+     *         claim right to purchase.
+     * @param _ownedNFTTokenId ERC-721 NFT token ID owned by _vault being used
+     *         to claim right to purchase.
+     * @param _vault Vault being purchased on behalf of.
```
ryley-o, line 589: nit: we should include "acceptable to be `address(0)` if no vault"

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,791 @@
...
+     *         the token's owner to `_to`, as a delegate, (the `msg.sender`)
+     *         on behalf of an explicitly defined vault.
+     * @param _to Address to be the new token's owner.
+     * @param _projectId Project ID to mint a token on.
+     * @param _ownedNFTAddress ERC-721 NFT address owned by msg.sender being used to
+     *         claim right to purchase.
+     * @param _ownedNFTTokenId ERC-721 NFT token ID owned by _vault being used
+     *         to claim right to purchase.
+     * @param _vault Vault being purchased on behalf of.
+     * @return tokenId Token ID of minted token
```
ryley-o, line 562: nit: we should include "acceptable to be `address(0)` if no vault"

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,791 @@
...
+     */
+    function purchaseTo_dlc(
+        address _to,
+        uint256 _projectId,
+        address _ownedNFTAddress,
+        uint256 _ownedNFTTokenId,
+        address _vault // acceptable to be `address(0)` if no vault
```
ryley-o, line 597: nit: we should probably just move this detail to the natspec param above, since it will then be included in our auto-generated docs.

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,791 @@
...
+                    _ownedNFTTokenId // tokenId
+                );
+            require(isValidVault, "Invalid delegate-vault pairing");
+            vault = _vault;
+        }
+
+        // EFFECTS
+        tokenId = minterFilter.mint(_to, _projectId, vault);
+
+        // NOTE: delegate-vault handling **ends here**.
```
ryley-o, line 655: should this comment be above the // EFFECTS line?

scripts/core-suite-deployments/1_reference_mainnet_flagship_V3Core_deployer.ts
```diff
@@ -13,7 +13,7 @@ import { MinterSetPriceERC20V2__factory } from "../contracts/factories/MinterSet
 import { MinterDALinV2__factory } from "../contracts/factories/MinterDALinV2__factory";
```
ryley-o, line 13: FWIW, I don't think you have to go back and update all of these reference files. might be nice to add a single deployer script for adding this minter to existing core deployements (see `./scripts/minter-deployments/`), but most of these scripts are kept for archival purposes/starting points for future deployments. We should formalize what our entire script retention policy is too, just haven't prioritized that 👍

test/minter-suite-minters/MinterHolder/MinterHolderV2.test.ts
```diff
@@ -0,0 +1,476 @@
...
+  "GenArt721CoreV3_Explorations", // V3 core explorations contract
+];
+
+/**
+ * These tests intended to ensure Filtered Minter integrates properly with V3
+ * core contract.
+ */
+for (const coreContractName of coreContractsToTest) {
+  describe(`MinterHolderV2_${coreContractName}`, async function () {
+    let fakeDelegationRegistry: FakeContract<IDelegationRegistry>;
```
ryley-o, line 30: Similar to a previous comment, I think we should update this minter's architecture and tests to be in alignment with #370

ryley-o (commented):
🙏 Thank you ser!

At the moment, my comments are limited to minor nits, plus a general preference to change this minter's architecture to be more aligned with what we did in #370 (assuming we all end up liking the suggested patterns in that PR 👍). Admittedly, switching to that PR's architecture is difficult to do prior to merging that PR into main, so I would recommend either treating that as a blocker for this PR, or re-basing this PR off of that branch (with the risk of adding re-work if we end up not liking that architecture).

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
mchrupcala: Ah I moved this line here because the vault's referenced in `mint()` on line 653. But I was wondering the same thing since we're not doing any 'handling' on line 653. What's your preference?

ryley-o: hmm, yeah, since vault is referenced on 653, I see what you mean. I'll resolve this and let's just leave as-is 😄

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,791 @@
...
+ * users should take care to ensure that the delegation registry is secure.
+ * Delegations must be configured by the vault owner prior to purchase. Supported
+ * delegation types include token-level, contract-level (via genArt721CoreAddress), or
+ * wallet-level delegation. Contract-level delegations must be configured for the core
+ * token contract as returned by the public immutable variable `genArt721CoreAddress`.
+ */
+contract MinterHolderV2 is ReentrancyGuard, IFilteredMinterHolderV0 {
+    /// Delegation registry address
+    address public constant DELEGATION_REGISTRY_ADDRESS =
+        0x00000000000076A84feF008CDAbe6409d2FE638B;
```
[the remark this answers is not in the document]
mchrupcala: Agreed! This minter's been refactored and I added the generic event as well 👍🏻

scripts/core-suite-deployments/1_reference_mainnet_flagship_V3Core_deployer.ts
mchrupcala: Sounds good, ok I went ahead and added the deployment scripts for dev & mainnet

jakerockland:
Updated diff that I'm re-reviewing against based on latest contract commit: https://www.diffchecker.com/jutwUHj0

ryley-o (approved):
🙏 This is awesome!

Couple of nits is all 🎉

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
```diff
@@ -0,0 +1,795 @@
...
+     * project IDs `_ownedNFTProjectIds` to mint on project `_projectId`.
+     * `_ownedNFTAddresses` assumed to be aligned with `_ownedNFTProjectIds`.
+     * e.g. Removes holders of project `_ownedNFTProjectIds[0]` on token
+     * contract `_ownedNFTAddresses[0]` from mint allowlist.
+     */
+    event RemovedHoldersOfProjects(
+        uint256 indexed _projectId,
+        address[] _ownedNFTAddresses,
+        uint256[] _ownedNFTProjectIds
+    );
```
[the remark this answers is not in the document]
mchrupcala: Makes sense! Yeah sounds like it's the perfect time to make this change.

With these moved, the ordering of the contract's registry assignments will look like this (after the other nits to re-arrange these after the events):
```
contract MinterHolderV2 is ReentrancyGuard, IFilteredMinterHolderV1 {
    // add Enumerable Set methods
    using EnumerableSet for EnumerableSet.AddressSet;

    /// Delegation registry address
    address public immutable delegationRegistryAddress;

    /// Delegation registry address
    IDelegationRegistry private immutable delegationRegistryContract;
```

jakerockland (approved):
LGTM as well–awesome work @mchrupcala

contracts/minter-suite/Minters/MinterHolder/MinterHolderV2.sol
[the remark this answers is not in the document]
ryley-o: this is perfect! that `using x for y` is great to have at the very top of the contract IMO. I can't find anywhere in the solidity style guide stating where it should be, but I much prefer it at the top since it defines the library functions available across the entire contract 😄

PR #356, opened 2022-10-24
Deploy scripts for MinterMerkleV2, mainnet and dev

ryley-o:
Add deploy scripts for deployment of MinterMerkleV2 for dev and mainnet environments.

Note that dev deployments have already been executed, and are available here: 
- (dev flagship) https://goerli.etherscan.io/address/0x126eC6A5Dd59f9414c778c7433E6C9100c4936fB#code
- (dev explorations) https://goerli.etherscan.io/address/0x723D5203689bA5A249Ea97630A5bE3FDd6B69cd7#code

@jakerockland - mainnet deployments should be g2g with your deployer wallet, when you get the chance 👍 The script deploys a MinterMerkleV2 minter for both the flagship and the explorations contracts.

I've also stopped contract sizer from polluting the terminal on every compile.

PR #360, opened 2022-10-25
Update hardhat config to use auto-gas and remove unused testnets

jakerockland:
Minor config cleanup while performing deployments for #356


ryley-o (approved):
LGTM!

PR #361, opened 2022-10-25
Update Art Blocks Engine/Collaborations spec (V3)

jakerockland:
This PR updates the Art Blocks Engine template to make use of the benefits of V3, intended to be used across both standalone Engine deployments and Collaborations Engine deployments (previously described as PRTNR contracts in old contract nomenclature before V3).

Shared V3 test coverage has also been added/adjusted accordingly to include shared support for this new updated `_Engine` variant template, and isolated `V3_Engine`-specific view tests have been added in `GenArt721CoreV3_Views_Engine.test.ts` via https://github.com/ArtBlocks/artblocks-contracts/pull/361/commits/73f958aebe6376a6144ea3c0dab9bfcc35395203 

---

Diff-log between `GenArt721CoreV3_Engine` and `GenArt721CoreV3`: https://www.diffchecker.com/nJTzctlU

Diff-log between `IGenArt721CoreContractV3` and `IGenArt721CoreContractV3_Engine`: https://www.diffchecker.com/0gf9LMVE

---
  - https://app.asana.com/0/0/1203222636168458


jakerockland:
Definitely not ready for review yet, but up on Github for visibility / discussion (and so that I don't loose it again if I have another computer get rekd... 😅 ).

jakerockland:
@ryley-o This is definitely not in a merge-ready state (most importantly haven't yet added tests) but I do think that it is in a review ready state from the perspective of asking for you to TAL over the set of diffs captured above (and the changelog) and making sure that we are directionally on the same page about the above before getting deeper into test coverage.

LMK if this plan sounds good to you, and if so if you could do an initial review as an initial a sign off before I start getting into writing tests, that would be super appreciated.

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+        _updatePlatformProviderPrimarySalesAddress(_platformProviderAddress);
+        _updatePlatformProviderSecondarySalesAddress(_platformProviderAddress);
+        _updateRandomizerAddress(_randomizerContract);
+        // set AdminACL management contract as owner
+        _transferOwnership(_adminACLContract);
+        // initialize default base URI
+        _updateDefaultBaseURI(
+            string.concat(
+                "https://token.artblocks.io/",
+                address(this).toHexString(),
```
jakerockland, line 399: A bit initially confused why this is failing from the error message alone–any thoughts here @ryley-o https://app.circleci.com/pipelines/github/ArtBlocks/artblocks-contracts/1493/workflows/6fb418cb-6a86-4ab1-8584-2790674439dd/jobs/1738?invite=true#step-110-16

ryley-o:
>LMK if this plan sounds good to you, and if so if you could do an initial review as an initial a sign off before I start getting into writing tests, that would be super appreciated.

Will take a look now!

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+            // renderProviderRevenue_ percentage is always <=100, so guaranteed to never underflow
+            projectFunds = _price - renderProviderRevenue_;
+        }
+        platformProviderRevenue_ =
+            (_price * uint256(_platformProviderPrimarySalesPercentage)) /
+            ONE_HUNDRED;
+        unchecked {
+            // platformProviderRevenue_ percentage is always <=100, so guaranteed to never underflow
+            projectFunds = _price - platformProviderRevenue_;
+        }
```
ryley-o, line 1862: I think would be good to add come comments here, and potentially move this outside of an unchecked block. This is now a three-way split of a pie, any loss of precision caused by a very low `_price` variable is now less straightforward to understand. In the case of a low price, Solidity should be rounding down when losing precision (we should confirm that in tests by testing that the artist receives all revenues when a sale price of 1 is input into this function), meaning that this should still never underflow. All things considered, I would lean towards keeping this out of an unchecked block, or writing enough tests here to ensure we feel very confident that Solidity's rounding down is working as intended. We should also add some inline comments that indicate why rounding down will prevent underflow here.

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+        string memory _tokenSymbol,
+        address _renderProviderAddress,
+        address _platformProviderAddress,
+        address _randomizerContract,
+        address _adminACLContract,
+        uint248 _startingProjectId,
+        bool _autoApproveArtistSplitProposals
+    )
+        ERC721_PackedHashSeed(_tokenName, _tokenSymbol)
+        onlyNonZeroAddress(_randomizerContract)
```
ryley-o, line 381: since we now specify `_renderProviderAddress` & `_platformProviderAddress` as input addresses, we should add a `onlyNonZeroAddress()` modifier for both of them.

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+    /// to accept artist address change proposals, or if these proposals
+    /// should always auto-approve, as determined by the business process
+    /// requirements of the Engine partner using this contract.
+    ///
+    /// ^does not apply in the case where contract-ownership itself is revoked
+    bool public immutable autoApproveArtistSplitProposals;
+
+    /// version & type of this core contract
+    string public constant coreVersion = "v3.0.0";
+    string public constant coreType = "GenArt721CoreV3";
```
ryley-o, line 275: Because this conforms to a new interface that indexing will have to adjust to, I think we should update this to be the following:
```suggestion
    string public constant coreVersion = "v3.1.0";
    string public constant coreType = "GenArt721CoreV3_Engine";
```

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+ * - updateProjectScriptType
+ * - updateProjectAspectRatio
+ * ----------------------------------------------------------------------------
+ * The following functions are restricted to only the Artist address:
+ * - proposeArtistPaymentAddressesAndSplits (Note that this has to be accepted
+ *   by adminAcceptArtistAddressesAndSplits to take effect, which is restricted
+ *   to the Admin ACL contract, or the artist if the core contract owner has
+ *   renounced ownership. Also note that a proposal will be automatically
+ *   accepted if the artist only proposes changed payee percentages without
+ *   modifying any payee addresses, or is only removing payee addresses.)
```
ryley-o, line 65: This should be updated for the added `autoApproveArtistSplitProposals` logic

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+    using Bytes32Strings for bytes32;
+    using Strings for uint256;
+    uint256 constant ONE_HUNDRED = 100;
+    uint256 constant ONE_MILLION = 1_000_000;
+    uint24 constant ONE_MILLION_UINT24 = 1_000_000;
+    uint256 constant FOUR_WEEKS_IN_SECONDS = 2_419_200;
+    uint8 constant AT_CHARACTER_CODE = uint8(bytes1("@")); // 0x40
+
+    // numeric constants
+    uint256 constant MAX_SECONDARY_SALES_BPS = 10000; // 10_000 BPS = 100%
```
ryley-o, line 112: WDYT about naming this `MAX_PROVIDER_SECONDARY_SALES_BPS` (since it is the max of sum render provider and platform provider secondary sales bps values)

ryley-o (commented):
This is really great, ty @jakerockland!

Adding this as a comment review since this is an initial pass prior to test coverage 🙏

contracts/GenArt721CoreV3_Engine.sol
jakerockland: Aha! Awesome–thank you ser.

contracts/interfaces/0.8.x/IGenArt721CoreContractV3_Engine.sol
```diff
@@ -0,0 +1,181 @@
...
+    // @dev The render provider secondary sales royalties payment address
+    function platformProviderSecondarySalesAddress()
+        external
+        view
+        returns (address payable);
+
+    // @dev Basis points of secondary sales allocated to the render provider
+    function renderProviderSecondarySalesBPS() external view returns (uint256);
+
+    // @dev Basis points of secondary sales allocated to the render provider
```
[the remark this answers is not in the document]
jakerockland: Ah whoops–thank you!!

contracts/interfaces/0.8.x/IGenArt721CoreContractV3_Engine.sol
```diff
@@ -0,0 +1,181 @@
...
+        view
+        returns (uint256);
+
+    // @dev The render provider secondary sales royalties payment address
+    function renderProviderSecondarySalesAddress()
+        external
+        view
+        returns (address payable);
+
+    // @dev The render provider secondary sales royalties payment address
```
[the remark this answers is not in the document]
jakerockland: Ah whoops–thank you!!

contracts/interfaces/0.8.x/IGenArt721CoreContractV3_Engine.sol
```diff
@@ -0,0 +1,181 @@
...
+        view
+        returns (address payable);
+
+    // @dev Percentage of primary sales allocated to the render provider
+    function renderProviderPrimarySalesPercentage()
+        external
+        view
+        returns (uint256);
+
+    // @dev Percentage of primary sales allocated to the render provider
```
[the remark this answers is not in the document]
jakerockland: Thank you!!

contracts/interfaces/0.8.x/IGenArt721CoreContractV3_Engine.sol
```diff
@@ -0,0 +1,181 @@
...
+            bool locked
+        );
+
+    // @dev The render provider primary sales payment address
+    function renderProviderPrimarySalesAddress()
+        external
+        view
+        returns (address payable);
+
+    // @dev The render provider primary sales payment address
```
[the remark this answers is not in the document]
jakerockland: Thank you!!

contracts/GenArt721CoreV3_Engine.sol
jakerockland: Great point–applied.

contracts/GenArt721CoreV3_Engine.sol
jakerockland: I dig it ser–updating!

contracts/GenArt721CoreV3_Engine.sol
jakerockland: Yeah great point, realized we should also have for `_adminACLContract` – updated accordingly (leaving this unresolved to make sure you see the note re: `_adminACLContract`).

contracts/GenArt721CoreV3_Engine.sol
jakerockland: Great catch–will do so (leaving this comment unresolved until I do so).

contracts/GenArt721CoreV3_Engine.sol
jakerockland: Great point; agree with your rationale here–will update to do so (leaving this comment unresolved until I do so).

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity 0.8.17;
+
+// Created By: Art Blocks Inc.
+
+import "./interfaces/0.8.x/IRandomizerV2.sol";
+import "./interfaces/0.8.x/IAdminACLV0.sol";
```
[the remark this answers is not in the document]
jakerockland: My thinking was that given the current ACL implementation is super simple (designed mainly to support the superuser use case only) we may as well add in support from this from the get-go as it will not functionally impact partners in how they interact with their contracts by default (we'll still ask for one initial wallet, multi-sig or cold wallet upon setup) but it does enable more flexibility/upgrade-ability to support partners' own business practices/contract management practices as needed, for those wanting to support more flexible behavior down the line. WDYT ser?

contracts/GenArt721CoreV3_Engine.sol
jakerockland: Addressed here and below.

contracts/GenArt721CoreV3_Engine.sol
jakerockland: Updated to remove unchecked logic and to expand on documentation a bit–can continue to iterate on this as needed.

contracts/GenArt721CoreV3_Engine.sol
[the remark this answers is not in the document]
ryley-o: Yeah, I think your logic there all makes sense 👍
One last thing I recently realized - did how we end up having `owner` return the AdminACL contract itself end up preventing automatic integration of contracts on OpenSea, etc.? For example, if we had the core contract's `owner` function return the adminACL's `superAdmin`, would integration with third party sites go smoother? (this is something I realized yesterday...) Too late for our flagship contract, but maybe something we could think of for the Engine contract.

contracts/V3_ENGINE_CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
...
+
+## The following changes were made in the Core V3 Engine contract:
+
+- Removes reference to "curation registry" concept
+- Removes on-chain reference to previous flagship core contracts
+- Changes "artblocks" payee to be split into a "renderProvider" and "platformProvider" set of payees
+- Removes "backwards compatible" oriented financial view-only methods
+- Updates royalty-limit logic to account for two providers – "render" and "platform" providers.
+- Added support for a global "auto approve" proposals setting `autoApproveArtistSplitProposals`, that is determined at time of contract deployment (this likely will default to `true` based on most Engine partners' onboarding processes currently)
+- TODO: integrates with the Engine registration-beacon event emitter (https://github.com/ArtBlocks/artblocks-contracts/issues/358)
```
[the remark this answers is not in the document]
jakerockland: I think in this case, we'll just want to start deploying the Engine setup with a minter filter + accompanying single minter.

contracts/GenArt721CoreV3_Engine.sol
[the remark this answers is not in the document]
jakerockland: This is a great point ser – updated with https://github.com/ArtBlocks/artblocks-contracts/pull/361/commits/d638a55b09ffa412eb5cf64859e37065fd8efe70

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+        // Validate that the sum of a) the current *platform* provider %; and
+        // b) the proposed *render* provider %, does not exceed 100%.
+        require(
+            (renderProviderPrimarySalesPercentage_ + /* proposed */
+                _platformProviderPrimarySalesPercentage) <= ONE_HUNDRED, /* current */
+            "Max sum of ONE_HUNDRED %"
+        );
+        _renderProviderPrimarySalesPercentage = uint8(
+            renderProviderPrimarySalesPercentage_
+        );
```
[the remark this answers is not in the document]
jakerockland: Updating! TY ser.

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+        // Validate that the sum of a) the current *render* provider %; and
+        // b) the proposed *platform* provider %, does not exceed 100%.
+        require(
+            (platformProviderPrimarySalesPercentage_ + /* proposed */
+                _renderProviderPrimarySalesPercentage) <= ONE_HUNDRED, /* current */
+            "Max sum of ONE_HUNDRED %"
+        );
+        _platformProviderPrimarySalesPercentage = uint8(
+            platformProviderPrimarySalesPercentage_
+        );
```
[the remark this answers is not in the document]
jakerockland: Updating–TY ser!

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+     */
+    function getRoyalties(uint256 _tokenId)
+        external
+        view
+        onlyValidTokenId(_tokenId)
+        returns (address payable[] memory recipients, uint256[] memory bps)
+    {
+        // initialize arrays with maximum potential length
+        recipients = new address payable[](3);
+        bps = new uint256[](3);
```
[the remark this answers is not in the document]
jakerockland: 🤦 good catch–fixed! :)

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+        emit ProjectUpdated(_projectId, FIELD_PROJECT_SCRIPT);
+    }
+
+    /**
+     * @notice Updates script for project `_projectId` at script ID `_scriptId`.
+     * @param _projectId Project to be updated.
+     * @param _scriptId Script ID to be updated.
+     * @param _script The updated script value.
+     */
+    function updateProjectScript(
```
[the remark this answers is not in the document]
jakerockland: Good call out–updating!

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2137 @@
...
+        // the contract that no longer exists with the newly created one
+        project.scriptBytecodeAddresses[_scriptId] = _script.writeToBytecode();
+        emit ProjectUpdated(_projectId, FIELD_PROJECT_SCRIPT);
+    }
+
+    /**
+     * @notice Removes last script from project `_projectId`.
+     * @param _projectId Project to be updated.
+     */
+    function removeProjectLastScript(uint256 _projectId)
```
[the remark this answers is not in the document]
jakerockland: Good call out–updating!

contracts/EngineRegistryV0.sol
```diff
@@ -0,0 +1,71 @@
...
+        registeredContractAddresses[_contractAddress] = true;
+    }
+
+    /**
+     * @inheritdoc IEngineRegistryV0
+     */
+    function unregisterContract(address _contractAddress) external {
+        // CHECKS
+        // Validate against `tx.origin` rather than `msg.sender` for consistency with the above approach,
+        // as we expect in usage of this contract `msg.sender == tx.origin` to be a true assessment.
```
lyaunzbe, line 57: super super minor nit, but I feel like the `msg.sender == tx.origin` is an important assumption in this context and  makes sense to highlight just in the above check comments (in registerContract), since you're referencing them here anyway

jakerockland: Ah – well in the case of the above callsite, this actually isn't true that we expect `msg.sender == tx.origin`, we actually only expect it typically in the `unregisterContract` path. In the `registerContract` path, we expect `msg.sender != tx.origin`, and to intentionally need to validate against `tx.origin.

lyaunzbe: ah got it, appreciate the clarification here.

more so for my own understanding, is this because it's more likely that a registration call is potentially coming from a contract vs EOA, while we expect that not to be the case for unregisterContract calls (and expect those to primarily come from an EOA)? if so, why is that?

lyaunzbe: sorry bit of a brain short circuit from my end... registration calls are pretty much solely coming from the constructor of the contracts being deployed by that deployer address, whereas unregistering would be a manual process so this all makes sense 👍

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -408,6 +410,12 @@ contract GenArt721CoreV3_Engine is
         // initialize next project ID
         _nextProjectId = _startingProjectId;
         emit PlatformUpdated(FIELD_NEXT_PROJECT_ID);
+        // register contract as an Engine contract
+        IEngineRegistryV0(_engineRegistryContract).registerContract(
```
yoshiwarab, line 414: I think no one would fork the Engine contract to use with their own infra (would imagine they would just use the normal core contract) but if we do imagine that might be a possibility should we handle the case that the registry contract doesn't exist?

yoshiwarab (approved):

contracts/EngineRegistryV0.sol
jakerockland: Yep yep–not a worry ser; your EDIT response is spot on: registrations are expected to happen via contract-being-deployed auto-registering in the constructor, vs. unregistering likely being something that is manual (though we don't have current plans/expected cases of unregistrations), but even if it were automated it would need to be from a `tx.origin` of the allowed deployer.

contracts/GenArt721CoreV3_Engine.sol
jakerockland: @yoshiwarab in that case couldn't they just deploy a variant that removes this line and constructor argument? I think that addresses what you're pointing out, but is preferable to making the constructor argument here optional. WDYT ser?

jakerockland:
> In the process of writing tests today, I realized that the new Engine contract as currently written exceeds deployment size limits... going to need to dive a little bit deeper and loop back to sort out if this means that the required digging makes it totally unreasonable (was already a stretch) to try to get the desired Collaborations contracts deployed before EOY 😓
> 
> Screenshot 2022-12-15 at 5 57 55 PM

I've resolved the issue – a by-EOY deployment is still potentially possible. 😓 😅

jakerockland:
Still working on tests–once those are done and there is proper coverage + peace of mind from said coverage, will update the Diffchecker diffs above between V3 and V3_Engine.

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2038 @@
...
+    /**
+     * @notice Transfers ownership of the contract to a new account (`newOwner`).
+     * Internal function without access restriction.
+     * @param newOwner New owner.
+     * @dev owner role was called `admin` prior to V3 core contract.
+     * @dev Overrides and wraps OpenZeppelin's _transferOwnership function to
+     * also update adminACLContract for improved introspection.
+     */
+    function _transferOwnership(address newOwner) internal override {
+        Ownable._transferOwnership(newOwner);
```
yoshiwarab, line 1935: Should this be?
```suggestion
        Ownable._transferOwnership(IAdminACLV0(newOwner).superAdmin());
```

yoshiwarab, line 1935: Do we still want the OwnershipTransferred event to fire with the ACL contract address now that owner() is returning the super admin?

contracts/GenArt721CoreV3_Engine.sol
```diff
@@ -0,0 +1,2038 @@
...
+     * @dev ref: https://docs.openzeppelin.com/contracts/4.x/api/access#Ownable
+     * @dev owner role was called `admin` prior to V3 core contract
+     */
+    function owner()
+        public
+        view
+        override(Ownable, IGenArt721CoreContractV3_Engine)
+        returns (address)
+    {
+        return adminACLContract.superAdmin();
```
yoshiwarab, line 1877: If we've renounced ownership, this will now revert. Is that the expected behavior?

jakerockland:
```
- Update the EIP-173-conforming `owner()` method to proxy the `superAdmin` value for the currently assigned ACL contract, rather than returning the contract itself.
```

@ryley-o @yoshiwarab based on a conversation that Ben and I were having based on some weird behavior we realized was implied both in this PR and over in #395, we've decided to remove this behavior from the `_Engine` fork and from #395. 

tl;dr: It is a better division of responsibility to not leak the implementation details of the ACL management contract into the `Ownable` behavior within the Core.

contracts/GenArt721CoreV3_Engine.sol
jakerockland: Resolving, as we've reverted this change based on convo we've had 1:1 offline about the weird implications of returning the `superAdmin`.

Also noted here, for posterity: https://github.com/ArtBlocks/artblocks-contracts/pull/361#issuecomment-1360549182

contracts/GenArt721CoreV3_Engine.sol
jakerockland: Resolving, as we've reverted this change based on convo we've had 1:1 offline about the weird implications of returning the `superAdmin`.

Also noted here, for posterity: https://github.com/ArtBlocks/artblocks-contracts/pull/361#issuecomment-1360549182

jakerockland:
Updated the diffchecker diff log to account for https://github.com/ArtBlocks/artblocks-contracts/pull/361/commits/276460c67dcf1e0660f3f8097673d9d8b1c931a6

yoshiwarab (approved):
LGTM 🚀!

jakerockland:
V3 engine specific tests have been added in https://github.com/ArtBlocks/artblocks-contracts/pull/361/commits/73f958aebe6376a6144ea3c0dab9bfcc35395203

This is not 100% coverage _fully comprehensive_ of every code path, but it is comprehensive enough coverage that I feel comfortable merging this PR and follow up with a pre-EOY yeet-deploy of V3 Engine (contract only, infra work will come in '23) for BMG, Pace, and an internal use case.

jakerockland:
Mergin' will send separate PR will deployment scripts as an isolated smol PR.

ryley-o:
This is wonderful 🙏 Gave this a look over, and everything LGTM. Some minor nits in a follow-on #416

PR #362, opened 2022-10-25
Update hardhat config to use a `gasMultiplier`

jakerockland:
While working through https://github.com/ArtBlocks/artblocks/issues/298, I found that the recently introduced usage of `"auto"` gas pricing is generally WAI, with the caveat that it is a bit less aggressive than I think we want it to be, for DevX reasons. 

This updates our config to also us a `gasMultiplier` param of `1.5` (vs. the default of `1`) to ensure that TXs are processed in a bit more timely of a manner.

For additional context, see: https://hardhat.org/hardhat-runner/docs/config#networks-configuration

cc @Asupkay for vis, if visibility into this approach is at all relevant for some of the application automation work you were doing and allows for any simplification of your approach (may be what you're already doing based on recent convo).

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -62,11 +62,13 @@ module.exports = {
       url: MAINNET_JSON_RPC_PROVIDER_URL,
       accounts: [`0x${MAINNET_PRIVATE_KEY}`],
       gasPrice: "auto",
+      gasMultiplier: 1.5,
     },
     goerli: {
       url: GOERLI_JSON_RPC_PROVIDER_URL,
       accounts: [`0x${TESTNET_PRIVATE_KEY}`],
       gasPrice: "auto",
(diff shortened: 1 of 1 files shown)


jakerockland:
> I do think we should probably try to deploy w/tx type 1 or 2 instead of 0, which it looks like hardhat doesn't support by default 😞 I think we could specify in ethers directly though, see transactionRequest.maxFeePerGas in https://docs.ethers.io/v5/api/providers/types/#providers-TransactionRequest

IIUC this is just a minor cost optimization thing, no? I think we can file an issue against hardhat for this, but IMO this would never be reason to move _off of_ hardhat just for minor optimization in deployment costs. 

Agree though this shouldn't block this PR, its somewhat orthogonal. Mergin' 🚢

PR #364, opened 2022-10-25
QSP-2 (PR#299) Docs surrounding selfdestruct logic

ryley-o:
Add documentation explaining why we have included selfdestruct logic that enables bytecode-storage contacts to be removed from Ethereum's state.

In general, the comments indicate that we believe selfdestructing is the most ideal behavior at the time of writing/deploying this contract, even though it does not reduce gas costs, due to reducing the usage of Ethereum state. The comments also indicate that the design is compatible with all current proposals that remove the self-destruct opcode, since this implementation does not rely on bytecode being removed from the blockchain.


jakerockland (approved):
🙏 🙏 🙏 🙏

jakerockland:
LGTM @ryley-o 🙏 🙏  TYVM! Enabling auto-merge.

PR #368, opened 2022-10-27
Testing README.md

mchrupcala:
After @ryley-o and I paired on some unit testing, we were talking about our patterns around testing and I thought it could help if we wrote down a reference for tricks & patterns we try to follow. This is meant to be a very bare-bones starting point that we can add to over time :)


ryley-o (approved):
A good starting point - I prefer to add this and keep iterating

PR #370, opened 2022-10-28
Update to MinterMerkleV3

ryley-o:
Add MerkleMinterV3 as a derivative of MerkleMinterV2 to be chain-agnostic with respect to how it's delegation registry contract address is defined.

This was the result of some pair-coding w/ @mchrupcala 👥 💯 

This adds MerkleMinterV3 to not hard-code the delegation registry contract in it's solidity source code. In general, this is a preferential solidity pattern, because we would like our contracts to work across all chain-IDs and networks, where the delegation registry's address may be changed.

> Note that in this case, we don't likely expect the delegation registry contract to change since it was crated via `CREATE2` calls, but that cannot be guaranteed, especially on different L2s where the original CREATE2 factory contracts used by 0xfoobar may not be deployed.

## Design Choices and Implications

Changing the delegation registry contract to immutable also enables easier testing. Deploying a contract at a given address in hardhat is only possible by impersonating accounts and re-creating steps on ETH mainnet - deploying to a specific address is not allowed. With a configurable delegation registry address in the constructor, we are more easily able to add end-to-end tests that integrate with a delegation registry deployed locally in the hardhat network. This enables us to move away from mocking smart contract return values for specific calls, which we think was a contributing factor to the sequence of events that led to a MinterMerkleV2 minter that was deployed with a bug that would have prevented integration with the delegation registry.

Note that MinterMerkleV3 emits events when configuring the delegation registry contract. This was done to enable off-chain indexing of MinterMerkleV3's delegation registry contract.

Lastly, we have opted to stick with our current pattern of copying in source code of external contracts that our contracts integrate with into our repo, and compiling them on our own local hardhat network while testing. Alternatively, we could choose to fork mainnet, remove external contracts from our repo, add an alchemy api key to our environment variables, and test that way. We could also attempt to use git submodules, but in this specific instance, the delegation registry is defined in a foundry repo, making that solution not desirable. There are pluses and minuses of each of those choices, and this may warrant some follow-on discussion to determine which philosophy we prefer. It likely depends on how complex our integrations become in the future.


contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV2.sol
```diff
@@ -51,8 +51,7 @@ contract MinterMerkleV2 is ReentrancyGuard, IFilteredMinterMerkleV0 {
     using MerkleProof for bytes32[];
 
     /// Delegation registry address
-    address public constant DELEGATION_REGISTRY_ADDRESS =
-        0x00000000000076A84feF008CDAbe6409d2FE638B;
+    address public immutable DELEGATION_REGISTRY_ADDRESS;
```
ryley-o, line 54: we opted to keep ALL_CAPS_CASING here to remain compatible with the existing, deployed MinterMerkleV2 contracts. If we don't like this, we should roll to a MitnerMerkleV3 in this PR

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV2.sol
```diff
@@ -122,12 +121,16 @@ contract MinterMerkleV2 is ReentrancyGuard, IFilteredMinterMerkleV0 {
...
-        ReentrancyGuard()
-    {
+    constructor(
+        address _genArt721Address,
+        address _minterFilter,
+        address _delegationRegistryAddress
+    ) ReentrancyGuard() {
         genArt721CoreAddress = _genArt721Address;
         genArtCoreContract = IGenArt721CoreContractV3(_genArt721Address);
+        DELEGATION_REGISTRY_ADDRESS = _delegationRegistryAddress;
```
ryley-o, line 133: see PR description regarding why events are not emitted here - tldr: if we want to index this off-chain, we should update this to be a MinterMerkleV3

contracts/minter-suite/Minters/MinterMerkle/V3Notes.md
```diff
@@ -0,0 +1,4 @@
+## TODOs for MinterMerkleV3
+
+- Emit event when assigning DELEGATION_REGISTRY_ADDRESS to enable off-chain indexing
+  - This was a constant in the original, original version of MerkleMinterV2, so indexing is not possible on that minter (a constant or call to the contract must be used on the frontend)
```
mchrupcala, line 4: Very minor typo nit ("original" twice) 😅

mchrupcala (approved):

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV2.sol
jakerockland: No strong opinion here on casing personally. If we don't typically do it for other immutables, we probably don't need to do it here.

ryley-o:
I'm going to update this to roll the version to MinterMerkleV3 based on some of the initial feedback. I think that helps remedy some of the weirdness with events, casing, etc.

contracts/minter-suite/Minters/MinterMerkle/V3Notes.md
ryley-o: ty! ended up removing this file since we decided to roll these changes into a new V3 minter 😄

contracts/mock/delegation-registry/DelegationRegistry.sol
```diff
@@ -0,0 +1,672 @@
+// SPDX-License-Identifier: CC0-1.0
+pragma solidity ^0.8.17;
```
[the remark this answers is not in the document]
ryley-o: no good reason - moved!

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV2.sol
ryley-o: I've updated this to be our typical casing for immutables, especially since we rolled these updates into a new version of the minter.

contracts/minter-suite/Minters/MinterMerkle/MinterMerkleV2.sol
ryley-o: After some more thought, totally agree here. I've made the updates and bumped this PR to add a new MinterMerkleV3. This is a more clean solution, and we can make the corresponding subgraph updates when we add MinterHolderV2 to our indexing/frontend.

ryley-o:
@jakerockland - See note about changing this PR to roll to MerkleMinterV3. Would appreciate a re-review whenever you find time. 🙏 thank you!
Cc @mchrupcala

jakerockland (approved):
This LGTM (the diff I reviewed against: https://www.diffchecker.com/UZpYivL0)–thank you for updating! 🙏

PR #373, opened 2022-10-31
Improve testing readme

ryley-o:
Update testing readme with a more detailed overview of our testing goals, philosophy, and implementation details.

This is intended to help contributors when writing contracts and adding tests in the future.


README.md
```diff
@@ -248,6 +248,10 @@ After this config is finished, you'll notice a `usd (avg)` column in the auto-ge
   - [0x0E8BD86663e3c2418900178e96E14c51B2859957](https://etherscan.io/address/0x0E8BD86663e3c2418900178e96E14c51B2859957)
   - These are the Smart contract that received funds from primary sales and split them between the artist(s) and the platform. Artists received funds directly from this contract.These minter contracts are no longer in use.
 
+## Contributing
+
+All code must pass all tests and be formatted with prettier before being merged into the main branch. To run the tests, run `yarn test`. To format the code, run `yarn format`. For additional information about our testing philosophy at Art Blocks, please see our [Testing Philosophy](./test/README.md) document.
```
mchrupcala, line 253: Is `yarn lint` a requirement too?

test/README.md
```diff
@@ -1,14 +1,92 @@
...
+- We do not currently implement a static analysis tool. We should add a static analysis tool, such as [slither](https://github.com/crytic/slither).
+  - (captured in https://github.com/ArtBlocks/artblocks-contracts/issues/372)
+
+Potential improvements that have benefits and drawbacks:
+
+- Forking mainnet or testnets to run tests. This would allow us to test against real-world data, but would also require us to maintain a forked node API key, and would raise the level of complexity for contributors.
+- Using git submodules to pull in (more complex) external contracts, and run tests against those contracts.
+
+# Helpful Examples
+
```
mchrupcala, line 86: It's just 1 method but do you think it's worth adding this to speed things up for others?
```suggestion
## Isolate Unit Tests

`.only()` is a helpful pattern to use when you're working on your tests. With this method any tests that fall outside of your isolated describe block(s) will be ignored, and this can speed up your development/testing cycle by an order of magnitude.

  describe.only("common tests", async function () {
    await AdminACLV0V1_Common("AdminACLV0");
  });
```

mchrupcala (approved):

README.md
jakerockland: `yarn lint` just runs the prettier check against what `yarn format` performs, so I don't think we need to call it out separately (we already mention that must "be formatted with prettier").

test/README.md
ryley-o: I've went ahead and addressed both of your concerns in [this commit](https://github.com/ArtBlocks/artblocks-contracts/pull/373/commits/a44fb7cd8f79016cdcc7eed0fa75683adc0a77c4):
- Added @mchrupcala 's helpful tip (which is actively used by devs, even if we don't recommend in the readme)
- Added an eslint rule that causes failure if any tests use a `describe.only()` block

Thanks for the comments!

ryley-o:
>Could you please share this updated README more broadly in our eng-chat for visibility?

SGTM - will do once the CI runs and I merge this 👍

PR #382, opened 2022-11-08
Exponential DA w/Refund

ryley-o:
Add new minter - Dutch Auction with Refund.

This minter enables the following:
- User may purchase

### DIFF
--- a/contracts/interfaces/0.8.x/IFilteredMinterDAExpRefundV0.sol
+++ b/contracts/interfaces/0.8.x/IFilteredMinterDAExpRefundV0.sol
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+import "./IFilteredMinterV1.sol";
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title This interface extends the IFilteredMinterV1 interface in order to
+ * add support for generic project minter configuration updates.
+ * @dev keys represent strings of finite length encoded in bytes32 to minimize
+ * gas.
+ * @author Art Blocks Inc.
+ * TODO - derive from IFilteredMinterV0 if generic events are not needed
+ */
+interface IFilteredMinterDAExpRefundV0 is IFilteredMinterV1 {
+    /// Auction details updated for project `projectId`.
+    event SetAuctionDetails(
+        uint256 indexed projectId,
+        uint256 _auctionTimestampStart,
+        uint256 _priceDecayHalfLifeSeconds,
+        uint256 _startPrice,
+        uint256 _basePrice
+    );
+
+    /// Auction details cleared for project `projectId`.
+    event ResetAuctionDetails(uint256 indexed projectId);
+
+    /// Maximum and minimum allowed price decay half lifes updated.
+    event AuctionHalfLifeRangeSecondsUpdated(
+        uint256 _minimumPriceDecayHalfLifeSeconds,
+        uint256 _maximumPriceDecayHalfLifeSeconds
+    );
+
+    /// sellout price updated for project `projectId`.
+    event SelloutPriceUpdated(uint256 _projectId, uint256 _selloutPrice);
+
+    /// admin validated a sellout price > base price for project `projectId`.
+    event SelloutPriceValidated(uint256 _projectId, uint256 _selloutPrice);
+
+    /// artist and admin have withdrawn revenues from refundable purchases for
+    /// project `projectId`.
+    event ArtistAndAdminRevenuesWithdrawn(uint256 _projectId);
+}
--- a/contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
+++ b/contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
@@ -603,14 +603,16 @@ contract MinterDAExpRefundV0 is ReentrancyGuard, IFilteredMinterDAExpRefundV0 {
     {
         // CHECKS
         // get the current price, which returns the sellout price if the
-        // auction sold out before reaching base price. Reverts if
-        // auction is unconfigured or has not started.
+        // auction sold out before reaching base price, or returns the base
+        // price if auction has reached base price without reaching max
+        // invocations on this minter. Reverts if auction is unconfigured or
+        // has not started.
         uint256 currentPriceInWei = _getPrice(_projectId);
 
         // EFFECTS
         // calculate the refund amount
         Receipt storage receipt = receipts[msg.sender][_projectId];
-        // implicit overflow check in solidity ^0.8
+        // implicit overflow/underflow checks in solidity ^0.8
         uint256 amountDue = receipt.numPurchased * currentPriceInWei;
         uint256 refund = receipt.netPaid - amountDue;
         // reduce the netPaid (in storage) to value after refund deducted
--- a/contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
+++ b/contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
@@ -1,0 +1,834 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+import "../../../interfaces/0.8.x/IGenArt721CoreContractV3.sol";
+import "../../../interfaces/0.8.x/IMinterFilterV0.sol";
+import "../../../interfaces/0.8.x/IFilteredMinterDAExpRefundV0.sol";
+
+import "@openzeppelin-4.5/contracts/security/ReentrancyGuard.sol";
+import "@openzeppelin-4.5/contracts/utils/math/SafeCast.sol";
+
+pragma solidity 0.8.17;
+
+/**
+ * @title Filtered Minter contract that allows tokens to be minted with ETH.
+ * Pricing is achieved using an automated Dutch-auction mechanism, with a
+ * refund mechanism for tokens purchased before the auction ends.
+ * This is designed to be used with IGenArt721CoreContractV3 contracts.
+ * @author Art Blocks Inc.
+ * @notice Privileged Roles and Ownership:
+ * This contract is designed to be managed, with limited powers.
+ * Privileged roles and abilities are controlled by the core contract's Admin
+ * ACL contract and a project's artist. Both of these roles hold extensive
+ * power and can modify minter details.
+ * Care must be taken to ensure that the admin ACL contract and artist
+ * addresses are secure behind a multi-sig or other access control mechanism.
+ * ----------------------------------------------------------------------------
+ * The following functions are restricted to the core contract's Admin ACL
+ * contract:
+ * - setAllowablePriceDecayHalfLifeRangeSeconds (note: this range is only
+ *   enforced when creating new auctions)
+ * - resetAuctionDetails (note: this will prevent minting until a new auction
+ *   is created)
+ * - TODO
+ * ----------------------------------------------------------------------------
+ * The following functions are restricted to a project's artist:
+ * - setAuctionDetails (note: this may only be called when there is no active
+ *   auction)
+ * - TODO
+ * ----------------------------------------------------------------------------
+ * Additional admin and artist privileged roles may be described on other
+ * contracts that this minter integrates with.
+ *
+ * @dev Note that while this minter makes use of `block.timestamp` and it is
+ * technically possible that this value is manipulated by block producers, such
+ * manipulation will not have material impact on the price values of this minter
+ * given the business practices for how pricing is congfigured for this minter
+ * and that variations on the order of less than a minute should not
+ * meaningfully impact price given the minimum allowable price decay rate that
+ * this minter intends to support.
+ */
+contract MinterDAExpRefundV0 is ReentrancyGuard, IFilteredMinterDAExpRefundV0 {
+    using SafeCast for uint256;
+
+    /// Core contract address this minter interacts with
+    address public immutable genArt721CoreAddress;
+
+    /// This contract handles cores with interface IV3
+    IGenArt721CoreContractV3 private immutable genArtCoreContract;
+
+    /// Minter filter address this minter interacts with
+    address public immutable minterFilterAddress;
+
+    /// Minter filter this minter may interact with.
+    IMinterFilterV0 private immutable minterFilter;
+
+    /// minterType for this minter
+    string public constant minterType = "MinterDAExpRefundV0";
+
+    uint256 constant ONE_MILLION = 1_000_000;
+
+    struct ProjectConfig {
+        // this value is updated only if maxHasBeenInvoked is true, and
+        // represents the purchase price of the final token during a sellout
+        // auction. If a sellout is not achieved before the auction ends (i.e.
+        // reaches base price), this value will be 0.
+        uint256 selloutPrice;
+        // on this minter, hasMaxBeenInvoked is updated only during every
+        // purchase, and is only true if this minter minted the final token.
+        // this enables the minter to know when a sellout price is greater than
+        // the auction's base price.
+        bool maxHasBeenInvoked;
+        // set to true by the admin if the auction has ended and the sellout
+        // or base price is considered valid.
+        bool auctionIsValid;
+        // number of tokens minted that have potential of future refunds.
+        // max uint24 > 16.7 million tokens > 1 million tokens/project max
+        uint24 numRefundableInvocations;
+        // max uint64 ~= 1.8e19 sec ~= 570 billion years
+        uint64 timestampStart;
+        uint64 priceDecayHalfLifeSeconds;
+        uint256 startPrice;
+        uint256 basePrice;
+    }
+
+    mapping(uint256 => ProjectConfig) public projectConfig;
+
+    /// Minimum price decay half life: price must decay with a half life of at
+    /// least this amount (must cut in half at least every N seconds).
+    uint256 public minimumPriceDecayHalfLifeSeconds = 300; // 5 minutes
+    /// Maximum price decay half life: price may decay with a half life of no
+    /// more than this amount (may cut in half at no more than every N seconds).
+    uint256 public maximumPriceDecayHalfLifeSeconds = 3600; // 60 minutes
+
+    struct Receipt {
+        uint256 netPaid;
+        uint256 numPurchased;
+    }
+    /// user address => project ID => receipt
+    mapping(address => mapping(uint256 => Receipt)) receipts;
+
+    // modifier to restrict access to only AdminACL or the artist
+    modifier onlyCoreAdminACLOrArtist(uint256 _projectId, bytes4 _selector) {
+        require(
+            (msg.sender ==
+                genArtCoreContract.projectIdToArtistAddress(_projectId)) ||
+                (
+                    genArtCoreContract.adminACLAllowed(
+                        msg.sender,
+                        address(this),
+                        _selector
+                    )
+                ),
+            "Only Artist or Admin ACL"
+        );
+        _;
+    }
+
+    // modifier to restrict access to only AdminACL allowed calls
+    // @dev defers which ACL contract is used to the core contract
+    modifier onlyCoreAdminACL(bytes4 _selector) {
+        require(
+            genArtCoreContract.adminACLAllowed(
+                msg.sender,
+                address(this),
+                _selector
+            ),
+            "Only Core AdminACL allowed"
+        );
+        _;
+    }
+
+    modifier onlyArtist(uint256 _projectId) {
+        require(
+            (msg.sender ==
+                genArtCoreContract.projectIdToArtistAddress(_projectId)),
+            "Only Artist"
+        );
+        _;
+    }
+
+    /**
+     * @notice Initializes contract to be a Filtered Minter for
+     * `_minterFilter`, integrated with Art Blocks core contract
+     * at address `_genArt721Address`.
+     * @param _genArt721Address Art Blocks core contract address for
+     * which this contract will be a minter.
+     * @param _minterFilter Minter filter for which
+     * this will a filtered minter.
+     */
+    constructor(address _genArt721Address, address _minterFilter)
+        ReentrancyGuard()
+    {
+        genArt721CoreAddress = _genArt721Address;
+        genArtCoreContract = IGenArt721CoreContractV3(_genArt721Address);
+        minterFilterAddress = _minterFilter;
+        minterFilter = IMinterFilterV0(_minterFilter);
+        require(
+            minterFilter.genArt721CoreAddress() == _genArt721Address,
+            "Illegal contract pairing"
+        );
+    }
+
+    /**
+     * @notice This function is not implemented on this minter, and exists only
+     * for interface conformance reasons. This minter checks if max invocations
+     * have been reached during every purchase to determine if a sellout has
+     * occurred. Therefore, the local caching of max invocations is not
+     * beneficial or necessary.
+     */
+    function setProjectMaxInvocations(
+        uint256 /*_projectId*/
+    ) external pure {
+        // not implemented because maxInvocations must be checked during every mint
+        // to know if final price should be set
+        revert(
+            "setProjectMaxInvocations not implemented - updated during every mint"
+        );
+    }
+
+    /**
+     * @notice Warning: Disabling purchaseTo is not supported on this minter.
+     * This method exists purely for interface-conformance purposes.
+     */
+    function togglePurchaseToDisabled(uint256 _projectId)
+        external
+        view
+        onlyArtist(_projectId)
+    {
+        revert("Action not supported");
+    }
+
+    /**
+     * @notice projectId => has project reached its maximum number of
+     * invocations while being minted with this minter?
+     * Note that this returns a local cache of the core contract's
+     * state, and may be out of sync with the core contract. This is
+     * intentional. A false negative will only result in a gas cost increase,
+     * since the core contract will still enforce max invocations during during
+     * minting. A false negative will also only occur if the max invocations
+     * was either reduced on the core contract to equal current invocations, or
+     * if the max invocations was reached by minting on a different minter.
+     * In both of these cases, we expect the net purchase price (after refund)
+     * shall be the base price of the project's auction. This prevents an
+     * an artist from benefiting by reducing max invocations on the core mid-
+     * auction, or by minting on a different minter.
+     * Note that if an artist wishes to reduce the max invocations on the core
+     * to something less than the current invocations, but more than max
+     * invocations (with the hope of increasing the sellout price), an admin
+     * function is provided to manually reduce the sellout price to a lower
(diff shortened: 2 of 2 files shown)


ryley-o:
@jakerockland - would appreciate your thoughts and review on this initial implementation, before I go write a bunch of tests.

diff between DAExpV2 and DAExpRefundV0: https://www.diffchecker.com/5X7KY4xK

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,786 @@
...
+        );
+        require(
+            _newSelloutPrice >= _projectConfig.basePrice,
+            "May only reduce sellout price to base price or greater"
+        );
+        require(
+            _projectConfig.auctionIsValid == false,
+            "Auction already marked as valid"
+        );
+        _projectConfig.selloutPrice = _newSelloutPrice;
```
jakerockland, line 420: Should this method also toggle the value of `auctionIsValid`?

jakerockland: AH nevermind–I see why now this is not relevant, given the below method.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,786 @@
...
+        external
+        onlyCoreAdminACLOrArtist(
+            _projectId,
+            this.withdrawArtistAndAdminRevenues.selector
+        )
+    {
+        ProjectConfig storage _projectConfig = projectConfig[_projectId];
+        // get the current net price of the auction
+        uint256 _price = _getPrice(_projectId);
+        // if the price is not base price, the auction is only valid if the
```
jakerockland, line 478: May be worth noting that this is a reasonable assumption given that price is monotonically decreasing.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,786 @@
...
+        // reaches base price), this value will be 0.
+        uint256 selloutPrice;
+        // on this minter, hasMaxBeenInvoked is updated only during every
+        // purchase, and is only true if this minter minted the final token.
+        // this enables the minter to know when a sellout price is greater than
+        // the auction's base price.
+        bool maxHasBeenInvoked;
+        // set to true by the admin if the auction has ended and the sellout
+        // or base price is considered valid.
+        bool auctionIsValid;
```
jakerockland, line 84: I think we should name this something a bit more explicit like `auctionIsVerifiedComplete`.

jakerockland (approved):
Added some minor comments, but directionally really love this approach and LGTM to proceed re: testing coverage, etc.

jakerockland:
cc @mchrupcala @lyaunzbe  for added vis.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: cool, resolving 🙏

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: updated!

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: good call, updated!

ryley-o:
I've added events that are emitted every time a user's receipt is updated in e17c8ae

ryley-o:
Added functions that enable a user to claim refunds across multiple projects in a single transaction in commit b839fd0

ryley-o:
@jakerockland - I've updated this PR's description post merge of #386, and tagged you in a design-ish question in the TODO section 🙏  Also tagged you in a note in ## Trust Assumptions 🙏

ryley-o:
@jakerockland - tagging this as ready for review now that tests have been added and some minor iterations have been completed. Although you have already given a ✅ , would appreciate a second read-though of the minter.

diffchecker between DAExpV2 and DAExpRefundV0 here: https://www.diffchecker.com/WnpFrP0c

Any suggestions on any additional reviewers for this?

jakerockland:
> @jakerockland - tagging this as ready for review now that tests have been added and some minor iterations have been completed. Although you have already given a ✅ , would appreciate a second read-though of the minter.
> 
> diffchecker between DAExpV2 and DAExpRefundV0 here: https://www.diffchecker.com/WnpFrP0c
> 
> Any suggestions on any additional reviewers for this?

I will do an initial review pass by EOD–but given the sensitivity of this functionality lets add @yoshiwarab @lyaunzbe and @mchrupcala as additional reviewers here as well.

ryley-o:
I do want to note that there is probably some gas golf that can be played on this minter, but I'm comfortable merging and doing that as follow-on work. totally open to feedback that improves gas efficiency here though!

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,965 @@
...
+        require(_to != address(0), "No refund to the zero address");
+        // EFFECTS
+        // for each project, tally up the refund amount and update the receipt
+        // in storage
+        uint256 refund;
+        uint256 projectIdsLength = _projectIds.length;
+        for (uint256 i = 0; i < projectIdsLength; ) {
+            uint256 projectId = _projectIds[i];
+            ProjectConfig storage _projectConfig = projectConfig[projectId];
+            Receipt storage receipt = receipts[msg.sender][projectId];
```
jakerockland, line 782: Given how this is abstracted, I believe this should be checking `receipts[_to][_projectId]` instead, no?

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,965 @@
...
+            // price if auction has reached base price and artist has withdrawn
+            // revenues.
+            // @dev if user is elligible for a refund, they have purchased a token,
+            // therefore we are guaranteed to not have a populated
+            // latestPurchasePrice
+            uint256 netTokenPrice = _projectConfig.latestPurchasePrice;
+            // calculate the refund amount
+            // implicit overflow/underflow checks in solidity ^0.8
+            uint256 amountDue = receipt.numPurchased * netTokenPrice;
+            refund += receipt.netPaid - amountDue;
```
jakerockland, line 800: Nit: I would put the RHS in parens just to make this a bit easier to read when combined with the `+=`

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,965 @@
...
+            // reduce the netPaid (in storage) to value after refund deducted
+            receipt.netPaid = amountDue;
+            // emit event indicating new receipt state
+            emit ReceiptUpdated(
+                msg.sender,
+                projectId,
+                receipt.numPurchased,
+                receipt.netPaid
+            );
+            // gas efficiently increment i
```
jakerockland, line 810: Nit: may want to explain that this is safe because we run out of gas way before we overflow

jakerockland (approved):
Did another pass–and generally LGTM @ryley-o 🔥 🙌 

Opened up a handful of very smol comments, PLMK if you'd like me to another pass before merging, but would be great to get some additional eyes from at least one of @mchrupcala @yoshiwarab or @lyaunzbe I think as well before merging; please adjust the reviewer set as you think makes sense 🙏 💜 

Great work!!

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,965 @@
...
+     * Sends refund to address `_to`.
+     * @param _to Address to send refund to.
+     * @param _projectId Project ID to refund payment on.
+     */
+    function claimRefundTo(address payable _to, uint256 _projectId)
+        public
+        nonReentrant
+    {
+        ProjectConfig storage _projectConfig = projectConfig[_projectId];
+        Receipt storage receipt = receipts[msg.sender][_projectId];
```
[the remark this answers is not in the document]
ryley-o: nope - consider where addr `X` made purchase, and now has refund available. addr `Y` is some other random addr.
`X` must call this function, and is instructing to send the refund amount to `Y`. In that case, the Receipt is located at `receipts[X][_projectId]` (during the original purchase by `X`), so we point to that Receipt.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,965 @@
...
+        // CHECKS
+        // input validation
+        require(_to != address(0), "No refund to the zero address");
```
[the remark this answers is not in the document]
ryley-o: `_to` is an input parameter to this public function, so I think input validation to help prevent users from sending funds to the zero address makes sense (e.g. they forgot to populate the field on etherscan or something)

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,965 @@
...
+     * @param _to Address to send refund to.
+     * @param _projectIds Array of project IDs to refund payments on.
+     */
+    function claimRefundsTo(address payable _to, uint256[] memory _projectIds)
+        public
+        nonReentrant
+    {
+        // CHECKS
+        // input validation
+        require(_to != address(0), "No refund to the zero address");
```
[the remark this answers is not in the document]
ryley-o: `_to` is an input parameter to this public function, so I think input validation to help prevent users from sending funds to the zero address makes sense (e.g. they forgot to populate the field on etherscan or something)

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: (see my response for analogous comment in `claimRefundTo` comment 🙏 )

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: good call, updated!

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: good call, updated!

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,965 @@
...
+    function setAuctionDetails(
+        uint256 _projectId,
+        uint256 _auctionTimestampStart,
+        uint256 _priceDecayHalfLifeSeconds,
+        uint256 _startPrice,
+        uint256 _basePrice
+    ) external onlyArtist(_projectId) {
+        // CHECKS
+        ProjectConfig storage _projectConfig = projectConfig[_projectId];
+        require(
```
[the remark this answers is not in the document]
ryley-o: yeah, adding that constraint may simplify some of our logic elsewhere (e.g. non-zero latestPurchasePrice would now indicate the auction has purchases). I'll take a look at how this ripples through this minter, and if it looks like a net-positive, might open a separate PR to merge into this one.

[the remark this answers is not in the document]
ryley-o: @jakerockland - I went ahead and implemented this in the following commit: [cefd2e0](https://github.com/ArtBlocks/artblocks-contracts/pull/382/commits/cefd2e0bf7c595f3631b613b1c9a2f4745173ba5)

I think a net benefit to how this minter may be rationalized. I'm not completely sure if this enabled the removal of `selloutPrice` state variable in that commit, but somewhat related since it all relates to improved rationalizing about what has been purchased when 😄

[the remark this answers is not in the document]
jakerockland: Rad rad rad – yeah this is great, removal entirely of `selloutPrice` is a nice perk! LGTM.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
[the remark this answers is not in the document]
jakerockland: Ohhhhh–I misunderstood intended use here. Yep, this entirely makes sense (and is also aligned with the documentation, that's my bad). LGTM.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
jakerockland: Ohhhhh–I misunderstood intended use here. Yep, this entirely makes sense (and is also aligned with the documentation, that's my bad). LGTM.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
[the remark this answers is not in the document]
jakerockland: Yep yep yep–I see I see; makes sense to me (my misreading).

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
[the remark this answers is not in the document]
jakerockland: Yep yep yep–I see I see; makes sense to me (my misreading).

ryley-o:
@jakerockland - heads-up that I added a convenience getter function to help with subgraph code efficiency in commit [7a7bf22](https://github.com/ArtBlocks/artblocks-contracts/pull/382/commits/7a7bf22798c101baf85acd52c8a78506ea2b3744) 🙏

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,997 @@
...
+        // input validation
+        require(_to != address(0), "No refund to the zero address");
+        // require that a user has purchased at least one token on this project
+        require(receipt.numPurchased > 0, "No purchases made by this address");
+        // get the latestPurchasePrice, which returns the sellout price if the
+        // auction sold out before reaching base price, or returns the base
+        // price if auction has reached base price and artist has withdrawn
+        // revenues.
+        // @dev if user is elligible for a refund, they have purchased a token,
+        // therefore we are guaranteed to not have a populated
```
mchrupcala, line 709: Nit: would you mind adding an extra note here please? I wasn't sure why (for an ongoing auction) wouldn't `latestPurchasePrice` be populated if the user bought a token.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,997 @@
...
+        // latestPurchasePrice
+        uint256 netTokenPrice = _projectConfig.latestPurchasePrice;
+
+        // EFFECTS
+        // calculate the refund amount
+        // implicit overflow/underflow checks in solidity ^0.8
+        uint256 amountDue = receipt.numPurchased * netTokenPrice;
```
mchrupcala, line 716: Nit: would something like "netPaidAfterRefund" or "baseCost" be more accurate here vs. "amountDue" maybe?

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,997 @@
...
+            // auction sold out before reaching base price, or returns the base
+            // price if auction has reached base price and artist has withdrawn
+            // revenues.
+            // @dev if user is elligible for a refund, they have purchased a token,
+            // therefore we are guaranteed to not have a populated
+            // latestPurchasePrice
+            uint256 netTokenPrice = _projectConfig.latestPurchasePrice;
+            // calculate the refund amount
+            // implicit overflow/underflow checks in solidity ^0.8
+            uint256 amountDue = receipt.numPurchased * netTokenPrice;
```
mchrupcala, line 797: Nit: would something like "netPaidAfterRefund" or "baseCost" be more accurate here vs. "amountDue" maybe?

mchrupcala (changes requested):
The design & the logic are 🔥 🔥 🔥 I found some nits and small sections of the logic that I had questions about, and overall this looks amazing!

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,997 @@
...
+        // reduce the netPaid (in storage) to value after refund deducted
+        receipt.netPaid = amountDue;
+        // emit event indicating new receipt state
+        emit ReceiptUpdated(
+            msg.sender,
+            _projectId,
+            receipt.numPurchased,
+            receipt.netPaid
+        );
+
```
mchrupcala, line 727: Could it help us either to emit a ReceiptClaimed event here, or change ReceiptUpdated to add a `totalClaimed` maybe?

I'm thinking the website will need to send a list of `_projectIDs` to `claimRefundsTo()`, and that it will need to retrieve some map of {projectID: numRefundableInvocations, ...} per user. This is something we could maybe index on the subgraph, but I think we'd need an event trigger to track & delete `numRefundableInvocations` after a claim's been made.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,997 @@
...
+            "Must send minimum value to mint"
+        );
+        // emit event indicating new receipt state
+        emit ReceiptUpdated(
+            msg.sender,
+            _projectId,
+            _receipt.numPurchased,
+            _receipt.netPaid
+        );
+
```
mchrupcala, line 621: Same comment as below -- Could it help us either to emit a ReceiptClaimed event here, or change ReceiptUpdated to add a totalClaimed maybe?

contracts/interfaces/0.8.x/IFilteredMinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    /// receipt has an updated state
+    event ReceiptUpdated(
+        address indexed _purchaser,
+        uint256 indexed _projectId,
+        uint256 _netPaid,
+        uint256 _numPurchased
+    );
+
+    /// returns latest purchase price for project `_projectId`, or 0 if no
```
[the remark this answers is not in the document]
ryley-o: We made a design decision a while ago to not allow a base price of zero. It is documented here: https://github.com/ArtBlocks/artblocks-contracts/pull/382#discussion_r1025882588

Overall, not allowing zero just makes the entire system easier to rationalize about, one example being that a value of zero returned here is guaranteed to mean no purchases have been made 😄 

Not allowing "potentially free mints" seems like a reasonable constraint on this minter, especially since ~the same effect could be made by having a base price of 1 wei.

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,997 @@
...
+    struct Receipt {
+        uint256 netPaid;
+        uint256 numPurchased;
+    }
+    /// user address => project ID => receipt
+    mapping(address => mapping(uint256 => Receipt)) receipts;
+
+    // modifier to restrict access to only AdminACL or the artist
+    modifier onlyCoreAdminACLOrArtist(uint256 _projectId, bytes4 _selector) {
```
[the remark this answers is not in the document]
ryley-o: `_selector` here is the [4-byte function selector](https://docs.soliditylang.org/en/latest/abi-spec.html#function-selector) of the function being called. It is used when asking the AdminACL contract if msg.sender has permission to call the function on this contract.
This modifier allows either artist or admin to call, so we need to know the project Id (to determine if the artist is msg.sender) and the function selector (to determine if an allowed admin is msg.sender)

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
```diff
@@ -0,0 +1,997 @@
...
+        // _getPrice reverts if auction is unconfigured or has not started
+        uint256 currentPriceInWei = _getPrice(_projectId);
+
+        // EFFECTS
+        // update the purchaser's receipt and require sufficient net payment
+        Receipt storage _receipt = receipts[msg.sender][_projectId];
+        _receipt.netPaid += msg.value;
+        _receipt.numPurchased++;
+        require(
+            _receipt.netPaid >= _receipt.numPurchased * currentPriceInWei,
```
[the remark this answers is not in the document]
ryley-o: Revert invalidates all state changes in a given call, so it would not lead to an inaccurate receipt here.
I like the readability of the current code just because it is very easy to rationalize "require that the user paid >= price*quantity".

That being said, we are mixing a check after some effects. However, because the effects are all very local and easy to see that no reentrancy can occur, there is no cause for concern here.

There would be a way to improve gas efficiency here, but it would lead to less readable code, so 🤷 seems not worth it.

I think this is okay as-is, but I'll let you resolve or continue the discussion @mchrupcala

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: good callout - I've updated this comment to appropriately say `therefore we are guaranteed to have a populated...` - the `not` was incorrect, but code logic is correct 🙏  (also updated by misspelling of eligible 😓)

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: hmm amountDue makes the most sense in my head, but I could be living in my own world here 😆 @mchrupcala and @jakerockland - any other thoughts on variable name here?

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: If all we know is a user's netPaid and numPurchased, as well as the project's latestPurchasePrice on the minter, we can index a stable calculated field of "excess claimable funds" simply equal to `netPaid - (numPurchased * latestPurchasePrice)`. Happy to chat about this more, but this lets us be very efficient in our indexing and querying of "which projects does this user address have claimable funds > 0 ETH on?"

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: <see other response>

ryley-o:
@mchrupcala - thank you for your review! I've pushed an update to fix typo. please see my responses and either resolve or continue the convo 🙏 ✌️

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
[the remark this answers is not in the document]
mchrupcala: Nope I 100% agree with you re: readability -- especially for the sensitive nature of the refund's math, I agree that it's great to prioritize readability over gas-savings here. Thanks for confirming that a revert would invalidate any state changes. Looks great to me!

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
mchrupcala: Ah 😅 I think I see it -- so "amount due" reads to me like it's owed to the user (and so is the refund), which is why it read strangely to me when subtracting `amountDue` to calculate `refund`. But this `amountDue` is actually owed to a split of the artist/artist payee/AB. It's a very minor nit but what about `amountDueToPayees` or something similar?

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
mchrupcala: Awesome, yeah "Claimable funds" will def be helpful but I'm not sure how the front-end will efficiently be able to grab a user's list of all Project ID's for which a refund claim is available. 

Judging by the new Receipt entity we're adding on the subgraph, we'll be able to loop all the receipts for any given user and get their project IDs. So it's outside the scope of our smart contract but do you think this is something we'd maybe want a computed field for, too?

ryley-o: Potentially, but I think the design document currently recommends just filtering for receipts with excess claimable funds greater than zero in the frontend graphql query.

I'm going to resolve this here for this PR's sake, but definitely chat with @Jbern16 about this (as he will be implementing the frontend portion of this minter) 🙏

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: (see discussion above for resolution of this comment 🙏 )

contracts/minter-suite/Minters/MinterDAExpRefund/MinterDAExpRefundV0.sol
ryley-o: good call, I see how that is confusing! I've pushed an update to change from `amountDue` to `requiredAmountPosted` in c94d151. That kind of jives with `receipt.netPosted` I think?

ryley-o: resolving, and we can iterate on this going forward!

mchrupcala (approved):

PR #384, opened 2022-11-08
Add delay functionality to DA w/Refund PR

ryley-o:
Add functionality to delay an ongoing DA w/Refund. Intended to be merged into #382.

This is specifically implemented to handle the following type of failure mode:
- DA w/Refund is started, a few users mint directly from contract (preventing auction from being reset), but Art Blocks website is down, making it difficult for most users to purchase tokens

Instead of scrambling to get the website up and running before the auction reaches base price (which may be lower than perceived value), the artist or Art Blocks may put the DA in "delay" mode. "Delay" mode keeps price at a constant value until the delay is ended, at which point it continues along its previously programmed price curve, just shifted to the right.

This type of failure mode is good to handle based off of our previous experiences.

Drawbacks of introducing this kind of delay functionality are:
- Adds complexity when estimating auction complete time (still doable)
- Adds complexity when plotting auction price curves
- Adds complexity to frontend by adding an additional possible state of being in a "delay period"
- Adds complexity to indexing
- Adds complexity to smart contract logic
- Adds complexity to admin auction state management/operations

## Design Decisions
- It was elected to not disable purchases while in an active delay period. This prevents the poor UX of submitting a purchase transaction before a delay has started, but having it be reverted if a delay was started between when the tx was sent, and it was included in a block.

## Alternative Solutions to handle this failure mode

Alternative solutions to handle this PR's described failure mode include:
- Adding capability to reset an auction even after it has started and people make refundable bids. Some schema to properly provide refunds to the original bidders could be drawn up.
- Providing an alternate site to mint DA w/Refund auctions - something that could be ran locally by users in the case that our frontend is not working at the time of mint.

### DIFF
--- a/contracts/interfaces/0.8.x/IFilteredMinterDAExpRefundV0.sol
+++ b/contracts/interfaces/0.8.x/IFilteredMinterDAExpRefundV0.sol
@@ -55,4 +55,13 @@ interface IFilteredMinterDAExpRefundV0 is IFilteredMinterV1 {
         uint256 _netPaid,
         uint256 _numPurchased
     );
+
+    /// receipt has an updated state
+    event DelayStarted(
+        uint256 indexed _projectId,
+        uint256 _activeDelayBeginTimestamp
+    );
+
+    /// receipt has an updated state
+    event DelayEnded(uint256 indexed _projectId, uint256 _delaySecondsAdded);
 }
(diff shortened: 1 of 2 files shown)


ryley-o:
@jakerockland - no tests or anything, but curious for your thoughts on this functionality being merged into #384.

This seems like an important failure mode for us to be able to gracefully handle, despite the added complexity.

ryley-o:
Based on some conversations and brainstorming w/ @jakerockland, happily closing this in favor of searching for a simpler implementation to handle the described failure mode in this PR.

PR #388, opened 2022-11-15
Engine-V2-compatible exponential DA minter.

jakerockland:
This PR introduces a exponential DA minter that is compatible with the current existing V2 Engine (PBAB/PRTNR) contracts.

---

In addition to the necessary "migration changes" to make this compatible with V2 contracts, this minter also notably allows for the state of the start-price and end-price of the auction being the same (for use in "fixed price mode") which is not enabled on the minter-suite compatible variant (due to this not being necessary in a post-minter-suite world).

---

Diff between this new Engine-V2 compatible minter and the V3-suite variant it is forked from: https://www.diffchecker.com/jiU8B5ml/

---

Note: additional test coverage has been added for this new minter via `GenArt721CoreV2_EngineDAMinter_Integration`; however, this new test suite is intended to only cover cases that are not covered by the V3-suite variant test coverage (which is quite comprehensive) which namely means covering the updated payment splitting logic flow, the "fixed price mode" flow, and the misc. ACL-check differences, but not the core-automated-DA logic itself which is unchanged. 

This trade-off is made in the interest of time given the good test coverage for `MinterDAExpV3` and given that this V2-variant is only intended to be something we maintain/deploy for a limited amount of time until the V3-Engine contracts are ready.


contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,584 @@
...
+ *
+ * @dev Note that while this minter makes use of `block.timestamp` and it is
+ * technically possible that this value is manipulated by block producers, such
+ * manipulation will not have material impact on the price values of this minter
+ * given the business practices for how pricing is congfigured for this minter
+ * and that variations on the order of less than a minute should not
+ * meaningfully impact price given the minimum allowable price decay rate that
+ * this minter intends to support.
+ */
+contract GenArt721MinterDAExp_PBAB is ReentrancyGuard {
```
ryley-o, line 49: WDYT about calling this `MinterDAExpEngineV0`? - The GenArt721 prefix is something we dropped off of our minter suite to simplify the naming, and I think probably makes sense to do the same here (also applies to file name and `minterType` constant below)

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,582 @@
...
+     * @dev this enables gas reduction after maxInvocations have been reached -
+     * core contracts shall still enforce a maxInvocation check during mint.
+     * @dev function is intentionally not gated to any specific access control;
+     * it only syncs a local state variable to the core contract's state.
+     */
+    function setProjectMaxInvocations(uint256 _projectId) external {
+        uint256 maxInvocations;
+        (, maxInvocations, , , , ) = genArtCoreContract.projectStateData(
+            _projectId
+        );
```
ryley-o, line 148: IIUC, `projectStateData` is not available on PBAB core, so this needs to be updated

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,582 @@
...
+                uint256 artblocksRevenue_,
+                address payable artblocksAddress_,
+                uint256 artistRevenue_,
+                address payable artistAddress_,
+                uint256 additionalPayeePrimaryRevenue_,
+                address payable additionalPayeePrimaryAddress_
+            ) = genArtCoreContract.getPrimaryRevenueSplits(
+                    _projectId,
+                    _currentPriceInWei
+                );
```
ryley-o, line 466: This function does not exist on the PBAB core, so new logic will be needed here. Additionally, this function needs to handle payment to an additional fourth address, the render provider.

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,582 @@
...
+     * @dev Note that it is intentionally supported here that the configured
+     * price may be explicitly set to `0`.
+     */
+    function setAuctionDetails(
+        uint256 _projectId,
+        uint256 _auctionTimestampStart,
+        uint256 _priceDecayHalfLifeSeconds,
+        uint256 _startPrice,
+        uint256 _basePrice
+    ) external onlyArtist(_projectId) {
```
ryley-o, line 287: we should add a modifier `onlyValidProjectId(_projectId)`, same as what we have on our `MinterFilterV1.setMinterForProject` function.
This is because:
- PBAB core does not check for valid project ID when setting an artist address on a project (e.g. could set a future project's artist address by mistake)
- If this function does not check for a valid project ID, an auction could be configured before a project is created, resulting in a dangerous and unexpected state

ryley-o (commented):
Looking good overall - Overall this is a relatively simple port.
Left a few comments for you!

jakerockland:
Closing this out as we do not intend to use this anymore, per conversation with @cooper-temple, and it is likely that we will never want to use this given V3 Engine contract plans, and given some early-stage thinking around things along the lines of https://github.com/ArtBlocks/artblocks-contracts/issues/389

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,584 @@
...
+        require(
+            genArtCoreContract.adminACLAllowed(
+                msg.sender,
+                address(this),
+                _selector
+            ),
+            "Only Core AdminACL allowed"
+        );
+        _;
+    }
```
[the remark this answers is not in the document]
jakerockland: TY 🙏 👑 – updated!

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,584 @@
...
+    );
+
+    /// Core contract address this minter interacts with
+    address public immutable genArt721CoreAddress;
+
+    /// This contract handles cores with interface IV3
+    IGenArt721CoreV2_PBAB private immutable genArtCoreContract;
+
+    /// minterType for this minter
+    string public constant minterType = "GenArt721MinterDAExp_PBAB";
```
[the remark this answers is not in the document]
jakerockland: No strong opinion from me on the naming here. That SGTM!

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
jakerockland: My inclination would be to keep the file naming the same as it is now in the PR, just for consistency with the other 2 V2-Engine compatible minters, but it SGTM to update the `minterType`. 

That said, thinking about it more, I say we just remove the `minterType` altogether actually, as we don't intend for this minter to be mintersuite compatible anyway.

Screenshot 2023-01-12 at 5 46 18 PM

Given that we're going to be writing a new set of minters to be compatible with a V3-engine-plus-minter-suite world, I think backwards consistency for now in these V2-compatible minters makes sense (as we'll be archiving them all anyway by the EOQ).

WDYT ser? @ryley-o

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
jakerockland: Will do ser – will resolve this comment when updated.

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
jakerockland: Will do ser – will resolve this comment when updated.

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
jakerockland: Will do ser – will resolve this comment when updated.

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
ryley-o: yep that makes sense ser 👍 🙏

jakerockland: Rad–thank you for confirming ser. :)

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,584 @@
...
+    );
+
+    /// Auction details cleared for project `projectId`.
+    event ResetAuctionDetails(uint256 indexed projectId);
+
+    /// Maximum and minimum allowed price decay half lifes updated.
+    event AuctionHalfLifeRangeSecondsUpdated(
+        uint256 _minimumPriceDecayHalfLifeSeconds,
+        uint256 _maximumPriceDecayHalfLifeSeconds
+    );
```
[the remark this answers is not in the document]
ryley-o: As you pointed out, since this is a one-off minter for V2 core, keeping the events in the contract vs. interface may actually cause the least amount of code-bloat here. 🙏 resolving this

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
jakerockland: Done in https://github.com/ArtBlocks/artblocks-contracts/pull/388/commits/006ed711985aef8704fc1d82ced5f39f08024eb6

Have updated both this auction configuration setter, and the reset method, to both use this type of modifier.

Thank you for flagging ser! 🙏

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
jakerockland: I've revamped the implementation here to adjust the payment-splitting logic to use minter-internalized owner address/percentage in the same way that `GenArt721Minter_PBAB` does. All set now 💜 (still need to add tests, but will do so separately).

contracts/PBAB+Collabs/GenArt721MinterDAExp_PBAB.sol
jakerockland: I've updated this now :) 

(tests are still a TODO, my next step)

contracts/engine/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,611 @@
...
+        // move parameters to memory if used more than once
+        uint256 _timestampStart = uint256(_projectConfig.timestampStart);
+        uint256 _priceDecayHalfLifeSeconds = uint256(
+            _projectConfig.priceDecayHalfLifeSeconds
+        );
+        uint256 _basePrice = _projectConfig.basePrice;
+
+        require(block.timestamp > _timestampStart, "Auction not yet started");
+        require(_priceDecayHalfLifeSeconds > 0, "Only configured auctions");
+        uint256 decayedPrice = _projectConfig.startPrice;
```
ryley-o, line 537: if we think this minter might be ~commonly used as a "fixed price" minter via setting start price == end price, I would recommend a short-circuit to avoid some math (ga$) by
```solidity
// pseudo-code
if (startPrice == basePrice) {
    return basePrice;
}
// else calc price as usual
...
```

ryley-o:
overall this is looking good to me!

contracts/engine/GenArt721MinterDAExp_PBAB.sol
jakerockland: Good call, saves a lil gas and is also nice and readable (and will be covered by the updated tests noted above). Updating now.

jakerockland: Actually... it appears that this increases gas costs when we do:

```
        // Return early if configured in "fixed price mode" for gas efficiency
        if (_projectConfig.startPrice == _projectConfig.basePrice) {
            return _projectConfig.basePrice;
        }
``` 

While it is slightly more idiomatic, I think it is fine to just let this fall through as it would otherwise, but PLMK if I'm overlooking here.

Screenshot 2023-01-19 at 6 48 46 PM

jakerockland:
@ryley-o this is now ready for review – note that I'll include deployment script details in a follow-on PR.

contracts/engine/GenArt721MinterDAExp_PBAB.sol
ryley-o:
```
// Return early if configured in "fixed price mode" for gas efficiency
        if (_basePrice == _decayedPrice) {
            return _projectConfig.basePrice;
        }
```

ryley-o: (might be better names we could use 😅)

jakerockland: All set! :) Done!

contracts/engine/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,611 @@
...
+        _;
+    }
+
+    /**
+     * @notice Initializes contract to be integrated with
+     * Art Blocks core contract at address `_genArt721Address`.
+     * @param _genArt721Address Art Blocks core contract address for
+     * which this contract will be a minter.
+     */
+    constructor(address _genArt721Address) ReentrancyGuard() {
```
ryley-o, line 136: optional nit: guard preventing zero address as input parameter could be added

contracts/engine/GenArt721MinterDAExp_PBAB.sol
```diff
@@ -0,0 +1,611 @@
...
+     * @dev function is intentionally not gated to any specific access control;
+     * it only syncs a local state variable to the core contract's state.
+     */
+    function setProjectMaxInvocations(uint256 _projectId) external {
+        uint256 maxInvocations;
+        (, , , maxInvocations, , , , , ) = genArtCoreContract.projectTokenInfo(
+            _projectId
+        );
+        // update storage with results
+        projectConfig[_projectId].maxInvocations = uint24(maxInvocations);
```
ryley-o, line 177: I think we want the following for PBAB minter, since core max invocations may be increased (see `GenArt721Minter_PBAB.sol`:
```
        uint256 maxInvocations;
        uint256 invocations;
        (, , invocations, maxInvocations, , , , , ) = genArtCoreContract
            .projectTokenInfo(_projectId);
        // update storage with results
        projectMaxInvocations[_projectId] = maxInvocations;
        // update locally stored project max invocations has been reached
        projectConfig[_projectId].maxHasBeenInvoked = (invocations == maxInvocations)
```

we also gate this function to core whitelisted on the current PBAB minter, but I don't see a major reason to add that gate (since max invocations is ultimately controlled by the core, and this is just a "sync" to that value).

ryley-o (commented):
partial review

contracts/engine/GenArt721MinterDAExp_PBAB.sol
jakerockland: Updating logic here to:

```
    function setProjectMaxInvocations(uint256 _projectId) external {
        uint256 maxInvocations;
        uint256 invocations;
        (, , invocations, maxInvocations, , , , , ) = genArtCoreContract.projectTokenInfo(
            _projectId
        );
        // update storage with results
        projectConfig[_projectId].maxInvocations = uint24(maxInvocations);
        if (invocations < maxInvocations) {
            projectConfig[_projectId].maxHasBeenInvoked = false;
        }
    }
    ```

ryley-o (approved):

contracts/engine/GenArt721MinterDAExp_PBAB.sol
jakerockland: Ack. Not doing this for now as this is a dev-x ask and we don't plan to deploy this minter for long.

PR #391, opened 2022-11-17
DA w/Refund Edge Case Patch

ryley-o:
This updates the DA w/Refund branch (currently in PR form) to more smoothly handle an edge case where artist reduces max invocations to current invocations on core contract mid-auction.


ryley-o:
closing this after some discussion w/ @jakerockland - we like the existing behavior of forcing an auction to go to base price if an artist were to reduce maxInvocations mid-auction, since that is fairly suspicious behavior for an artist.

PR #395, opened 2022-11-19
Initial dependency registry

yoshiwarab:
Initial implementation of a DependencyRegistry contract to store dependency details (potentially including dependency code) for projects on GenArt721Core contracts.


contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,661 @@
...
+
+    uint8 constant AT_CHARACTER_CODE = uint8(bytes1("@")); // 0x40
+
+    /// admin ACL contract
+    IAdminACLV0 public adminACLContract;
+
+    event ProjectDependencyOverrideAdded(
+        address indexed _coreContractAddress,
+        uint256 indexed _projectId,
+        bytes32 _dependencyTypeId
```
yoshiwarab, line 36: @jakerockland @ryley-o Do you think it makes sense to always emit the dependencyType as a string instead of bytes?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,661 @@
...
+
+    function removeDependencyTypeAdditionalCDNAtIndex(
+        bytes32 _dependencyTypeId,
+        uint256 _index
+    )
+        external
+        onlyAdminACL(this.removeDependencyTypeAdditionalCDNAtIndex.selector)
+        onlyExistingDependencyType(_dependencyTypeId)
+    {
+        uint24 additionalCDNCount = dependencyTypeInfo[_dependencyTypeId]
```
yoshiwarab, line 398: Following a similar pattern to the flex contract here so that we never have a sparse "array". I don't think order should ever matter here.

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,661 @@
...
+        bytes32 _dependencyTypeId,
+        uint256 _index
+    )
+        external
+        onlyAdminACL(
+            this.removeDependencyTypeAdditionalRepositoryAtIndex.selector
+        )
+        onlyExistingDependencyType(_dependencyTypeId)
+    {
+        uint24 additionalRepositoryCount = dependencyTypeInfo[_dependencyTypeId]
```
yoshiwarab, line 476: Following a similar pattern to the flex contract here so that we never have a sparse "array". I don't think order should ever matter here.

contracts/DependencyRegistryV0.sol
jakerockland: @yoshiwarab I would say so ser–I think that would be a bit more idiomatic, WDYT @ryley-o ?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,661 @@
...
+ * Permissions managed by ACL contract
+ */
+contract DependencyRegistryV0 is Ownable {
+    using BytecodeStorage for string;
+    using BytecodeStorage for address;
+    using Bytes32Strings for bytes32;
+    using Strings for uint256;
+    using EnumerableSet for EnumerableSet.Bytes32Set;
+
+    uint8 constant AT_CHARACTER_CODE = uint8(bytes1("@")); // 0x40
```
jakerockland, line 28: Nit: could we just call this `AT_CHARACTER`?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,661 @@
...
+
+    struct DependencyType {
+        string preferredCDN;
+        mapping(uint256 => string) additionalCDNs;
+        uint24 additionalCDNCount;
+        string preferredRepository;
+        mapping(uint256 => string) additionalRepositories;
+        uint24 additionalRepositoryCount;
+        string projectWebsite;
+        uint24 scriptCount;
```
jakerockland, line 100: We should consider packing our struct here with all `uint24`s at the end.

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,661 @@
...
+    event DependencyTypeScriptUpdated(bytes32 indexed _dependencyTypeId);
+
+    struct DependencyType {
+        string preferredCDN;
+        mapping(uint256 => string) additionalCDNs;
+        uint24 additionalCDNCount;
+        string preferredRepository;
+        mapping(uint256 => string) additionalRepositories;
+        uint24 additionalRepositoryCount;
+        string projectWebsite;
```
jakerockland, line 99: Nit: can we rename this "dependencyProjectWebsite" or something to that effect?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,661 @@
...
+        uint24 scriptCount;
+        // mapping from script index to address storing script in bytecode
+        mapping(uint256 => address) scriptBytecodeAddresses;
```
jakerockland, line 102: Should we also include a `scriptBytecodeAddressCount` type variable here?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,661 @@
...
+        delete dependencyTypeInfo[_dependencyTypeId];
+
+        emit DependencyTypeRemoved(_dependencyTypeId);
+    }
+
+    /**
+     * @notice Returns a list of registered depenency types.
+     * @return List of registered depenency types.
+     */
+    function getRegisteredDependencyTypes()
```
jakerockland, line 206: I think we should probably make this a "get at index" type of method, to avoid issues with string becoming too big to return.

jakerockland (commented):
Great work @yoshiwarab 🙌 

I did an initial pass here @yoshiwarab but didn't have as much time as I would like to today – going to do an additional pass tomorrow ser and add follow on comments / thoughts 🙏 😄

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+     * @param _dependencyTypeId Dependency type to be updated.
+     * @param _preferredCDN URL for preferred CDN.
+     */
+    function updateDependencyTypePreferredCDN(
+        bytes32 _dependencyTypeId,
+        string memory _preferredCDN
+    )
+        external
+        onlyAdminACL(this.updateDependencyTypePreferredCDN.selector)
+        onlyNonEmptyString(_preferredCDN)
```
ryley-o, line 336: needs `onlyExistingDependencyType` modifier

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+     * @param _dependencyTypeId Dependency type to be updated.
+     * @param _preferredRepository URL for preferred repository.
+     */
+    function updateDependencyTypePreferredRepository(
+        bytes32 _dependencyTypeId,
+        string memory _preferredRepository
+    )
+        external
+        onlyAdminACL(this.updateDependencyTypePreferredRepository.selector)
+        onlyNonEmptyString(_preferredRepository)
```
ryley-o, line 354: needs `onlyExistingDependencyType` modifier

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+
+    /**
+     * @notice Updates project website for dependency type `_dependencyTypeId`.
+     * @param _dependencyTypeId Dependency type to be updated.
+     * @param _projectWebsite URL for project website.
+     */
+    function updateDependencyTypeProjectWebsite(
+        bytes32 _dependencyTypeId,
+        string memory _projectWebsite
+    ) external onlyAdminACL(this.updateDependencyTypeProjectWebsite.selector) {
```
ryley-o, line 372: needs `onlyExistingDependencyType` modifier.
are we okay having an empty string as input here?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+     * @param _additionalCDN CDN URL to be added. Required to be a non-empty string,
+     * but no further validation is performed.
+     */
+    function addDependencyTypeAdditionalCDN(
+        bytes32 _dependencyTypeId,
+        string memory _additionalCDN
+    )
+        external
+        onlyAdminACL(this.addDependencyTypeAdditionalCDN.selector)
+        onlyNonEmptyString(_additionalCDN)
```
ryley-o, line 393: needs `onlyExistingDependencyType` modifier

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+    {
+        uint24 additionalCDNCount = dependencyTypeInfo[_dependencyTypeId]
```
ryley-o, line 395: nit: might benefit from loading `dependencyTypeInfo[_dependencyTypeId]` to a storage pointer here

contracts/DependencyRegistryV0.sol
ryley-o: ++ to following that pattern.
nit - I think would be better & more efficient to do all math with uint256, then cast as uint24 in operation
```
--- at top of contract
using SafeCast for uint24;
--- below
dependencyTypeInfo[_dependencyTypeId]
            .additionalCDNCount = uint24(lastElementIndex);
``` 
(also, since this only decrements uint24, would be fine to not even use the SafeCast library)

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+     * @param _index Additional CDN index.
+     * @param _additionalCDN New CDN URL.
+     */
+    function updateDependencyTypeAdditionalCDNAtIndex(
+        bytes32 _dependencyTypeId,
+        uint256 _index,
+        string memory _additionalCDN
+    )
+        external
+        onlyAdminACL(this.updateDependencyTypeAdditionalCDNAtIndex.selector)
```
ryley-o, line 459: should we also have a no empty string here?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+    {
+        uint24 additionalCDNCount = dependencyTypeInfo[_dependencyTypeId]
+            .additionalCDNCount;
+        require(_index < additionalCDNCount, "Asset index out of range");
+
+        dependencyTypeInfo[_dependencyTypeId].additionalCDNs[
+            _index
+        ] = _additionalCDN;
+
```
ryley-o, line 468: I think need to increment `dependencyTypeInfo[_dependencyTypeId].additionalCDNCount`?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+     * @param _additionalRepository Repository URL to be added. Required to be a non-empty string,
+     * but no further validation is performed.
+     */
+    function addDependencyTypeAdditionalRepository(
+        bytes32 _dependencyTypeId,
+        string memory _additionalRepository
+    )
+        external
+        onlyAdminACL(this.addDependencyTypeAdditionalRepository.selector)
+        onlyNonEmptyString(_additionalRepository)
```
ryley-o, line 488: needs `onlyExistingDependencyType` modifier.

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+     */
+    function updateDependencyTypeAdditionalRepositoryAtIndex(
+        bytes32 _dependencyTypeId,
+        uint256 _index,
+        string memory _additionalRepository
+    )
+        external
+        onlyAdminACL(
+            this.updateDependencyTypeAdditionalRepositoryAtIndex.selector
+        )
```
ryley-o, line 561: maybe a no empty string modifier?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+     * @return additionalCDNCount Count of additional CDN URLs for dependency type
+     * @return preferredRepository Preferred repository URL for dependency type
+     * @return additionalRepositoryCount Count of additional repository URLs for dependency type
+     * @return projectWebsite Project website URL for dependency type
+     * @return availableOnChain Whether dependency type is available on chain
+     * @return scriptCount Count of on-chain scripts for dependency type
+     */
+    function dependencyTypeDetails(bytes32 _dependencyTypeId)
+        external
+        view
```
ryley-o, line 593: needs `onlyExistingDependencyType` modifier.

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+        returns (
+            string memory typeAndVersion,
```
ryley-o, line 595: do you think we should return this, if it is just the string version of what the user is already sending as calldata?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+     */
+    function addProjectDependencyOverride(
+        address _contractAddress,
+        uint256 _projectId,
+        bytes32 _dependencyTypeId
+    ) external onlyAdminACL(this.addProjectDependencyOverride.selector) {
+        require(
+            _dependencyTypes.contains(_dependencyTypeId),
+            "Dependency type is not registered"
+        );
```
ryley-o, line 638: Can this be replaced by `onlyExistingDependencyType` modifier?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+     * @notice Removes the script type and version override for a given 
+     * project (`projectId`) on a given core contract (`_contractAddress`).
+     * @param _contractAddress Core contract address.
+     * @param _projectId Project to remove override for.
+     */
+    function removeProjectDependencyOverride(
+        address _contractAddress,
+        uint256 _projectId
+    ) external onlyAdminACL(this.addProjectDependencyOverride.selector) {
+        delete projectDependencyOverrides[_contractAddress][_projectId];
```
ryley-o, line 660: input validation to ensure that the override exists would be good here

contracts/interfaces/0.8.x/IGenArtDependencyConsumer.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+/**
+ * @notice Interface for Art Blocks contracts that expose script details
+ */
+interface IGenArtDependencyConsumer {
```
ryley-o, line 9: Oops, probably would have been smart if these functions were included in the `IGenArt721CoreContractV3`.

To ensure future derivatives of the V3 core contract remain compatible with this registry, WDYT about just adding these functions to `IGenArt721CoreContractV3.sol`, and avoiding this new interface file altogether?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+        }
+
+        try
+            IGenArtDependencyConsumer(_contractAddress).projectScriptDetails(
+                _projectId
+            )
+        returns (string memory scriptTypeAndVersion, string memory, uint256) {
+            return scriptTypeAndVersion;
+        } catch {
+            return "";
```
ryley-o, line 695: should we potentially revert when encountering this case?

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+        uint24 additionalRepositoryCount;
+        string projectWebsite;
+        uint24 scriptCount;
+        // mapping from script index to address storing script in bytecode
+        mapping(uint256 => address) scriptBytecodeAddresses;
+    }
+
+    EnumerableSet.Bytes32Set private _dependencyTypes;
+    mapping(bytes32 => DependencyType) dependencyTypeInfo;
+    mapping(address => mapping(uint256 => bytes32)) projectDependencyOverrides;
```
ryley-o, line 107: WDYT about a dependency-registry level, admin-configurable, enumerable set of supported core contract `address`es? That would restrict us to, for example, only define overrides for our flagship core contracts. Would that be a good thing?

ryley-o (commented):
This looks great! A few comments and thoughts noted, I'll do another review pass after this round of discussions/edits 🙏

contracts/DependencyRegistryV0.sol
yoshiwarab: Good question, I could imagine a website goes down so it's no longer relevant but here's nothing new to point at

contracts/DependencyRegistryV0.sol
jakerockland: AH–yeah let's use consistency here. That is my bad :) 

> If we wanted to shove the whole "verify a single @" logic in an expanded version of the existing library being used, that might be a good pattern since we might use this across many contracts in the future.


@ryley-o sorry what library do you mean here?

contracts/DependencyRegistryV0.sol
jakerockland: Ahhh–yeah I overlooked, SY ser.

contracts/DependencyRegistryV0.sol
jakerockland: Quite like this idea a lot–this makes sense to me.

contracts/DependencyRegistryV0.sol
ryley-o: >@ryley-o sorry what library do you mean here?

@jakerockland We already use a AB-created Bytes32Strings library to call the `containsExactCharacterQty` function, so was thinking we could add a function to that library called `containsExactlyOneAtCharacter` or something. Makes the library less generalized, but if we are doing this across multiple contracts, might be helpful/cleaner.

jakerockland: Ah I see–I think that is an overly specific thing to add to a generalized `Bytes32Strings` library personally. 

Having a specific constant exist across multiple contracts I think is a bit more idiomatic, easy to follow than pushing that into an otherwise generalized library.

contracts/DependencyRegistryV0.sol
jakerockland: I like that :)

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+        external
+        onlyAdminACL(this.removeDependencyType.selector)
+    {
+        require(
+            _dependencyTypes.contains(_dependencyTypeId),
+            "Dependency type does not exist"
+        );
+
+        _dependencyTypes.remove(_dependencyTypeId);
+        delete dependencyTypeInfo[_dependencyTypeId];
```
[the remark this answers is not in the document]
jakerockland: > I think we should probably just require additionalCDNCount, additionalRepositoryCount, and scriptCount to all be zero when this function is called, which would ensure we have no stale values in the mappings.

This approach SGTM, given that removal of dependency types should be a generally infrequent occurrence and, as such, it being a somewhat manual process (multiple contract calls) seems reasonable to me.

contracts/DependencyRegistryV0.sol
jakerockland: > are we okay having an empty string as input here?

Yeah to @yoshiwarab's point I think allowing empty string probably makes sense, but worth clarifying that this is intentional in the doc comment for this method.

contracts/DependencyRegistryV0.sol
jakerockland: Yeah that makes sense to me.

contracts/DependencyRegistryV0.sol
jakerockland: In this case agree re: no-empty-string modifier.

contracts/DependencyRegistryV0.sol
jakerockland: It seems most idiomatic to include it in response to allow for response being display-able without client (e.g. imagine a "registry explorer" client app) needing to combine back together the call data and the response client-side.

contracts/DependencyRegistryV0.sol
jakerockland: ++

contracts/DependencyRegistryV0.sol
jakerockland: ++; also this would be good place to add a modifier check against valid contract address being specified.

contracts/DependencyRegistryV0.sol
jakerockland: Yeah I think reverting with an unknown-contract-project pairing type message here makes sense.

Also, having a specific admin-defined set of supported contracts would be useful from the perspective of being able to throw a more idiomatic error message in this method in the case where contract is unknown vs. where e.g. an override is not specified vs. where contract is not "new enough" (not V3+).

jakerockland (commented):
Ok finished doing my review pass from yesterday that I had started 😄  

Generally LGTM–please re-ping me for another review once you've had a chance to do a pass of updates over things here @yoshiwarab and potentially added tests (or happy to do another review pass both before and after tests are added). 

🙏 🍻

contracts/DependencyRegistryV0.sol
ryley-o: sounds good, resolving this one 🥳

contracts/DependencyRegistryV0.sol
ryley-o: nvm - this is an update, not add - resolving

contracts/DependencyRegistryV0.sol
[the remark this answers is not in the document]
yoshiwarab: Any reason to not clear things in this function vs forcing it to be done manually?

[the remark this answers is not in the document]
ryley-o: I think clearing here would be fine, since we wouldn't expect to hit the gas limit. Since the operation is likely rare, keeping on-chain code simple would be the only reason to not do that in this function. A judgement call for sure!

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,751 @@
...
+        DependencyType storage dependencyType = dependencyTypeInfo[
+            _dependencyTypeId
+        ];
+        // store script in contract bytecode
+        dependencyType.scriptBytecodeAddresses[
+            dependencyType.scriptCount
+        ] = _script.writeToBytecode();
+        dependencyType.scriptCount = dependencyType.scriptCount + 1;
+
+        emit DependencyTypeScriptUpdated(_dependencyTypeId);
```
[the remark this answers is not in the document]
yoshiwarab: Was planning on handling this similar to project scripts currently and just refresh the whole script whenever anything gets updated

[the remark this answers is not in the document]
ryley-o: that sounds good, resolving 🎉

contracts/DependencyRegistryV0.sol
yoshiwarab: Okay, going to leave as is for simplicity plus the ability to set _dependencyTypeId as an indexed event parameter.

contracts/interfaces/0.8.x/IGenArtDependencyConsumer.sol
yoshiwarab: Sounds good!

yoshiwarab:
@jakerockland this is ready for another round of review whenever you get the chance!

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,919 @@
...
+import "@openzeppelin-4.5/contracts/utils/math/SafeCast.sol";
+
+import "./libs/0.8.x/BytecodeStorage.sol";
+import "./libs/0.8.x/Bytes32Strings.sol";
+
+/**
+ * @title Art Blocks Dependency Registry, V0.
+ * @author Art Blocks Inc.
+ * @notice Privileged Roles and Ownership:
+ * Permissions managed by ACL contract
```
jakerockland, line 23: @yoshiwarab Given that we do not typically use the proxy upgradable pattern at Art Blocks, I think it is worth expanding on the "why" aspect of why we've implemented this contract to allow for upgrades, based on our prior internal conversation on the topic.

jakerockland: In doing so–this is also probably a place to call out that if/when we ever do call `renounceOwnership` (which we could do if no longer maintaining the registry, and instead intending to "freeze" it at a fixed period in time) that this becomes a frozen immutable registry, as no upgrades can be performed.

jakerockland (approved):
This is fantastic @yoshiwarab – really excited with how much progress we've made against this; we are further along than I had realized and excited to be in a place of being ready to ship this quite soon!! 😱 💯 💜

contracts/DependencyRegistryV0.sol
jakerockland: Another related thought: we may want to consider testing an upgrade on testnet before deploying to mainnet.

jakerockland (approved):

contracts/interfaces/0.8.x/IDependencyRegistryV0.sol
```diff
@@ -0,0 +1,66 @@
+pragma solidity ^0.8.17;
+
+interface IDependencyRegistryV0 {
```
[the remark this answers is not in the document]
yoshiwarab: SGTM!

contracts/interfaces/0.8.x/IGenArt721CoreContractV3.sol
```diff
@@ -58,6 +58,12 @@ interface IGenArt721CoreContractV3 is IManifold {
     // backwards-compatible (pre-V3) admin - equal to owner()
     function admin() external view returns (address);
 
+    /// Dependency registry managed by Art Blocks
```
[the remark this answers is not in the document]
yoshiwarab: I actually did have this as a separate interface before but @ryley-o suggested adding it here instead. That makes sense to me though.

@ryley-o please let me know if you feel strongly about having it here instead!

contracts/interfaces/0.8.x/IDependencyRegistryV0.sol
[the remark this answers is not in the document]
yoshiwarab: @jakerockland I added only what will be used in the subgraph + what I felt was most core to the functionality but please let me know if you think I should add anything else!

contracts/DependencyRegistryV0.sol
yoshiwarab: @jakerockland, updated the comment here but can't remember exactly what we discussed internally about upgradeability so left it pretty generic. Let me know if you'd like me to add anything else1

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,926 @@
...
+     * @dev ref: https://docs.openzeppelin.com/contracts/4.x/api/access#Ownable
+     * @dev owner role was called `admin` prior to V3 core contract
+     */
+    function owner()
+        public
+        view
+        override(OwnableUpgradeable)
+        returns (address)
+    {
+        if (address(adminACLContract) == address(0)) {
```
yoshiwarab, line 907: @jakerockland let me know if this behavior seems right to you

contracts/interfaces/0.8.x/IGenArt721CoreContractV3.sol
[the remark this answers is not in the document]
jakerockland: Ah interesting – do you know what more of the context was around not having it as an interface? From my perspective seems useful given that we can readily use this across not just our flagship core contracts but the Engine ecosystem more broadly as well.

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,919 @@
...
+     * @dev ref: https://docs.openzeppelin.com/contracts/4.x/api/access#Ownable
+     * @dev owner role was called `admin` prior to V3 core contract
+     */
+    function owner()
+        public
+        view
+        override(OwnableUpgradeable)
+        returns (address)
+    {
+        return OwnableUpgradeable.owner();
```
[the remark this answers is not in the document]
jakerockland: We chatted offline and decided to remove this suggestion.

contracts/interfaces/0.8.x/IDependencyRegistryV0.sol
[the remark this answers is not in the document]
jakerockland: Roger that–makes sense to me, let's double click on this in our 1:1 sync and then can close out.

contracts/DependencyRegistryV0.sol
```diff
@@ -0,0 +1,923 @@
...
+/**
+ * @title Art Blocks Dependency Registry, V0.
+ * @author Art Blocks Inc.
+ * @notice Privileged Roles and Ownership:
+ * Permissions managed by ACL contract. If/when we ever call
+ * renounceOwnership() this will becom a frozen, immutable registry
+ * as no upgrades will be possible.
+ * This contract has been made upgradeable as we expect
+ * its required functionality in relation to the Art Blocks ecosystem
+ * to evolve over time.
```
jakerockland, line 28: Would add onto this sentence at the beginning: "This contract is intended to be an auxiliary reference registry to our non-upgradeable and immutable ERC-721 conforming core contracts, and has been made upgradeable..."

contracts/DependencyRegistryV0.sol
jakerockland: Added one suggestion here but otherwise LGTM! https://github.com/ArtBlocks/artblocks-contracts/pull/395/files#r1054630021

Thank you ser.

contracts/interfaces/0.8.x/IDependencyRegistryV0.sol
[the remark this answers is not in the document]
jakerockland: Chatted in our sync, we should just trim this down to what is needed by the subgraph.

jakerockland:
> Another related thought: we may want to consider testing an upgrade on testnet before deploying to mainnet.

Regarding this though–I don't think we need to block this PR on this, just something we should do pre-mainnet.

jakerockland (approved):

PR #397, opened 2022-11-29
Add CI Coverage

### LINKED ISSUE #245: Add CI for coverage reports
Per solidity-coverage docs [here](https://github.com/sc-forks/solidity-coverage/blob/master/docs/faq.md#continuous-integration), recommend using Coveralls for this.

It is free for open-source repos.

ryley-o:
Adds CI coverage via CircleCI + Coveralls integrations.

Closes #245 

Per solidity-coverage docs [here](https://github.com/sc-forks/solidity-coverage/blob/master/docs/faq.md#continuous-integration), Coveralls is the preferred coverage service (over something like codecov) due to its branch coverage reporting capabilities (which we use).

Interestingly, Coveralls actually is a bit more strict in branch coverage than our default `solidity-coverage` library reporting (which was used by us and our auditors for the V3 core contract). We may want some follow-on work to reach 100% coverage on the new Coveralls reports, mostly dealing with branch coverage on function modifiers.

Using a parallelized coverage pattern to speed up coverage. Bumped up solidity-coverage version to help prevent running out of memory during compiling contracts step on CircleCi's largest free-tier `large` instances (8gb memory), although updating the .solcover.js skipped directories to skip old minters and unused PBAB solidity files greatly improved that detail.


ryley-o:
here is a current report - should probably do some tweaking to only report non-PBAB contracts, and the branch coverage looks a little odd... 
https://coveralls.io/builds/54646910

lindsgil:
> 

In the report what does this mean under "Run Details" `1740 of 4100 branches covered (42.44%)`?

.circleci/config.yml
```diff
@@ -36,21 +36,77 @@ jobs:
...
+          name: coverage
+          command: |
+            TEST_FILES="{$(circleci tests glob "./test/**/*.ts" | circleci tests split | xargs | sed -e 's/ /,/g')}"
+            yarn coverage --testfiles "${TEST_FILES}"
       - run:
-          name: Success
-          command: echo "Success"
+          name: save coverage
+          command: |
+            mkdir -p /tmp/coverage
```
lindsgil, line 59: is there any cleanup we should do with this tmp dir?

ryley-o:
>In the report what does this mean under "Run Details" 1740 of 4100 branches covered (42.44%)?

@lindsgil Currently, the percentage is quite low due to a bunch of redundant PBAB files being included in the coverage report. For branch coverage, if you have an `if (x) {y} else {z}` pattern, 100% branch coverage would be ensuring `x` evaluated as true and false in test suite. 50% coverage would be only evaluating `x` as either true or false, and 0% would be missing the statement altogether.

It does appear that coveralls picks up some branches on our solidity modifiers that our existing reporting doesn't pick up, so looking into that a bit. I expected to see 100% branch coverage on the V3 core contract (which is what us and our auditors measured), but coveralls highlights a few missed branches.

.circleci/config.yml
ryley-o: I'm not 100% sure, but I think circleci should clean up everything (including the persisted workspace) after the entire workflow is completed. https://circleci.com/docs/workspaces/

I'm ~copying the pattern in this repo: https://github.com/SetProtocol/set-v2-strategies/blob/29db4a3de4a6ecd55a2c86d85a002b7e1106f8b2/.circleci/config.yml#L79-L143

jakerockland (approved):

.circleci/config.yml
jakerockland: Yeah that seems correct to me – also consistent w/ this (from those docs):

> The workspace is additive only, jobs may add files to the workspace but cannot delete files from the workspace.

PR #398, opened 2022-11-29
Update to DA w/Settlement Nomenclature

ryley-o:
Update to a more idiomatic DA w/Settlement nomenclature instead of DA w/Refund.

Because our new minter is truly acting like a settlement layer vs. a payment receiver + refunder, updating smart contract naming and documentation to more appropriately reflect that behavior.

This is intended to be a discrete no-op that may be merged into #382 prior to merging into main.


jakerockland:
This is fantastic ser – TYVM for the quick update here 🙏

jakerockland (approved):

PR #399, opened 2022-11-30
Merklev2 Type to Merklev3

mchrupcala:


ryley-o (approved):
LGTM - I'm going to tag you in a follow-on PR, as this exposed a slight hole in our tests (although, this would have never made it to mainnet as we immediately saw)

PR #405, opened 2022-12-13
Add interfaces for shared events for existing minters to simplify subgraph logic

yoshiwarab:
This updates/adds interfaces for all of the existing IFilteredMinters that have shared events. This should enable us to implement a single handler for common events rather than multiple events that then call a generic handler.


jakerockland (approved):
This is awesome–thank you ser. LGTM.

ryley-o:
Thank you very much for making these updates @yoshiwarab! 🙏 🙏

PR #409, opened 2022-12-15
Minters: Adding ability to manually set max project invocations

lyaunzbe:
We want to add functionality that allows minters to manually set the project max invocations (in the current state of the world, minters can only sync the project max invocations from the core contract). This primarily allows us to decrease/increase project max invocations as we see fit, as long as it remains below or equal to the value of the project max invocations that is set on the core contract.

New minters added:

- MinterSetPriceERC20V3
- MinterSetPriceV3
- MinterMerkleV4
- MinterHolderV3
- MinterDALinV3
- MinterDAExpV3

Note: MinterDAExpSettlement was not bumped as it does not use/depend on max invocations at all


contracts/minter-suite/Minters/MinterSetPrice/MinterSetPriceV3.sol
```diff
@@ -0,0 +1,392 @@
...
+
+    /**
+     * @notice Manually sets the local maximum invocations of project `_projectId`
+     * with the provided `_maxInvocations`, checking that `_maxInvocations` is less
+     * than or equal to the value of project `_project_id`'s maximum invocations that is
+     * set onthe core contract.
+     * @param _projectId Project ID to set the maximum invocations for.
+     * @param _maxInvocations Maximum invocations to set for the project.
+     */
+    function manuallySetProjectMaxInvocations(
```
lyaunzbe, line 115: flagging new function location

contracts/minter-suite/Minters/MinterSetPrice/MinterSetPriceV3.sol
```diff
@@ -0,0 +1,392 @@
...
+    function setProjectMaxInvocations(uint256 _projectId) external {
+        uint256 maxInvocations;
+        (, maxInvocations, , , , ) = genArtCoreContract.projectStateData(
+            _projectId
+        );
+        // update storage with results
+        projectConfig[_projectId].maxInvocations = uint24(maxInvocations);
+    }
+
+    /**
```
lyaunzbe, line 107: flagging new function location below

lyaunzbe:
Had some funkiness with getting gas calc tests passing(unrelated to my changes/additions), will be squashing this PR before it goes in 😅

lyaunzbe:
leaving note to self: after async discussion with @jakerockland, we want new minter interfaces to inherit from new `IFilteredMinterV2` vs re-implementing IFilteredMinterV2 business logic in each one

jakerockland:
@lyaunzbe is this ready for re-review?

lyaunzbe:
@jakerockland yessir, ready for re-review 🖖

test/minter-suite-minters/MinterMerkle/MinterMerkleV4.test.ts
```diff
@@ -0,0 +1,725 @@
...
+    //     );
+    //     const receipt = await tx.deployTransaction.wait();
+    //     // target event is the last log
+    //     const targetLog = receipt.logs[0];
+    //     // expect "PlatformUpdated" event as log 0
+    //     await expect(targetLog.topics[0]).to.be.equal(
+    //       ethers.utils.keccak256(
+    //         ethers.utils.toUtf8Bytes("DelegationRegistryUpdated(address)")
+    //       )
+    //     );
```
jakerockland, line 214: @lyaunzbe I'm confused why we have this long set of commented out tests – also, I do wonder if we should just be updating our testing suite to only test the latest of each minter version, rather than having additional tests for each version.. WDYT here @ryley-o ?

jakerockland (approved):
The contract changes themselves here LGTM @lyaunzbe but would like to double-click on our testing strategy here – cc @ryley-o

test/minter-suite-minters/MinterMerkle/MinterMerkleV4.test.ts
lyaunzbe: ah thank you for catching this @jakerockland this was a relic of me debugging something (gas values from common tests were causing failures in circle), but this should be uncommented!

regarding your suggestion for overall testing strategy, I don't have too strong of an opinion here. Would agree that only  testing latest minter versions would be streamlined/nice dev ex generally, but also in the context of smart contract dev specifically and the fact that earlier versions will continue to be used in the wild (something we cant really control), it may be nice to have those older tests for reference more than anything. curious to hear more of your thoughts as well @ryley-o

ryley-o: I've added an issue regarding archiving tests associated with deprecated minters here: https://github.com/ArtBlocks/artblocks-contracts/issues/420

For new minters, I think utilizing the common tests to achieve 100% coverage is good practice. Hopefully the common tests make the lift relatively minimal when small additional functionality is added in PRs like these. Happy to revisit if we think it is overkill for this kind of PR though!

contracts/interfaces/0.8.x/IFilteredMinterV2.sol
```diff
@@ -0,0 +1,30 @@
...
+ * @title This interface extends the IFilteredMinterV1 interface in order to
+ * add support for manually setting project max invocations.
+ * @author Art Blocks Inc.
+ */
+interface IFilteredMinterV2 is IFilteredMinterV1 {
+    /**
+     * @notice Local max invocations for project `_projectId`, tied to core contract `_coreContractAddress`,
+     * updated to `_maxInvocations`.
+     */
+    event ManuallySetProjectMaxInvocations(
```
ryley-o, line 18: WDYT of calling this "ProjectMaxInvocationsManuallyLimited"?

contracts/interfaces/0.8.x/IFilteredMinterV2.sol
```diff
@@ -0,0 +1,30 @@
...
+        uint256 indexed _projectId,
+        address indexed _coreContractAddress,
```
ryley-o, line 20: In our current minter suite, we heavily rely on each minter being tied to a single core contract address. It is a relationship that we will likely remove in some future iteration, but for now, I think we should continue to use that assumption, and this address parameter could be removed from this event.

contracts/interfaces/0.8.x/IFilteredMinterV2.sol
```diff
@@ -0,0 +1,30 @@
...
+        uint256 _maxInvocations
+    );
+
+    // Sets the local max invocations for a given project, checking that the provided max invocations is
+    // less than or equal to the global max invocations for the project set on the core contract.
+    function manuallySetProjectMaxInvocations(
```
ryley-o, line 26: WDYT about calling this `manuallyLimitProjectMaxInvocations`?

contracts/interfaces/0.8.x/IFilteredMinterV2.sol
```diff
@@ -0,0 +1,30 @@
...
+     * updated to `_maxInvocations`.
+     */
+    event ManuallySetProjectMaxInvocations(
+        uint256 indexed _projectId,
+        address indexed _coreContractAddress,
+        uint256 _maxInvocations
+    );
+
+    // Sets the local max invocations for a given project, checking that the provided max invocations is
+    // less than or equal to the global max invocations for the project set on the core contract.
```
ryley-o, line 25: I think we should add something along the lines saying, "This does not affect the value defined on the core contract, and only limits mints occurring on this minter".

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,613 @@
...
+     * @notice Syncs local maximum invocations of project `_projectId` based on
+     * the value currently defined in the core contract. Only used for gas
+     * optimization of mints after maxInvocations has been reached.
+     * @param _projectId Project ID to set the maximum invocations for.
+     * @dev this enables gas reduction after maxInvocations have been reached -
+     * core contracts shall still enforce a maxInvocation check during mint.
+     * @dev function is intentionally not gated to any specific access control;
+     * it only syncs a local state variable to the core contract's state.
+     */
+    function setProjectMaxInvocations(uint256 _projectId) external {
```
ryley-o, line 142: In EVERY minter, this needs to have an added `onlyArtist` modifier, since it modifies the same value we now allow to manually be reduced.
per comment in this function's description above:
```
* @dev function is intentionally not gated to any specific access control;
* it only syncs a local state variable to the core contract's state.
```
(this function's description should be updated to reflect that it is more than a simple gas-optimization, and has interplay with the `manuallySetProjectMaxInvocations` function.

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,613 @@
...
+
+    /**
+     * @notice Manually sets the local maximum invocations of project `_projectId`
+     * with the provided `_maxInvocations`, checking that `_maxInvocations` is less
+     * than or equal to the value of project `_project_id`'s maximum invocations that is
+     * set on the core contract.
+     * @param _projectId Project ID to set the maximum invocations for.
+     * @param _maxInvocations Maximum invocations to set for the project.
+     */
+    function manuallySetProjectMaxInvocations(
```
ryley-o, line 219: for all minters, I think this makes more sense to be immediately after the `setProjectMaxInvocations` function

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,613 @@
...
+            _projectId
+        );
+        require(
+            _maxInvocations <= maxInvocations,
+            "Cannot increase project max invocations above core contract set project max invocations"
+        );
+
+        // EFFECTS
+        // update storage with results
+        projectConfig[_projectId].maxInvocations = uint24(_maxInvocations);
```
ryley-o, line 236: For all minters, we also need to update `projectConfig[_projectId].maxHasBeenInvoked`, since we could be increasing _maxInvocations now. I'm not completely sure if this will affect any downstream logic in subgraph, hasura, or frontend, but I think it is generally okay... We will need to be careful in DA w/Settlement (not included in this PR at the moment)

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,613 @@
...
+     * @dev function is intentionally not gated to any specific access control;
+     * it only syncs a local state variable to the core contract's state.
+     */
+    function setProjectMaxInvocations(uint256 _projectId) external {
+        uint256 maxInvocations;
+        (, maxInvocations, , , , ) = genArtCoreContract.projectStateData(
+            _projectId
+        );
+        // update storage with results
+        projectConfig[_projectId].maxInvocations = uint24(maxInvocations);
```
ryley-o, line 148: For all minters, we also need to update `projectConfig[_projectId].maxHasBeenInvoked`, since we could be increasing _maxInvocations now. I'm not completely sure if this will affect any downstream logic in subgraph, hasura, or frontend, but I think it is generally okay... We will need to be careful in DA w/Settlement (not included in this PR at the moment)

ryley-o (changes requested):
A few thoughts on some of the logic surrounding when/how max invocations is set, and re-updating maxHasBeenInvoked 🙏

contracts/interfaces/0.8.x/IFilteredMinterV2.sol
lyaunzbe: agreed. i think the inline comments above the actual implementation of the functions do a good job of describing this, so will probably just copy over most of that content to the interface inline docs

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
lyaunzbe: makes sense (we don't want to allow anyone to arbitrarily override the local state that an artist can now manually set). will update inline docs to reflect this as well

jakerockland:
cc @shantanubala for vis.

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,639 @@
...
+        onlyArtist(_projectId)
+    {
+        uint256 maxInvocations;
+        uint256 invocations;
+        (invocations, maxInvocations, , , , ) = genArtCoreContract
+            .projectStateData(_projectId);
+
+        // update storage with results
+        projectConfig[_projectId].maxInvocations = uint24(maxInvocations);
+
```
lyaunzbe, line 153: @ryley-o please review this section below that updates `maxHasBeenInvoked` accordingly 🙏 much appreciated!

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,639 @@
...
+            .projectStateData(_projectId);
+        require(
+            _maxInvocations <= maxInvocations,
+            "Cannot increase project max invocations above core contract set project max invocations"
+        );
+
+        // EFFECTS
+        // update storage with results
+        projectConfig[_projectId].maxInvocations = uint24(_maxInvocations);
+        // We need to ensure maxHasBeenInvoked is correctly set after manually setting the
```
lyaunzbe, line 197: @ryley-o please review this section below that updates `maxHasBeenInvoked` accordingly 🙏 much appreciated!

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
lyaunzbe: updated for review

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
lyaunzbe: updated for review

test/minter-suite-minters/DA/MinterDAExp/MinterDAExpV3.test.ts
```diff
@@ -0,0 +1,307 @@
...
+      });
+    });
+
+    describe("setProjectMaxInvocations", async function () {
+      it("allows artist to call setProjectMaxInvocations", async function () {
+        await this.minter
+          .connect(this.accounts.artist)
+          .setProjectMaxInvocations(this.projectZero);
+      });
+
```
lyaunzbe, line 151: these two tests below are testing the logic around setting/reseting maxHasBeenInvoked accordingly when either setMaxProjectInvocations or manuallyLimitProjectMaxInvocations is called

test/minter-suite-minters/DA/MinterDAExp/MinterDAExpV3.test.ts
```diff
@@ -0,0 +1,307 @@
...
+          .connect(this.accounts.artist)
+          .manuallyLimitProjectMaxInvocations(this.projectZero, 1);
+
+        // expect maxInvocations on the minter to be 1
+        const localMaxInvocations3 = await this.minter
+          .connect(this.accounts.artist)
+          .projectConfig(this.projectZero);
+        expect(localMaxInvocations3.maxInvocations).to.equal(1);
+
+        // expect projectMaxHasBeenInvoked to now be false
```
ryley-o, line 250:
```suggestion
        // expect projectMaxHasBeenInvoked to now be true
```

ryley-o (commented):
Thanks @lyaunzbe - Made a couple simplification suggestions, and maybe I can take one last peek after everything has been applied to all the new minters (just to get another set of eyes on copy and pastes)?

test/minter-suite-minters/DA/MinterDAExp/MinterDAExpV3.test.ts
ryley-o: LGTM!

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,639 @@
...
+        // ensure that the manually set maxInvocations is not greater than what is set on the core contract
+        uint256 maxInvocations;
+        uint256 invocations;
+        (invocations, maxInvocations, , , , ) = genArtCoreContract
+            .projectStateData(_projectId);
+        require(
+            _maxInvocations <= maxInvocations,
+            "Cannot increase project max invocations above core contract set project max invocations"
+        );
+
```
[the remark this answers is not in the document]
lyaunzbe: yep, agreed. gonna update the tests to check this revert

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,639 @@
...
+        projectConfig[_projectId].maxInvocations = uint24(_maxInvocations);
+        // We need to ensure maxHasBeenInvoked is correctly set after manually setting the
+        // local maxInvocations value.
+        if (
+            invocations % ONE_MILLION < projectConfig[_projectId].maxInvocations
+        ) {
+            projectConfig[_projectId].maxHasBeenInvoked = false;
+        } else {
+            projectConfig[_projectId].maxHasBeenInvoked = true;
+        }
```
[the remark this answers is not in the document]
lyaunzbe: 👍 👍 much cleaner

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,639 @@
...
+        // equal to the previous value of maxInvocations stored locally.
+        unchecked {
+            // okay if this underflows because if statement will always eval false.
+            if (
+                invocations % ONE_MILLION !=
+                projectConfig[_projectId].maxInvocations - 1
+            ) {
+                projectConfig[_projectId].maxHasBeenInvoked = false;
+            }
+        }
```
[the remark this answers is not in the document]
lyaunzbe: lgtm but id like to leave some of the comments so will recommit this one myslef

lyaunzbe:
@ryley-o sgtm, yes would definitely appreciate a set of eyes to make sure nothing is missed. will ping for re-review once all minters/tests are in

justtry1512:
effective implementation.

test/minter-suite-minters/DA/MinterDAExp/MinterDAExpV3.test.ts
```diff
@@ -0,0 +1,306 @@
...
+  getAccounts,
+  assignDefaultConstants,
+  deployAndGet,
+  deployCoreWithMinterFilter,
+  safeAddProject,
+} from "../../../util/common";
+import { ONE_MINUTE, ONE_HOUR, ONE_DAY } from "../../../util/constants";
+import { MinterDAExp_Common } from "./MinterDAExp.common";
+import { MinterDAV1V2_Common } from "../MinterDAV1V2.common";
+
```
ryley-o, line 24: Please add the following to all minter tests to prevent nuisance logs about event overloading:
```js
// hide nuisance logs about event overloading
Logger.setLogLevel(Logger.levels.ERROR);
```

test/minter-suite-minters/DA/MinterDAExp/MinterDAExpV3.test.ts
```diff
@@ -0,0 +1,306 @@
...
+    });
+
+    describe("common DAEXP tests", async function () {
+      await MinterDAExp_Common();
+    });
+
+    describe("common DA V1V2 tests", async function () {
+      await MinterDAV1V2_Common();
+    });
+
```
ryley-o, line 121: We do want the same tests that were ran on V2 to be ran here.
Perhaps we refactor this in the future, but for now, I would probably rename `MinterDAV2_Common()` to `MinterDAV2V3_Common()`, then add the imports and run here, as well as update the names in `MinterDAExpV2.test.ts`.

```js
    describe("common DA V2 tests", async function () {
      await MinterDAV2_Common();
    });
```
This same pattern also should be updated in `MinterDALinV3.test.ts`, `MinterSetPriceV3.test.ts`, and `MinterSetPriceERC20V3.ts` 👍

test/minter-suite-minters/DA/MinterDAExp/MinterDAExpV3.test.ts
```diff
@@ -0,0 +1,306 @@
...
+        );
+      await ethers.provider.send("evm_mine", [this.startTime]);
+    });
+
+    describe("common DAEXP tests", async function () {
+      await MinterDAExp_Common();
+    });
+
+    describe("common DA V1V2 tests", async function () {
+      await MinterDAV1V2_Common();
```
ryley-o, line 119: could potentially update this to be named `MinterDAV1V2V3_Common()`... oof this should get better after the archiving of old tests in #444 😓

ryley-o (approved):
Generally LGTM with a few follow-on discussions below.
Happy to take a final look if want after these are resolved 👍

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,618 @@
...
+
+        // update storage with results
+        projectConfig[_projectId].maxInvocations = uint24(maxInvocations);
+
+        // We need to ensure maxHasBeenInvoked is correctly set after manually syncing the
+        // local maxInvocations value with the core contract's maxInvocations value.
+        // This synced value of maxInvocations from the core contract will always be greater
+        // than or equal to the previous value of maxInvocations stored locally.
+        projectConfig[_projectId].maxHasBeenInvoked =
+            invocations == maxInvocations;
```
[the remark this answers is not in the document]
lyaunzbe: good call + yes i like reusing the same event but generalizing the name like that 👍

contracts/minter-suite/Minters/MinterDAExp/MinterDAExpV3.sol
```diff
@@ -0,0 +1,618 @@
...
+     * set on the core contract.
+     * @param _projectId Project ID to set the maximum invocations for.
+     * @param _maxInvocations Maximum invocations to set for the project.
+     */
+    function manuallyLimitProjectMaxInvocations(
+        uint256 _projectId,
+        uint256 _maxInvocations
+    ) external onlyArtist(_projectId) {
+        // CHECKS
+        // ensure that the manually set maxInvocations is not greater than what is set on the core contract
```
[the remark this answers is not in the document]
ryley-o: as discussed in DMs, require statement isn't needed because the only time this could be set to zero is when invocation count was zero, at which point `maxHasBeenInvoked` would be set to true, preventing all minting.

Might be worth adding an in-line comment here noting why that case is okay. Feel free to resolve ser! 🙏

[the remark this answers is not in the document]
lyaunzbe: after discussion in slack with @ryley-o, the only time someone could set this to 0 is if the invocation count was 0
and then when that would be set it would also set masHasBeenInvoked to true, short circuiting the minter's purchase function before reaching core contract. will add an in-line comment to make this case clear

test/minter-suite-minters/SetPrice/MinterSetPrice/MinterSetPriceV2.test.ts
```diff
@@ -19,8 +19,8 @@ import {
 } from "../../../util/common";
 
 import { MinterSetPrice_ETH_Common } from "./MinterSetPrice.common";
-import { MinterSetPriceV1V2_Common } from "../MinterSetPriceV1V2.common";
-import { MinterSetPriceV2_Common } from "../MinterSetPriceV2.common";
+import { MinterSetPriceV1V2V3_Common } from "../MinterSetPriceV1V2V3.common";
+import { MinterSetPriceV2V3_Common } from "../MinterSetPriceV2V3.common";
```
ryley-o, line 23: I think this import is missing, which, tbh, I'm really surprised isn't causing this test to fail.
IIUC, the fix here would be to update the export name from `MinterSetPriceV2_Common` to `MinterSetPriceV2V3_Common` in `"../MinterSetPriceV2V3.common"`

ryley-o (approved):
LGTM - amazing work here ser 🙏

test/minter-suite-minters/SetPrice/MinterSetPrice/MinterSetPriceV2.test.ts
0xQTpie: ah geez, literally had this in my editor updated but just forgot to hit save on the file before pushing 😅 but ya super weird it didn't actually cause a test to fail on CI

PR #410, opened 2022-12-16
Implement a "Polyptych" minter that allows a token to inherit its hash from a previously minted token

shantanubala:
# Summary of changes

This PR implements a "polyptych" minter that enables the following user experience:

1. User visits the Art Blocks collaborations site and mints the first panel of the project using the standard Art Blocks dutch auction minter
2. Their minted token will contain metadata containing information about additional panels of the art that can be minted via the features provided by the artist features script
3. Web site displays a user's eligibility to mint additional token based on their initial minted token(s)
4. User or delegate demonstrates proof of ownership of their initial token from step 1 and mints a second piece that inherits the same token hash as the first token

## Architecture

This is a quick overview of the proposed modifications:

ROUGH  Token-copier contract architecture


contracts/PolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,77 @@
...
+    function setPolyptychMinterAddress(address _polyptychMinterAddress)
+        external
+    {
+        require(msg.sender == genArt721Core.owner(), "Only admin may call");
+        polyptychMinterAddress = _polyptychMinterAddress;
+    }
+
+    // Allows the owner of the core contract to configure a project as a polyptych
+    function toggleProjectIsPolyptych(uint256 _projectId) external {
+        require(msg.sender == genArt721Core.owner(), "Only admin may call");
```
shantanubala, line 40: Checks on ownership will fail because the ACL contract owns the core contract.

shantanubala: Also, this toggle should have an event emission that emits the boolean value as a best practice.

contracts/PolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,77 @@
...
+
+pragma solidity 0.8.17;
+
+import "./interfaces/0.8.x/IRandomizerV2.sol";
+import "./interfaces/0.8.x/IGenArt721CoreContractV3.sol";
+import "./interfaces/0.8.x/IFilteredMinterPolyptychV0.sol";
+
+import "@openzeppelin-4.7/contracts/access/Ownable.sol";
+
+contract PolyptychRandomizerV0 is IRandomizerV2, Ownable {
```
shantanubala, line 12: Rename this `BasicPolyptychRandomizerV0`

contracts/PolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,77 @@
...
+        return _projectIsPolyptych[_projectId];
+    }
+
+    // When `genArt721Core` calls this, it can be assured that the randomizer
+    // will set a bytes32 hash for tokenId `_tokenId` on the core contract.
+    function assignTokenHash(uint256 _tokenId) external virtual {
+        require(msg.sender == address(genArt721Core), "Only core may call");
+
+        if (_projectIsPolyptych[_tokenId / ONE_MILLION]) {
+            bytes32 seededHash = IFilteredMinterPolyptychV0(
```
shantanubala, line 59: Require that this seeded hash is not zero

contracts/PolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,77 @@
...
+                polyptychMinterAddress
+            ).polyptychSeedForToken(_tokenId);
```
shantanubala, line 61: Ensure this function call fails if the method does not exist on the minter contract.

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,690 @@
...
+        // we need the new token ID in advance of the randomizer setting a token hash
+        (uint256 _invocations, , , , , ) = genArtCoreContract.projectStateData(
+            _projectId
+        );
+        uint256 _newTokenId = (_projectId * ONE_MILLION) + _invocations;
+
+        // EFFECTS
+
+        // we need to store the new token ID before it is minted so the randomizer can query it
+        polyptychSeedHashes[_newTokenId] = genArtCoreContract.tokenIdToHash(
```
shantanubala, line 546: Add validation to ensure that a valid token hash is being used.

shantanubala: Move this data mapping to the randomizer - make the minter call the randomizer instead of requiring the randomizer to call the minter.

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,704 @@
...
+        genArtCoreContractWithRandomizer.randomizerContract()
+            .setPolyptychHashSeed(_newTokenId, _hashSeed);
+
+        // once mint() is called, the polyptych randomizer will either:
+        // 1) assign a random token hash
+        // 2) if configured, obtain the token hash from the `polyptychSeedHashes` mapping
+        tokenId = minterFilter.mint(_to, _projectId, vault);
+
+        // NOTE: delegate-vault handling **ends here**.
+
```
ryley-o, line 569: As configured in our default _Engine contracts, we don't have any code in our `_beforeTokenTransfer` or `_afterTokenTransfer` ERC721_PackedHashSeed hooks. However, I could imagine a partner potentially wanting to add something to those at some point. If they were to add something to those, a reentrancy bug could be opened that could potentially allow a user to set an arbitrary token hash on the randomizer.

I would recommend we add code here that requires that the hash seed is as expected on the core contract:
```
        // redundant check against reentrancy
        _assignedHashSeed = genArtCoreContractWithRandomizer.tokenIdToHashSeed(
            tokenId
        );
        require(_assignedHashSeed == _hashSeed);
```

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,704 @@
...
+        (uint256 _invocations, , , , , ) = genArtCoreContract.projectStateData(
+            _projectId
+        );
+        uint256 _newTokenId = (_projectId * ONE_MILLION) + _invocations;
+
+        // EFFECTS
+
+        // we need to store the new token ID before it is minted so the randomizer can query it
+        IGenArt721CoreContractV3WithRandomizer genArtCoreContractWithRandomizer = IGenArt721CoreContractV3WithRandomizer(genArt721CoreAddress);
+        bytes12 _hashSeed = genArtCoreContractWithRandomizer.tokenIdToHashSeed(
```
ryley-o, line 555: nit: WDYT about naming this `targetHashSeed`?

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,704 @@
...
+        // 1) assign a random token hash
+        // 2) if configured, obtain the token hash from the `polyptychSeedHashes` mapping
+        tokenId = minterFilter.mint(_to, _projectId, vault);
+
+        // NOTE: delegate-vault handling **ends here**.
+
+        // okay if this underflows because if statement will always eval false.
+        // this is only for gas optimization (core enforces maxInvocations).
+        unchecked {
+            if (tokenId % ONE_MILLION == _projectConfig.maxInvocations - 1) {
```
ryley-o, line 573: nit: could use `_invocations == _projectConfig.maxInvocations` here instead for slight gas savings

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,704 @@
...
+     * Allows holders of project `_ownedNFTProjectIds[0]` on token
+     * contract to mint `_projectId`.
+     * @param _projectId Project ID to enable minting on.
+     * @param _ownedNFTProjectIds Project IDs whose
+     * holders shall be allowlisted to mint project `_projectId`.
+     */
+    function allowHoldersOfProjects(
+        uint256 _projectId,
+        uint256[] memory _ownedNFTProjectIds
+    ) public onlyArtist(_projectId) {
```
ryley-o, line 219: Not quite sure why we don't include an array of `address[] memory _ownedNFTAddresses` in this function (like we do on MinterHolderV2).

I **think** this minter might be trying to enforce a polyptych minter to only work with projects on the same core contract. That might be slightly over-constraining, and could result in polyptych minting for projects disconnected in time potentially impossible. For example, if a core contract has to be deprecated (i.e. no new projects due to EVM versioning in the future), polyptych minting would not be possible for the projects on the deprecated core contract.

If only allowing polyptych for projects on the same core contract is desired (e.g. if we always want polyptych projects to be performed at the same time), then we should remove the `registerNFTAddress` and `unregisterNFTAddress` functions (since those become irrelevant).

After making these design choices, we should check this minter's interface for commonality with `IFilteredMinterHolderV0` to see if we can simplify any subgraph code down the line 😄

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,704 @@
...
+        // NOTE: delegate-vault handling **ends here**.
+
+        // okay if this underflows because if statement will always eval false.
+        // this is only for gas optimization (core enforces maxInvocations).
+        unchecked {
+            if (tokenId % ONE_MILLION == _projectConfig.maxInvocations - 1) {
+                _projectConfig.maxHasBeenInvoked = true;
+            }
+        }
+
```
ryley-o, line 577: I think we need to have a per-project mapping of "redeemedTokenIds" that is used to limit each token to only be able to be used once to mint a polyptych token. For example, I shouldn't be able to mint 5 tokens with the same hash.

ryley-o (commented):
This is really great, ty @shantanubala!

Let's discuss some of the comments below - async or sync is fine with me!

also, is CI still rugging you on this?

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
jakerockland: Need to think more about this, I'm not sure that we need to introduce this (for now), it is somewhat of a subjective thing as to whether or not we _need_ this (depends on the intended behavior), and given that the actual minting/vouchering of previous mints was going to be handled via off-chain mechanisms given in-person minting, it probably is superfluous to add this check, but it does add a bit of implementation complexity.

That said, the implementation complexity it adds is pretty small, so maybe we do add it just as a safety rail (what you're describing is aligned with the initial planned usage of this minter).

ryley-o: Good point, doesn't 100% need, as this could be handled via admin. I think it would be a nice safety rail, even if admin is running this, as it would protect us against any accidental duplicate mint transactions, etc., which tend to happen if there is a misunderstanding of blockchain state or something.

As discussed elsewhere, if this is not introduced, it does reduce complexity on the randomizer it we elect to guard the `setPolyptychHashSeed`, which we will likely do.

contracts/BasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,80 @@
...
+    function projectIsPolyptych(uint256 _projectId)
+        external
+        view
+        returns (bool)
+    {
+        return _projectIsPolyptych[_projectId];
+    }
+
+    function setPolyptychHashSeed(uint256 _tokenId, bytes12 _hashSeed)
+        external
```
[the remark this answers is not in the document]
shantanubala: Just implemented the changes we discussed below 👍

contracts/BasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,80 @@
...
+                address(this),
+                this.toggleProjectIsPolyptych.selector
+            ),
+            "Only Core AdminACL allowed"
+        );
+        _projectIsPolyptych[_projectId] = !_projectIsPolyptych[_projectId];
+    }
+
+    // Allows the current configuration of a project as a polyptych to be easily queried
+    function projectIsPolyptych(uint256 _projectId)
```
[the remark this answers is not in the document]
shantanubala: Thanks! Fixed 🙏

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,795 @@
...
+        view
+        returns (bytes12);
+}
+
+/**
+ * @title Filtered Minter contract that allows tokens to be minted with ETH
+ * when purchaser owns an allowlisted ERC-721 NFT. This contract does NOT track
+ * if a purchaser has/has not minted already -- it simply restricts purchasing
+ * to anybody that holds one or more of a specified list of ERC-721 NFTs.
+ * This is designed to be used with IGenArt721CoreContractV3 contracts.
```
ryley-o, line 41: I think this should be updated to say it may only be used with V3 Engine contracts that have an active PolyptychRandomizer, and that it is up to admin to appropriately configure the core contract.

contracts/BasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,90 @@
...
+    function setHashSeedSetterContract(address _contractAddress) external {
+        require(
+            genArt721Core.adminACLAllowed(
+                msg.sender,
+                address(this),
+                this.toggleProjectIsPolyptych.selector
+            ),
+            "Only Core AdminACL allowed"
+        );
+        hashSeedSetterContract = _contractAddress;
```
ryley-o, line 43: Although *we* don't currently plan on indexing this value, I recommend emitting an event along the lines of `HashSeedSetterUpdated(_contractAddress)` to support any indexing that partners may want to do. We want to avoid any chance of requiring call handlers in subgraphs.

ryley-o (commented):
ty @shantanubala - a few nits and items to take a look at after this round of iteration.

I think most are straightforward, and would appreciate one last chance to check after these are resolved! 🙏

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,795 @@
...
+        /**
+         * @dev Considered an interaction because calling ownerOf on an NFT
+         * contract. Plan is to only register AB/PBAB NFTs on the minter, but
+         * in case other NFTs are registered, better to check here. Also,
+         * function is non-reentrant, so being extra cautious.
+         */
+        require(
+            IERC721(_ownedNFTAddress).ownerOf(_ownedNFTTokenId) == vault,
+            "Only owner of NFT"
+        );
```
ryley-o, line 679: @shantanubala - as we discussed, wrapping this in an `if (msg.sender != ARTIST) will enable the artist to mint in a privileged manner without requiring the msg.sender to own the token.

After some additional thinking, we could say something along the lines of:
```
if (sender == artist) {
    // check that ownerOf token == vault
} else {
    // check that ownerOf token == _to
}
This would reduce the burden on the artist to ensure that the address they are minting to does in fact own the token, which seems desireable?

File: contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol (hunk 10 above)
Comment (shantanubala) [reply contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol:667]: @jakerockland - I talked to @ryley-o about this earlier today - here's the use case we should consider carefully:

- The artist wants to use the same project ID for a project that has three or more panels
- The tokens of the first panel are always minted out before the second panel is minted
- The second panel will share its token hash seed with the first panel, but in order to mint the third panel, we would need to allow that hash seed to be used more than once

If we want to support a polyptych within the same project (panels that share a project ID), we wouldn't be able to limit the original token's hash seed to a single reuse.

File: contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol (hunk 15 above)
Comment (shantanubala) [reply contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol]: This makes a lot of sense, will add that.

File: contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol [hunk 16]
```diff
@@ -0,0 +1,795 @@
...
+     *
+     * @notice Registers holders of NFTs at address `_NFTAddress` to be
+     * considered for minting. New core address is assumed to follow syntax of:
+     * `projectId = tokenId / 1_000_000`
+     * @param _NFTAddress NFT core address to be registered.
+     */
+    function registerNFTAddress(
+        address _NFTAddress
+    ) external onlyCoreAdminACL(this.registerNFTAddress.selector) {
+        _registeredNFTAddresses.add(_NFTAddress);
```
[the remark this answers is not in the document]
shantanubala: Added 👍

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
jakerockland: Hrm–yeah this makes sense. Thank you both for digging more and re-flagging.

I do think that even if it is superfluous in the diptych framing, having this support from the outset here for gating hash-reuse could be the type of safety mechanism that is just worth the added complexity to introduce from the outset.

WDYT @ryley-o @shantanubala

ryley-o: Some discussion ongoing in Slack, but perhaps a relatively minimal solution to prevent duplicate hashes on a given project is presented here: https://github.com/ArtBlocks/artblocks-contracts/pull/442/files

contracts/BasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,93 @@
...
+    // Sets the token hash seed to be re-used in the subsequent panels of a polyptych
+    function setPolyptychHashSeed(
+        uint256 _tokenId,
+        bytes12 _hashSeed
+    ) external {
+        require(
+            msg.sender == hashSeedSetterContract,
+            "Only hashSeedSetterContract"
+        );
+        polyptychHashSeeds[_tokenId] = _hashSeed;
```
ryley-o, line 67: nit: a comment below this line might be helpful:
```
// @dev event indicating token hash seed assigned is not required for subgraph indexing
// because token hash seeds are still assigned atomically in `assignTokenHash` function.
// If token hash seeds were assigned async, event emission may be required to support
// subgraph indexing.
```

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,809 @@
...
+        // 1) assign a random token hash
+        // 2) if configured, obtain the token hash from the `polyptychSeedHashes` mapping
+        tokenId = minterFilter.mint(_to, _projectId, vault);
+
+        // NOTE: delegate-vault handling **ends here**.
+
+        // redundant check against reentrancy
+        bytes12 _assignedHashSeed = genArtCoreContractWithRandomizer
+            .tokenIdToHashSeed(tokenId);
+        require(_assignedHashSeed == targetHashSeed);
```
ryley-o, line 669: nit: let's add a revert reason of `"Unexpected token hash seed"`

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,809 @@
...
+         */
+        if (msg.sender == _artist) {
+            require(
+                IERC721(_ownedNFTAddress).ownerOf(_ownedNFTTokenId) == vault,
+                "Only owner of NFT"
+            );
+        } else {
+            require(
+                IERC721(_ownedNFTAddress).ownerOf(_ownedNFTTokenId) == _to,
+                "Only owner of NFT"
```
ryley-o, line 691: I think `vault` and `_to` should be flipped here, but please triple check me on this and let's confirm in tests 😅 😅

ryley-o (approved):
@shantanubala 👑 👑 

A few minor comments below, and some ongoing discussion re: duplicate token hashes.

Overall, very excited about this implementation! Please take as much time as you need to write tests - extra tests are encouraged per our [testing philosopy](https://github.com/ArtBlocks/artblocks-contracts/blob/main/test/README.md#philosophy) 🙏

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
jakerockland: ++

contracts/BasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,126 @@
...
+    uint256 constant ONE_MILLION = 1_000_000;
+
+    // Mapping of token IDs to hash seeds
+    mapping(uint256 => bytes12) public polyptychHashSeeds;
+
+    // Mapping of the number of times a hash seed has been used
+    mapping(bytes12 => uint) public hashSeedUsageCounter;
+
+    // Configures the maximum number of hash seed copies that can exist per project (default zero)
+    mapping(uint256 => uint) public projectMaximumHashSeedCopies;
```
shantanubala, line 41: Instead of being a boolean, this keeps track of the number of times we allow a project's token hash to be copied.

This limit is set per-project to allow two different use cases:
1. If polyptych panels are all part of the same project, the artist can increment this maximum limit (default zero) to allow the next panel to be minted on the same project ID
2. If polyptych panels are in a new project, the new project should have a limit that is equal to the previous number of panels

contracts/BasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,133 @@
...
+    uint256 constant ONE_MILLION = 1_000_000;
+
+    // Mapping of token IDs to hash seeds
+    mapping(uint256 => bytes12) public polyptychHashSeeds;
+
+    // Mapping of the number of times a hash seed has been used
+    mapping(bytes12 => uint) public hashSeedUsageCounter;
+
+    // Configures the maximum number of hash seed copies that can exist per project (default zero)
+    mapping(uint256 => uint) public projectDuplicateHashSeedLimit;
```
shantanubala, line 41: Instead of being a boolean mapping, this keeps track of the number of times we allow a project's token hash to be copied.

This limit is set per-project to allow two different use cases:

1. If polyptych panels are all part of the same project, the artist can increment this maximum limit (default zero) to allow the next panel to be minted on the same project ID
2. If polyptych panels are in a new project, the new project should have a limit that is equal to the previous number of panels

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,821 @@
...
+                }("");
+                require(success_, "Render provider payment failed");
+            }
+            // Platform provider (e.g. engine partner) payment
+            if (platformProviderRevenue_ > 0) {
+                (success_, ) = platformProviderAddress_.call{
+                    value: platformProviderRevenue_
+                }("");
+                require(success_, "Platform provider payment failed");
+            }
```
shantanubala, line 752: The lines above ensure we handle the platform payment splits on the engine contract.

contracts/BasicPolyptychRandomizerV0.sol
ryley-o: Consider the following "panels in separate projects" (use case 1 above):
- Base Panel A: project 0
- Panel B: project 1
- Panel C: project 2
- Artist sets `projectDuplicateHashSeedLimit[1]` to 1, `projectDuplicateHashSeedLimit[2]` 2
- If Panel C is minted before Panel B:
  - Panel B can never be minted, because the randomizer has a duplicate hash seed limit of 1 for project 1
  - Panel C can be minted twice with the same token hash
  
Both of those last two bullets are unintended behaviors. To avoid, we essentially require that EVERY token is minted on Panel B's project before minting on C can be opened up, or require that an artist perform the mints via their privileged account while the project is paused, and mint in the intended order of Panel B, Fame C.

I also have some reservation with defining this on the randomizer, mostly in the sense that this feels a bit like ~minter state being tracked on the randomizer. If a new randomizer were used in the future, it would have to look up `hashSeedUsageCounter` and `projectDuplicateHashSeedLimit` values on the old minter based on some kind of logic, for any ongoing projects, which is possible, but is a bit gas inefficient.

Based off of last week's discussion, was thinking we would implement something like the following on the minter itself:
```
// mapping tracks which base tokens have been used in project's current panel
// panelId => baseTokenAddress => baseTokenId => panelIsMinted
mapping(uint256 => mapping(address => mapping(uint256 => bool))) panelIsMinted;
// mapping that identifies a project's current panel id
mapping(uint256 => uint256) projectToCurrentPanelId;

...

// function that allows artist to increment panel id
function incrementPanelId(_projectId) ... {
    ...
    projectToCurrentPanelId[_projectId]++;
}

...

// somewhere within purchase function...
uint256 currentPanelId = projectIdToCurrentPanelId[_projectId]
require(!panelIsMinted[currentPanelId][baseTokenAddress][baseTokenId], "Panel already minted));
!panelIsMinted[currentPanelId][baseTokenAddress][baseTokenId] = true;
...
```

That kind of solution ends up constraining only the mint-order of projects with multiple panels in the same project, in that it requires panel B be minted before panel C if we want panel B to exist. However, due to the nature of tokenIds and artist scripts, this is very likely already a constraint since a project's script must have a sense of which panel is being minted a-priori (e.g. Panel A is tokens 0 through 49, Panel B is tokens 50 through 99, etc.). 

Apologies for the long comment 😅 thoughts?

contracts/BasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,133 @@
...
+     * @notice Allows the owner of the core contract to configure a project as a polyptych
+     * @param _projectId - The ID of the project that is a polyptych panel (second, third, etc.)
+     * @param _maxDuplicates - The number of previous panels in the polyptych
+     */
+    function setprojectDuplicateHashSeedLimit(
+        uint256 _projectId,
+        uint _maxDuplicates
+    )
+        external
+        onlyCoreAdminACL(this.setprojectDuplicateHashSeedLimit.selector)
```
ryley-o, line 79: This function might be better as artist-only instead of admin?

ryley-o (commented):
added some thoughts re: tracking state on minter vs. randomizer - happy to chat more on this if helpful! 🙏

contracts/BasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,133 @@
...
+    IGenArt721CoreContractV3 public genArt721Core;
+
+    // Used to obtain the project ID from the token ID
+    uint256 constant ONE_MILLION = 1_000_000;
+
+    // Mapping of token IDs to hash seeds
+    mapping(uint256 => bytes12) public polyptychHashSeeds;
+
+    // Mapping of the number of times a hash seed has been used
+    mapping(bytes12 => uint) public hashSeedUsageCounter;
```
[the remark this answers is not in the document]
jakerockland: Agreed–that SGTM.

contracts/BasicPolyptychRandomizerV0.sol
jakerockland: ++; also should use `uint256` here rather than `uint` I think, per rationale discussed above.

jakerockland:
Couple follow-up item notes from our conversation with Bright Moments:
- Update minter to allow support for ERC20 minting in addition to / instead of only ETH
- Ensure that ability to view how many times a mint has been used as an external view (not only via private state)

jakerockland:
Couple follow-up item notes from our conversation with Bright Moments:
- Update minter to allow support for ERC20 minting in addition to / instead of only ETH
- Ensure (may already be case) that it is possible to mint from an arbitrary wallet w/ given ERC20 (if project is unpaused and active) given that the destination wallet is the wallet holding the parent-token (the token having its hash copied) or has been delegated this permission via the delegate.cash registry
- Ensure that ability to view how many times a mint has been used as an external view (not only via private state)

jakerockland:
Couple follow-up item notes from our conversation with Bright Moments:
- Update minter to allow support for ERC20 minting in addition to / instead of only ETH
- Ensure (may already be case) that it is possible to mint from an arbitrary wallet w/ given ERC20 (if project is unpaused and active) given that the destination wallet is the wallet holding the parent-token (the token having its hash copied) or has been delegated this permission via the delegate.cash registry
- Ensure that ability to view how many times a mint has been used as an external view (not only via private state)

cc @ryley-o @shantanubala

contracts/BasicPolyptychRandomizerV0.sol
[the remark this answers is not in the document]
shantanubala: Updated 👍

ryley-o:
>Ensure (may already be case) that it is possible to mint from an arbitrary wallet w/ given ERC20 (if project is unpaused and active) given that the destination wallet is the wallet holding the parent-token (the token having its hash copied) or has been delegated this permission via the delegate.cash registry

@jakerockland - I think the minter is behaving very slightly differently, but I think still WAI IIUC:
- it is possible to mint from **artist** wallet w/given ERC20 (in any paused/active state) given that the destination wallet is ownerOf a token with the parent hash (the token having its hash copied, OR
- if unpaused and active, possible to mint from vault or delegate of wallet w/given ERC20 that is ownerOf the token with the parent hash

If we want to enable arbitrary wallets to purchase to any wallet holding a token with the parent hash, we need to modify this minter.

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,1066 @@
...
+    /// Minter filter this minter may interact with.
+    IMinterFilterV0 private immutable minterFilter;
+
+    // Stores whether a panel with an index has been minted for a given base token
+    // panelIndex => baseTokenAddress => baseTokenId => panelIsMinted
+    mapping(uint256 => mapping(address => mapping(uint256 => bool)))
+        public polyptychPanelIsMinted;
+
+    // Identifies a project's current panel ID
+    mapping(uint256 => uint256) public polyptychProjectPanelId;
```
ryley-o, line 111: nit: Let's store this in the `ProjectConfig` struct, as a `uint24` right after `maxInvocations`. That will pack it into the same storage slot as a bunch of other things used during the mint sequence, which will remove a cold SLOAD and save some gas.

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,1066 @@
...
+    IGenArt721CoreContractV3WithRandomizer private immutable genArtCoreContract;
+
+    /// Minter filter address this minter interacts with
+    address public immutable minterFilterAddress;
+
+    /// Minter filter this minter may interact with.
+    IMinterFilterV0 private immutable minterFilter;
+
+    // Stores whether a panel with an index has been minted for a given base token
+    // panelIndex => baseTokenAddress => baseTokenId => panelIsMinted
```
ryley-o, line 106: nit: let's use "panel ID" instead of "panel index" in these comments to be consistent

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,1066 @@
...
+        tokenPriceInWei = _projectConfig.pricePerTokenInWei;
+        currencyAddress = _projectConfig.currencyAddress;
+        if (currencyAddress == address(0)) {
+            // defaults to ETH
+            currencySymbol = "ETH";
+        } else {
+            currencySymbol = _projectConfig.currencySymbol;
+        }
+    }
+}
```
ryley-o, line 1066: I think we wanted to expose a view convenience function to determine if current Panel has been minted for a given parent token's hash, right?

ryley-o (approved):
@shantanubala excellent work, and this minter truly does it all 😄 I don't see obvious ways we could remove any of the logic and still meet the requirements, so LGTM!
IIUC, you are still writing some tests, and I would encourage writing as many tests as possible with this. We are mixing in a lot of different existing capabilities here (delegation, ERC20, token ownerOf), as well as the new parent token hash seed capability.
Mostly just nits here, happy to discuss if you have any questions!

ryley-o:
@shantanubala - please add a test similar to those added in #457 to ensure we cover that edge case 🙏 thanks!

contracts/BasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * @title Basic Polytych Randomizer contract that enables the copying of token hash seeds.
+ * @notice This contract requires several configuration options to be set properly before
+ * it enables the copying of token hashes from an original token (the first panel of a polyptych)
+ * to subsequent diptych, triptych, or polyptych panels.
+ *
+ * Requirements to use the polyptych randomizer:
+ * - Core contract must be an instance of `GenArt721CoreV3_Engine`
+ * - The initial project used as the first panel must be minted with a completely random token hash
```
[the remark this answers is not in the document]
jakerockland: ++

contracts/interfaces/0.8.x/IBasicPolyptychRandomizerV0.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Creatd By: Art Blocks Inc.
+
+pragma solidity 0.8.17;
+
+interface IBasicPolyptychRandomizerV0 {
```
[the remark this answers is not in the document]
jakerockland: I am not sure that this was incorporated unless Github is trippin out? Re-opening for vis.

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,1066 @@
...
+        if (_vault != address(0)) {
+            // If a vault is provided, it must be valid, otherwise throw rather
+            // than optimistically-minting with original `msg.sender`.
+            // Note, we do not check `checkDelegateForAll` or `checkDelegateForContract` as well,
+            // as they are known to be implicitly checked by calling `checkDelegateForToken`.
+            bool isValidVault = delegationRegistryContract
+                .checkDelegateForToken(
+                    msg.sender, // delegate
+                    _vault, // vault
+                    genArt721CoreAddress, // contract
```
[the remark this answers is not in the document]
jakerockland: ++; 🙏

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,1066 @@
...
+            require(
+                IERC721(_ownedNFTAddress).ownerOf(_ownedNFTTokenId) == _to,
+                "Only owner of NFT"
+            );
+        } else {
+            require(
+                IERC721(_ownedNFTAddress).ownerOf(_ownedNFTTokenId) == vault,
+                "Only owner of NFT"
+            );
+        }
```
ryley-o, line 850:
```suggestion
        address owner = IERC721(_ownedNFTAddress).ownerOf(_ownedNFTTokenId);
        require(owner == _to || owner == vault);
```

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,1066 @@
...
+
+        // split funds
+        uint256 _pricePerTokenInWei = _projectConfig.pricePerTokenInWei;
+        address _currencyAddress = _projectConfig.currencyAddress;
```
ryley-o, line 854:
```suggestion
        address _currencyAddress = _projectConfig.currencyAddress;
        if (_currencyAddress == address(0) {
            // ETH
            if (msg.sender == _artist) {
                require(
                    IERC721(_ownedNFTAddress).ownerOf(_ownedNFTTokenId) == _to,
                    "Only owner of NFT"
                );
            } else {
                require(
                    IERC721(_ownedNFTAddress).ownerOf(_ownedNFTTokenId) == vault,
                    "Only owner of NFT"
                );
            }
        } else {
            // ERC20
            address _owner = IERC721(_ownedNFTAddress).ownerOf(_ownedNFTTokenId);
            require(_owner == _to || owner == vault);
        }
        // split funds
        uint256 _pricePerTokenInWei = _projectConfig.pricePerTokenInWei;
```

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,1066 @@
...
+    }
+
+    /**
+     * @dev splits ERC-20 funds between foundation, artist, and artist's
+     * additional payee, for a token purchased on project `_projectId`.
+     * @dev possible DoS during splits is acknowledged, and mitigated by
+     * business practices, including end-to-end testing on mainnet, and
+     * admin-accepted artist payment addresses.
+     */
+    function _splitFundsERC20(
```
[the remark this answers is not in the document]
jakerockland: I think splitting for now makes sense rather than burning.

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
jakerockland: Yes please, ++ :)

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
jakerockland: ++; I think this is necessary to allow for the flow that we discussed w/ BMG

jakerockland (approved):
Very close to done here 🙌 

Added a couple ++'s to a few things and flagged the destination check logic which @ryley-o and I discussed offline and left a note about here: https://github.com/ArtBlocks/artblocks-contracts/pull/410/files#r1084692381 (also related to your comment above @ryley-o 🙏 ).

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
shantanubala: Would this be consistent with the mapping here being public? Aside from making the mapping public, is another method required?

```
mapping(uint256 => mapping(address => mapping(uint256 => bool))) public polyptychPanelIsMinted;
```

This means you can call it like:

```
minterContract.polyptychPanelIsMinted(_panelId, _ownedNFTAddress, _ownedTokenId);
```

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,1088 @@
...
+    function polyptychPanelMintedWithToken(
+        uint24 _panelId,
+        address _ownedNFTAddress,
+        uint256 _ownedNFTTokenId
+    ) external view returns (bool panelMinted) {
+        return
+            polyptychPanelIsMinted[_panelId][_ownedNFTAddress][
+                _ownedNFTTokenId
+            ];
+    }
```
shantanubala, line 214: @ryley-o How does this look to you?

test/minter-suite-minters/PolyptychMinter/PolyptychMinterV0.test.ts
```diff
@@ -0,0 +1,837 @@
...
+        // expect field to be address of delegation registry as data 1
+        // zero-pad address to 32 bytes when checking against event data
+        const abiCoder = new AbiCoder();
+        expect(targetLog.data).to.be.equal(
+          abiCoder.encode(["address"], [this.delegationRegistry.address])
+        );
+      });
+    });
+
+    describe("polyptychPanelMintedWithToken", async function () {
```
shantanubala, line 259: Corresponding tests here for the getter function

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
shantanubala: https://github.com/ArtBlocks/artblocks-contracts/pull/410/files#r1090995019

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
jakerockland: Change in https://github.com/ArtBlocks/artblocks-contracts/pull/410#discussion_r1090992761 looks great to me!

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
```diff
@@ -0,0 +1,1088 @@
...
+
+    struct ProjectConfig {
+        bool maxHasBeenInvoked;
+        bool priceIsConfigured;
+        uint24 maxInvocations;
+        address currencyAddress;
+        uint256 pricePerTokenInWei;
+        string currencySymbol;
+        uint24 polyptychPanelId;
+    }
```
ryley-o, line 128: nit: solidity quirk - ordering matters when it comes to struct packing 😝 suggest the following to pack the first five parameters into a single 32-bytes slot, which will help w/gas efficiency during mint 👍  
```suggestion
    struct ProjectConfig {
        bool maxHasBeenInvoked;
        bool priceIsConfigured;
        uint24 maxInvocations;
        uint24 polyptychPanelId;
        address currencyAddress;
        uint256 pricePerTokenInWei;
        string currencySymbol;
    }
```

contracts/minter-suite/Minters/MinterPolyptych/MinterPolyptychV0.sol
ryley-o: looks great!

PR #415, opened 2022-12-22
V3 Engine Contract Deployment Scripts

jakerockland:
Diff between the `goerli` and `mainnet` scripts: https://www.diffchecker.com/a50vSIfs

---

A couple notes to follow up on in the new year:

- The core contracts above are too complicated to verify via passing in all construction arguments via command line arguments, and instead it was necessary to verify by following the [complex arguments instructions](https://hardhat.org/hardhat-runner/plugins/nomiclabs-hardhat-etherscan#complex-arguments) for Etherscan verification.
    - (Art Blocks x Pace) https://etherscan.io/address/0xEa698596b6009A622C3eD00dD5a8b5d1CAE4fC36#code
    - (Art Blocks x Bright Moments) https://etherscan.io/address/0x145789247973C5D612bF121e9E4Eef84b63Eb707#code
    - (Art Blocks Managed Engine) https://etherscan.io/address/0x73b4797E2FD04fA42a9F3c9bCfbcEe19374A9060#code
- Example details of complex argument verification: `yarn hardhat verify --constructor-args artblocks-x-pace-arguments.js 0xEa698596b6009A622C3eD00dD5a8b5d1CAE4fC36 --network mainnet` where `artblocks-x-pace-arguments.js ` is:
```
module.exports = [
    "Art Blocks x Pace", // name
    "ABXPACE", // ticker
    "0xB8559AF91377e5BaB052A4E9a5088cB65a9a4d63", // render provider
    "0xB8559AF91377e5BaB052A4E9a5088cB65a9a4d63", // platform provider
    "0x50DB09A76E9B762d9161c1710102B8C407Fd3ae0", // randomizer
    "0x4F68170A7b3C9B52780289ab2E50a5C26b08B09C", // admin acl
    5, // starting project id
    false, // auto approve artist split proposals
    "0x652490c8BB6e7ec3Fd798537D2F348D7904BBbc2" // engine registry
  ];
```

- The mint #0 action above is actually failing silently by the scripts, so I have re-performed this manually – relevant Etherscan links below (mainnet only)
    - Art Blocks x Pace
        - (Failed, automated) https://etherscan.io/tx/0xf63d48b0f208fe104955adf98746857ee0155f9ff0d67de6820a04debb498869
        - (Successful, manual) https://etherscan.io/tx/0xa0b92185f3b330e3926317d2083c525e58f5855d06008774e2c36c33b81c12d2
[...]

PR #416, opened 2023-01-04
Add minor, internal-only changes to V3_Engine & Registry contracts

ryley-o:
These changes are minor and in no way invalidate already-deployed V3_Engine contracts referenced in PR #415.

It may be worth the added security benefit to deploy a new EngineRegistryV0 and just add the engine contracts in PR #415 manually from the deployer wallet.

### EngineRegistryV0
The changes here prevent the case where our deployer wallet calls some unrelated, malicious contract, and the malicious contract adds/removes engine contracts from the EngineRegistry. This is a low-level vulnerability (since it could be remedied via subgraph changes), but may be worth patching and re-deploying.

### V3_Engine (Core)
Changes are limited to very minor internal nits.

### DIFF
--- a/contracts/EngineRegistryV0.sol
+++ b/contracts/EngineRegistryV0.sol
@@ -43,7 +43,11 @@ contract EngineRegistryV0 is IEngineRegistryV0, ERC165 {
             tx.origin == deployerAddress,
             "Only allowed deployer-address TX origin"
         );
-
+        // Prevent registration of a contract by an unrelated contract
+        require(
+            msg.sender == _contractAddress || msg.sender == deployerAddress,
+            "Only call by deployerAddress or _contractAddress"
+        );
         // EFFECTS
         emit ContractRegistered(_contractAddress, _coreVersion, _coreType);
         registeredContractAddresses[_contractAddress] = true;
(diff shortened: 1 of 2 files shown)


jakerockland:
@ryley-o I say we bump the `CORE_VERSION` to `"v3.1.1"` with these fixes as well, WDYT?

ryley-o:
>@ryley-o I say we bump the CORE_VERSION to "v3.1.1" with these fixes as well, WDYT?

Yeah, I think that is a good idea! (even though it won't affect how we index anything)

jakerockland:
> It may be worth the added security benefit to deploy a new EngineRegistryV0 and just add the engine contracts in PR https://github.com/ArtBlocks/artblocks-contracts/pull/415 manually from the deployer wallet. This is due to the minor security improvements made to that contract in this PR (and the relatively easy lift it would be to ship a new instance of that contract at this time).

@ryley-o this SGTM, would you be willing to follow up with a deployment script PR?

jakerockland (approved):

ryley-o:
>would you be willing to follow up with a deployment script PR?

@jakerockland definitely will do!

PR #425, opened 2023-01-06
Opcodes MinterDAExpSettlementV0 Audit Actions

ryley-o:
Update MinterDAExpSettlementV0 with audit actions identified from Opcodes audit.

Overall, average mint costs are reduced by ~6.7%, and redundant reentrancy protection is added. Gas costs of reclaiming excess funds are also very slightly reduced. A few coding review items are acknowledged and not implemented - see discussion below for more details.

No changes are required to downstream integrations such as our subgraph, and this minter is still not deployed to staging or mainnet, therefore the version number of the minter is unchanged.

The following items were considered in this update, all highlighted in the Opcodes audit:

Vulnerabilities
- 1-3. (Acknowledged/WAI)
- 4. (Implemented) Improved robustness by adding redundant reentrancy guard to withdrawArtistAndAdminRevenues() function
  - note: this adds redundant reentrancy protection - previous code was okay due to checks-effects-interactions pattern (I typically prefer redundant reentrancy protection to be fault tolerant in these areas)

Coding Review
- 1. (Acknowledged) Custom errors vs. revert - I'm hesitant to break consistency with existing minters due to additional workload of updating tests. I've added issue #426, with the recommendation that we implement custom errors during a larger refactor of our minter suite.
- 2. (Implemented) Packing prices reduced mint gas costs by ~1.8%. Opted to keep interfaces consistent with prior versions (i.e. functions continue to ask for and emit events with uint256 values).
- 3. (Acknowledged) Only benefits non-mint transactions, and requires new custom getter functions if interfaces are to be maintained. Opted to not implement at this time.
- 4. (Implemented) Reduced mint costs by ~4.7% by packing values in Receipt struct 🎉 
- 5. (Implemented) Reduced mint costs by ~0.5% by implementing memoization to remove a few SLOADs
- 6. (Implemented) Reduced reclaim excess settlement funds gas costs by ~0.5% by implementing memoization to remove a few SLOADs
- 7. (Implemented) negligible gas reduction, cleaner code by removing unnecessary initialization of variable to zero
- 8. (Acknowledged) Despite my initial inclination to add more data to the `ArtistAndAdminRevenuesWithdrawn` event, revenue data are already available in our subgraph. We already index `currentSettledPrice` and `numSettleableInvocations` in our subgraph, making total revenues collected simply available by the product of those two values. Changing the event signature at this point in our development process will create unnecessary iteration and not enable us to expose any new values in our subgraph. For those reasons, opted to not implement this suggestion.


ryley-o:
Cc @Jbern16 - this will be what gets deployed to artist staging and mainnet, but external interfaces remain unchanged from what is currently deployed to dev 👍

contracts/minter-suite/Minters/MinterDAExpSettlement/MinterDAExpSettlementV0.sol
```diff
@@ -98,9 +98,9 @@ contract MinterDAExpSettlementV0 is
         // max uint64 ~= 1.8e19 sec ~= 570 billion years
         uint64 timestampStart;
         uint64 priceDecayHalfLifeSeconds;
-        uint256 startPrice;
+        uint128 startPrice;
         // base price is non-zero for all configured auctions on this minter
-        uint256 basePrice;
+        uint128 basePrice;
```
jakerockland, line 103: We should consider a documentation comment that clarifies what the max price is in ETH terms.

jakerockland (approved):
This is great, thank you for the quick PR here @ryley-o and for the very robust and thoughtful PR description.

contracts/minter-suite/Minters/MinterDAExpSettlement/MinterDAExpSettlementV0.sol
ryley-o: good call, added a comment for this!

ryley-o:
I've updated this PR retroactively to indicate that the actions came out of an Opcodes coding review. Originally, the PR used the term "Audit", which was incorrect.

PR #429, opened 2023-01-06
Upgrade dependencies pt2

jakerockland:
To be merged after #428 

This updates `devDependencies` that fall in the `"<yellow>" : Minor Update backward-compatible features` category using `yarn upgrade-interactive --latest`.


ryley-o (approved):
LGTM 🙏 
@jakerockland - I pushed a commit to lock to our previous version of the `@gnosis.pm/safe-ethers-lib` to let our existing tests pass. I don't think sinking a bunch of time into a re-write of those tests adds much value to our repo at this time, as our main intent of minting with gnosis safe is to ensure that we allow smart-contract wallets to mint, in general.

ryley-o:
>That SGTM Ryley, though may be worth looking into in the future if Gnosis dependency is transitively bringing in dependencies that have vulnerabilities that cannot otherwise be upgraded.

This is a good point. I don't know what the issue is with the latest versions of the gnosis safe dependencies. Their latest EthersAdapter class seems to not work well with hardhat ethers...

Going to merge this and maybe tackle Safe in a follow-on PR...

jakerockland:
> > That SGTM Ryley, though may be worth looking into in the future if Gnosis dependency is transitively bringing in dependencies that have vulnerabilities that cannot otherwise be upgraded.
> 
> This is a good point. I don't know what the issue is with the latest versions of the gnosis safe dependencies. Their latest EthersAdapter class seems to not work well with hardhat ethers...
> 
> Going to merge this and maybe tackle Safe in a follow-on PR...

SGTM ser–thank you!

PR #442, opened 2023-01-10
Prevent duplicate token hashes on a given project

ryley-o:
A potential solution to prevent duplicate token hashes on a given project.

This forces the polyptych to adhere to the rule “No duplicate token hashes allowed on a given project”. This would require that if an artist wanted, for example, 3 “frames”, frame 1 would have its own project, frame 2 would need a separate project, and frame 3 would need another project.

This solution is implemented entirely on the randomizer. Data are also exposed as a public mapping, making future migrations to different randomizers possible during active project minting.


ryley-o:
closing this PR because a better solution for intended use cases was identified earlier today 🎉

PR #443, opened 2023-01-11
Archive Engine Contracts

### LINKED ISSUE #424: Revisit how we manage contracts repo for Engine partners
My inclination is that we may want to consider just having deployment scripts as something that gets stored in this repo for posterity, but not clone the contracts themeselves each time we do a new deployment.

I'd say that we consider this change as something we start doing with the introduction of V3 engine contracts into the wild–we could then migrate the DEPLOYMENT logs into the deployment script sub-dirs, and archive the contracts for engine contracts into a posterity folder.

Something warrant further discussion – happy to do async or sync @ryley-o – but just wanted to flag this as being worth consideration.

ryley-o:
Update how engine contract source code is archived, and update process around mainnet deployments to include tagged releases in this repository.

This fixes #424.

A benefit of this archival process is that already-deployed engine contracts are not compiled unnecessarily during development. It also prevents constant maintenance of the paths in `.solcover.js`, which tends to be forgotten and muddies up our coverage reports.


jakerockland:
@ryley-o I think should wait until V3 contracts are _fully production ready_ to be used in production to make this change (and the change in #444) as for now we still have ongoing Engine deployments in the pipeline and I think it is simpler to make the split clean with regard to when we change things to coincide with a new V3-oriented release process. WDYT here ser?

ryley-o:
@jakerockland This still enables deployments of V2 contracts and is intended to be more of a change in how we manage the contract code after contract(s) are deployed to mainnet for a given Engine partner. Specifically, we don't need to continue to compile specific partner contracts that are already deployed and no longer being actively developed. I don't think that is too much related to V2 vs. V3.

Also, we should definitely keep things in the `/contracts/engine/` directory that are actively being developed/deployed (e.g. only deployed to testnet). However, once deployed to mainnet and a release is tagged in this repo, archiving them after that seems like a good process to me.

Thoughts?

jakerockland (approved):

jakerockland:
> @jakerockland This still enables deployments of V2 contracts and is intended to be more of a change in how we manage the contract code after contract(s) are deployed to mainnet for a given Engine partner. Specifically, we don't need to continue to compile specific partner contracts that are already deployed and no longer being actively developed. I don't think that is too much related to V2 vs. V3.
> 
> Also, we should definitely keep things in the `/contracts/engine/` directory that are actively being developed/deployed (e.g. only deployed to testnet). However, once deployed to mainnet and a release is tagged in this repo, archiving them after that seems like a good process to me.
> 
> Thoughts?

Roger roger–I misunderstood the intention here, that SGTM.

ryley-o:
This PRs description has been updated to highlight some of the archival process, and Issue #424 will be updated to discuss some of the desired future state for the V3_Engine deployment process 🙏

jakerockland:
Thank you very very much ser!

PR #446, opened 2023-01-11
Add OpenZeppelin guidelines, refactor readme content organization

ryley-o:
Add OpenZeppelin guidelines and refactor readme content organization to make contributing easier to understand.

As part of this, we are officially saying that we want all new code to conform to OpenZeppelin's guidelines, which includes following the official Solidity style guide. That will be a change for some contracts (e.g. we typically use underscores in our function arguments), but a more solidified style guide is more important IMO.


jakerockland (approved):
This is awesome – thank you @ryley-o 

cc @shantanubala

PR #447, opened 2023-01-12
Use common V3 core base interface

ryley-o:
Refactor interfaces to use a common V3 core base interface to codify common portions of flagship derivatives (reduces subgraph code).

This is intended to be a pure refactor that has the benefits of:
- codifying the common portions of various V3 core contract derivatives (e.g. flagship and engine)
- greatly reducing subgraph code complexity/redundancy

Cc @lyaunzbe due to follow-on impacts to V3_EngineFlex work


jakerockland (approved):
🙏 🙏 🙏

Issue #371, opened 2022-10-31
Utilize Hardhat fixtures in our testing suite


ryley-o:
We currently use Mocha's `beforeEach` hook in our testing suite, but we would achieve a faster testing run time if we were to refactor and use [Hardhat fixtures](https://hardhat.org/hardhat-runner/docs/guides/test-contracts#using-fixtures).

This would likely be a relatively simple refactor, and the impact on test run time is currently not quantified.

PR #448, opened 2023-01-13
Prototype hardhat fixtures

ryley-o:
Prototype hardhat fixtures

Quick prototype build to identify the carrot of implementing #371.

An early subjective look shows that using a fixture instead of relying on testing framework's `beforeEach` reduces test time by a large percentage. However, there will be a decent amount of refactoring (we often use `this` to pass data around in current tests), and some complexities will arise in the areas where block times matter (e.g. when analyzing auction prices through times) due to Hardhat's constraint of current block timestamp > previous block timestamp


ryley-o:
closing in favor of #498

PR #449, opened 2023-01-19
V3_Engine compatible minter suite

ryley-o:
## Description of the change

Increment Minter Suite Minter versions and make them all compatible with V3_Engine core contracts (while maintaining V3 Flagship compatibility).

This provides Engine partners with compatible minters for V3_Engine contracts, while also taking us one step closer to the future state of minters that work with any V3 core contract.

Note that the main difference between Flagship Core and Engine Core contracts from the perspective of a minter is the `getPrimaryRevenueSplits` function - Flagship cores respond with three values and addresses, Engine cores respond with four values and addresses. The response must be appropriately handled when splitting funds.

## Design Choices
- All minters versions were increased
- `MinterDAExpSettlementV1` contains the logic necessary to integrate with an Engine core contract, but remains as the only minter that doesn't support a minter-level maxInvocations limit.
  - This is because that minter requires special attention when operating on maxInvocations due to it altering its behavior when an auction is "complete". A surgical PR focused on that change is much more appropriate than here.
- During construction, minters automatically infer if the core contract is an Engine contract or not based on the returned `coreType()` of the input `genArt721Core` contract address.
  - If not `GenArt721CoreV3`, the contract is assumed to be an engine contract. Note that Explorations contracts behave like flagship and respond with a coreType of `GenArt721CoreV3`.
  - After inferring type based off of `coreType()`, the length of the response to `getPrimaryRevenueSplits` is verified to either be `6*32 bytes` if the minter thinks the contract is a Flagship, or `8*32 bytes` if the minter thinks the contract is Engine.
    - This is purely for safety of the auto-configuration, which could result in loss of funds if a misconfiguration occurred.
- A `MinterUtils` library is added that houses some of the redundant, initialization checking logic
  - This may be expanded in the future and seems like a helpful design pattern to remove some of the more "cluttering" functions in minters that may not be very helpful to a user reading the contract.

## Tests
Tests are added that are intended to meet or exceed previous levels of coverage on the new minters, including all new added logic.

## Follow-on
These changes are comprehensive of these minter versions. Follow-on work will be done on subgraph to ensure these minters are handled properly.

Further opportunity to utilize the `MinterUtils` library to clean up redundant minter code may be pursued.


ryley-o:
diffchecker of minter diff here: https://www.diffchecker.com/GYifGN44/

contracts/libs/0.8.x/MinterUtils_v0_1_0.sol
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * @notice Validates that a GenArt721CoreV3 core contract's
+     * `getPrimaryRevenueSplits` function returns the expected number of
+     * return values based on the `isEngine` expected state of the core
+     * contract.
+     * @param isEngine Whether the core contract is expected to be an Art
+     * Blocks Engine contract or not.
+     * @param genArt721CoreAddress The address of the deployed core contract.
+     */
+    function ValidateV3CoreGetPrimaryRevenueSplitsResponse(
```
jakerockland, line 24: May want to consider making this a `modifier`?

ryley-o: @jakerockland - We can't call a modifier in a library from the minter (modifiers may only be called in the current class or a parent class). Additionally, even if we moved this to a parent class, we only know if `_isEngine` is true or false after some logic is performed in the constructor function itself.
We could move this function's logic to a modifier within the library itself, but I don't think that is what you mean here, yeah?

jakerockland:
> This may be expanded in the future and seems like a helpful design pattern to remove some of the more "cluttering" functions in minters that may not be very helpful to a user reading the contract.

Def like the idea of considering other areas to expand this in the future (e.g. potentially payment splitting logic itself).

contracts/libs/0.8.x/MinterUtils_v0_1_0.sol
jakerockland: Ah – roger that ser, SGTM. In that case, I think implemented as-is makes sense.

> Also note that we could move _isEngine to the minter's specified constructor args, but was thinking it is probably worth the extra on-chain gas during deployment to take some administrative burden off of the deployer and let the minter auto-configure itself as integrating with an engine or flagship core during deployment.

Very much agree with your take here.

jakerockland (approved):
This LGTM – with the one note that I do wonder if we want to consider moving `_splitFundsETH` into our shared library as part of this change set...

ryley-o:
>This LGTM – with the one note that I do wonder if we want to consider moving _splitFundsETH into our shared library as part of this change set...

@jakerockland I very much agree here, would prefer to toss that as a follow-on PR today 🙏 It will be a refactor of internal functions only, so it won't have any impacts to our downstream services 🎉

jakerockland:
> > This LGTM – with the one note that I do wonder if we want to consider moving _splitFundsETH into our shared library as part of this change set...
> 
> @jakerockland I very much agree here, would prefer to toss that as a follow-on PR today 🙏 It will be a refactor of internal functions only, so it won't have any impacts to our downstream services 🎉

SGTM ser.

jakerockland:
Thank you! :)

PR #455, opened 2023-01-20
Goerli Engine + Engine Flex deployments (late January '23)

jakerockland:
## Description of the change

Deployment details for the late January '23 Engine/Engine Flex deployments (goerli only).

Contract deployment work for the following:

- https://github.com/ArtBlocks/artblocks/issues/410
- https://github.com/ArtBlocks/artblocks/issues/401
- https://github.com/ArtBlocks/artblocks/issues/403
- https://github.com/ArtBlocks/artblocks/issues/416

Additionally, this PR restructures our goerli deployment template to use a shared contract reference paradigm, in alignment with the directional goals of our V3 Engine contract work (a stepping stone towards that) and also re-adjusts some of our `/posterity` directory management in accordance with this as well.

Please note that due to https://github.com/NomicFoundation/hardhat/issues/2287 some of these scripts "fail early" in verification process, but not in a way that is actually disruptive. 

That said, when https://github.com/NomicFoundation/hardhat/issues/2287 is resolved that will have a nice impact on our team, or we can implement some of the workarounds described in that thread in a follow-on PR if we're finding the current behavior to be too disruptive.


ryley-o (approved):
LGTM 🙏 :babby:

PR #456, opened 2023-01-23
Bug fix on MinterHolder

ryley-o:
## Description of the change

Delegation registry logic is updated to check delegation for owned token's core contract instead of tokenID on minter's core contract. This is limited to the MinterHolder logic only (MinterMerkle is unaffected)

This contract has not been deployed to mainnet, so no further action is required.

Please triple-check this updated logic to ensure this is the desired logic 🙏


ryley-o:
> we should probably also have test coverage for this path as well – but can add that in followup PR if you would prefer

Agreed - will add in a follow-up PR later today

PR #459, opened 2023-01-24
use appropriate DALin Interface in minters

ryley-o:
## Description of the change

Refactor DALin active minters to use appropriate interfaces, ensuring that all events are defined in the interfaces instead of the minter contract itself.

This is something that must have been unintentionally missed when writing DALinV2, because the Interfaces were already appropriately created, but were not being used.

This is a refactor and does not affect minter logic. It is implemented to clean up some of the subgraph work.

Cc @lyaunzbe


jakerockland (approved):

lyaunzbe:
👍

PR #460, opened 2023-01-25
Add action to publish to npm on release

yoshiwarab:
## Description of the change
This PR adds a github action workflow to publish the @artblocks/contracts package to npm on release. As part of this, a new hardhat config file specifically for this action has been added so that we don't require any private keys to compile. 

@jakerockland @ryley-o note that since the posterity directory is not within the contracts directory those contracts will not be compiled and won't make it into the build. We do want to be able to access those abis on the frontend so it may be worth moving them into the contracts directory.


jakerockland (approved):
This is very rad ser. Thank you very much!!

Regarding how this impacts the strategy for `/posterity` – before merging this PR it may be worth thinking this through a bit more as a trio between you @yoshiwarab, @ryley-o, and I, as moving items back out of posterity has its own headaches and I'm not sure if that means we may want to adjust the strategy of this github action instead (e.g. create temporary copy of `/posterity` into `/contracts` as part of this action).

ryley-o (approved):
this is really great! will work so much better with our infra.

re: un-archiving contracts... I'm leaning towards preferring to move any contracts that we want to be available in the npm package into our contracts directory, but in a deprecated/posterity directory. While it would be ideal to not compile those contracts when, e.g. we are running tests, managing the imports and import paths of files in our posterity folder to work if we temporarily copy them in the gh action seems like it would trip us up too frequently.

We could un-archive in a follow-on PR... @yoshiwarab - I'm happy to take that on -  which contracts do we want to make available? I'm thinking all previous core contracts, and maybe all previous minters?

yoshiwarab:
> We could un-archive in a follow-on PR... @yoshiwarab - I'm happy to take that on - which contracts do we want to make available? I'm thinking all previous core contracts, and maybe all previous minters?

Cool, that sounds good to me! And yes I think we'll want all previous core contracts and all previous minters

PR #463, opened 2023-01-25
Refactor minters with MinterUtils library

ryley-o:
## Description of the change

Refactor minters to remove redundant internal function code to the MinterUtils internal library.

Follow-on activity from #449.


contracts/libs/0.8.x/MinterUtils_v0_1_1.sol
```diff
@@ -0,0 +1,276 @@
...
+     * @notice Validates that a GenArt721CoreV3 core contract's
+     * `getPrimaryRevenueSplits` function returns the expected number of
+     * return values based on the `isEngine` expected state of the core
+     * contract.
+     * @param isEngine Whether the core contract is expected to be an Art
+     * Blocks Engine contract or not.
+     * @param genArt721CoreV3 The address of the deployed core contract.
+     */
+    function _validateV3CoreGetPrimaryRevenueSplitsResponse(
+        bool isEngine,
```
jakerockland, line 250: Should this be passed in, or should this just be determined internally (via a call to `getV3CoreIsEngine`)?

contracts/libs/0.8.x/MinterUtils_v0_1_1.sol
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }
+
+    function getV3CoreIsEngine(
+        IGenArt721CoreContractV3_Base genArt721CoreV3_Base
+    ) internal returns (bool isEngine) {
+        // determine if core is engine based off of coreType() response
+        isEngine =
+            keccak256(abi.encodePacked(genArt721CoreV3_Base.coreType())) ==
+            keccak256(abi.encodePacked(FLAGSHIP_CORE_TYPE));
```
jakerockland, line 231: I wonder if we should check types against all expected types (rather than just checking flagship)–and throw if we hit an unexpected type?

contracts/libs/0.8.x/MinterUtils_v0_1_1.sol
jakerockland: Ah wait–I see; the abstraction is happening this way, it is just happening in reverse, with `getV3CoreIsEngine` being the `internal` entry point and this method being `private`.

contracts/libs/0.8.x/MinterUtils_v0_1_1.sol
ryley-o: I think that would be ideal, but I didn't want to constrain our minters to a set of engine types in case we add more in the future. Right now, we have flagship and explorations that return three payees, and both of those (intentionally) return the same coreType. When we add EngineFlex in the future, this will automatically work.

Another option here is to simply skip the coreType check altogether, and just define isEngine based on the length of the bytes returned from the get primary splits function.

contracts/libs/0.8.x/MinterUtils_v0_1_1.sol
ryley-o: Yeah, thought we could let the library simply get and return if the core is engine, while validating that answer internally.

contracts/libs/0.8.x/MinterUtils_v0_1_1.sol
jakerockland: > Another option here is to simply skip the coreType check altogether, and just define isEngine based on the length of the bytes returned from the get primary splits function.

This def feels the move.

ryley-o: I'll refactor this to look only at the returned bytes, per our discussion 🎉

ryley-o: Refactor has been pushed 🥳

ryley-o:
@jakerockland - this is ready for a re-review now that it has been refactored to use a MinterBase class🙏

ryley-o:
@jakerockland - wanted to note that I eliminated the UpdatedIsEngine event during construction and from the interface. After integrating these minters into our indexing, that event seemed unnecessary and unhelpful, and removing it simplifies the interface model in this PR quite a bit.

contracts/minter-suite/Minters/MinterBase_v0_1_1.sol
```diff
@@ -0,0 +1,280 @@
...
+            "getPrimaryRevenueSplits(uint256,uint256)",
+            0,
+            0
+        );
+        (bool success, bytes memory returnData) = genArt721CoreV3.call(payload);
+        require(success, "getPrimaryRevenueSplits() call failed");
+        // determine whether core is engine or not, based on return data length
+        uint256 returnDataLength = returnData.length;
+        if (returnDataLength == 6 * 32) {
+            // 6 32-byte words returned if flagship (not engine)
```
jakerockland, line 270: Nit: expand this dev comment to clarify why 6 are expected; that there is a payout address and % for the: artist, artist additional, and art blocks

jakerockland (approved):
Have a couple nit suggestions–happy to discuss more as needed, otherwise LGTM! :)

contracts/minter-suite/Minters/MinterBase_v0_1_1.sol
ryley-o: good call, updated!

contracts/minter-suite/Minters/MinterBase_v0_1_1.sol
```diff
@@ -0,0 +1,280 @@
...
+            return false;
+        } else if (returnDataLength == 8 * 32) {
+            // 8 32-byte words returned if engine
```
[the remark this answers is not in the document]
ryley-o: good call, updated!

contracts/minter-suite/Minters/MinterBase_v0_1_1.sol
```diff
@@ -0,0 +1,280 @@
...
+            bool success;
+            // split funds between platforms, artist, and artist's
+            // additional payee
+            uint256 renderProviderRevenue_;
+            address payable renderProviderAddress_;
+            uint256 artistRevenue_;
+            address payable artistAddress_;
+            uint256 additionalPayeePrimaryRevenue_;
+            address payable additionalPayeePrimaryAddress_;
+            if (!isEngine) {
```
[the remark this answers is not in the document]
ryley-o: good call, updated!

contracts/minter-suite/Minters/MinterBase_v0_1_1.sol
```diff
@@ -0,0 +1,280 @@
...
+     * admin-accepted artist payment addresses.
+     * @param projectId Project ID for which funds shall be split.
+     * @param valueInWei Value to be split, in Wei.
+     */
+    function splitRevenuesETH(
+        uint256 projectId,
+        uint256 valueInWei,
+        address genArtCoreContract
+    ) internal {
+        if (valueInWei > 0) {
```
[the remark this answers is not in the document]
ryley-o: good call, updated!

contracts/libs/0.8.x/MinterUtils_v0_1_1.sol
jakerockland: Wooo–thank you ser :)

contracts/libs/0.8.x/MinterUtils_v0_1_1.sol
jakerockland: Yep yep–makes sense; resolving.

contracts/interfaces/0.8.x/IFilteredMinterV3_Mixin.sol
```diff
@@ -12,11 +12,6 @@ pragma solidity ^0.8.0;
  * @author Art Blocks Inc.
  */
 interface IFilteredMinterV3_Mixin {
```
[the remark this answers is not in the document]
ryley-o: yes, good callout. I've pushed an update to move the IsEngine public getter to a IMinterBaseV0 interface, and was able to remove all of the IFilteredMinterV3* interface stuff, which removed an entire interface version roll across our entire minter suite 🎉 

I'll tag you in some follow-on subgraph refactoring in a separate PR, but note that this in no way affects our ability to publish a subgraph (existing logic works).

contracts/libs/0.8.x/MinterUtils_v0_1_1.sol
ryley-o: resolving this based on subsequent review feedback 🙏

jakerockland (approved):
Thank you for the changes ser – LGTM.

contracts/interfaces/0.8.x/IFilteredMinterV3_Mixin.sol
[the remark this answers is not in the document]
jakerockland: LG LG :)

PR #464, opened 2023-01-27
Compile selected archived contracts

ryley-o:
## Description of the change

Compile selected archived contracts, included all published core contracts and minters, to enable their ABIs to be included in our published npm package.

In general, I've moved everything that is archived/deprecated into `/contracts/archive/` to ensure they are compiled and their import statements are in working order, but to indicate that they are no longer being actively developed.

Readme files are also updated to reflect this change.

This is follow-on work from #460 (per this [comment](https://github.com/ArtBlocks/artblocks-contracts/pull/460#issuecomment-1404243461))


jakerockland (approved):
🙏

PR #465, opened 2023-01-27
re-organize engine deployment details

ryley-o:
## Description of the change

Store Art Blocks Engine deployment details/logs in a root `/deployments/` directory instead of inside the `/scripts/` directory.

This organization more clearly divides the responsibility of deployment scripts and deployment logs. In a follow-on PR, the V3_Engine deployment script will be updated to be able to handle `input.json` files (also stored in the `/deployments/` directory, which will fully decouple specific engine deployment details from any specific contract or deployment script files. This will enable removing the requirement to copy and paste solidity code or typescript code when performing an Engine deployment.


jakerockland (approved):
Very excited about this progress. !vibe

PR #467, opened 2023-01-30
fix solcover syntax bug

ryley-o:
## Description of the change

Fix a bug in .solcover that is preventing all branches from merging

(not sure how this bug made it into main 🤷)


jakerockland (approved):
TYVM ser.

PR #476, opened 2023-02-02
Da exp settlement v2

ryley-o:
## Description of the change

Add DA Settlement minter that enables minter-local max invocations that may be less than or equal to the project's max invocations defined on the core contract.

This is similar to the functionality recently added to all other minter contracts, but with some added complexity of change on this minter, since "definition of done" is an important state to understand on a minter with on-chain settlement.

Note that this minter has the following trust assumption (which was also present in a slightly less-direct way in earlier versions of this minter):
- The artist is trusted to not reduce their project's minter-local maxInvocations mid-auction, inducing a `maxHasBeenInvoked` of true, and immediately withdrawing funds. This could potentially artificially inflate sellout price, which reduces settlement funds available to users.

## New Gotchas

This minter exposes a new gotcha that must be prevented by business practices. Consider the following:
- Core contract project's maxInvocations is set to `200`
- Artist manually limits minter maxInvocations to 100 on DA w/Settlement minter
- Project sells 100 tokens mid-auction, "sells out" above auction base price
- Before withdrawing revenues, artist calls `setProjectMaxInvocations`, updating minter's local maxInvocations to `200`
- The auction now becomes "live", and a step-change downward in purchase price occurs, potentially greatly reducing the revenues available to be withdrawn by the Artist. Additionally, the Artist may no longer withdraw funds until either the original auction reaches base price, or 200 tokens are sold.

The solution to the gotcha described above is to ensure the artist knows to call `withdrawArtistAndAdminRevenues` BEFORE calling `setProjectMaxInvocations` (or `manuallyLimitProjectMaxInvocations`), especially in the case of a sellout above auction base price.

## Performance

Overall, this minter is more gas-performant than the prior version of this minter. This is due to a removal of the previous constraint to sync the minter's `maxHasBeenInvoked` during every purchase. The logic in this updated version is a bit more complex because of the local limiting, but overall, this a more gas performant iteration.

V1 average gas used for mint: `109_429`
V2 average gas used for mint:  `105_503` (-3.5%)

## TODO:

- [ ] add tests for edge cases dealing with new max invocation scenarios (e.g. artist increases maxInvocations after sellout occurs, etc.)


ryley-o:
difchecker of current draft here: https://www.diffchecker.com/MN4DuwwS/

ryley-o:
closing this in favor of the design as iterated in #478

PR #478, opened 2023-02-02
DA w/Settlement V2: Local Max Invocations

ryley-o:
## Description of the change

Add DA Settlement minter that enables minter-local max invocations that may be less than or equal to the project's max invocations defined on the core contract.

This is similar to the functionality recently added to all other minter contracts, but with some added complexity of change on this minter, since "definition of done" is an important state to understand on a minter with on-chain settlement.

## Design Choices

- After initial feedback, this minter reverts when the `setProjectMaxInvocations` function is called. This is done to reduce the chance of operational error by an artist. Instead, an artist must call `manuallyLimitProjectMaxInvocations` if a different max invocations is desired than what is defined on the core contract.
- When initially configuring an auction, if an artist has not already called `manuallyLimitProjectMaxInvocations`, the local cache of max invocations on the minter is updated to be equal to the value on the core contract. This is done to simplify some of the assumptions about the state of the local max invocations on the minter.
- A guard rails is included to make a minter's local `maxHasBeenInvoked` value is a one-way switch from false to true. This prevents artists from unintentionally increasing max invocations on a project that has sold out, which could have resulted in decreased revenue from what would have been a fair auction.

Note that this was the favored design over a seemingly less straightforward implementation in #476.

## Trust Assumptions

Fundamentally, trust assumptions are almost equivalent to the V0 and V1 versions of the DA w/Settlement minter, other than artists may now reduce max invocations to current invocations on core contract to artificially increase sellout price instead of (current invocations + 1). At a high-level, this is a negligible change to the overall trust assumption model because a single invocation is insignificant.

Specifically, note that this minter has the following trust assumption (which was also present in a less-direct way in earlier versions of this minter):
- The artist is trusted to not reduce their project's minter-local maxInvocations mid-auction, inducing a `maxHasBeenInvoked` of true, and immediately withdrawing funds. This could potentially artificially inflate sellout price, which reduces settlement funds available to users who purchased early on in the auction.
There are also

## New Gotchas (minimal)
Guardrails are included in this miner to help artists from unintentionally increasing the minter's local maxInvocations after a sellout. However, artists should still be thoughtful if using this minter in combination with other minters. In general, an artist should withdraw revenues as soon as it becomes possible, and switch to other minters (e.g. fixed price) after a sellout.

## Performance

Overall, this minter is more gas-performant than the prior version of this minter. This is due to a removal of the previous constraint to sync the minter's `maxHasBeenInvoked` during every purchase. The logic in this updated version is a bit more complex because of the local limiting, but overall, this a more gas performant iteration.

V1 average gas used for mint: `109_429`
V2 average gas used for mint:  `105_835` (-3.3%)

## Tests

- Tests are added for coverage of additional logic
- Tests are added for new edge cases dealing with new max invocation scenarios (e.g. artist increases maxInvocations after sellout occurs, etc.)

## Follow-on
Currently, the functions in MinterDAExpSettlementV2 are ordered in a way for optimal diffchecking against V1. After merging this PR, the functions should be re-ordered to align with Solidity's style guide.

---
  - https://app.asana.com/0/0/1203896732830237


ryley-o:
current diff from v1 available here: https://www.diffchecker.com/qgETA4zP/

ryley-o:
current diff da w/settlement v1 to v2 here: https://www.diffchecker.com/5RPvP79i/

jakerockland:
> The solution to the gotcha described above is to ensure the artist knows to call withdrawArtistAndAdminRevenues BEFORE calling setProjectMaxInvocations(_projectId) (or manuallyLimitProjectMaxInvocations with a new, higher number), especially in the case of a sellout above auction base price. The order should go:
> (for auctions utilizing local minter manually set maximum invocations limit)
> 
> Auction sells out or reaches base price on minter
> Artist withdraws revenues (recommend as soon as possible)
> Project max invocations may be increased on the minter if desired, or a different minter may be switched to, etc.

@ryley-o just want to confirm here as I feel that this can be read in a couple ways – it is not the intention in a normal use case to call `setProjectMaxInvocations` after an auction completes already; normal usage should always involve withdrawing funds and moving to a new minter type I would think… there isn't a great reason to do a two stage DA where both stages are the same style, but reasons not to do so (weird pricing experience haha).

Just want to confirm we're on the same page here and I'm not misunderstanding your description to be a suggestion that there _is indeed a reason an artist would intend to do this_ that you have of mind.

contracts/minter-suite/Minters/MinterDAExpSettlementV2.sol
```diff
@@ -0,0 +1,1342 @@
...
+     * maximum invocations value will be set to the core contract's maximum
+     * invocations value, and the auction will be able to be invoked again,
+     * potentially reducing the artist's revenue below sellout price.
+     * @param _projectId Project ID to set the maximum invocations for.
+     * @dev this enables gas reduction after maxInvocations have been reached -
+     * core contracts shall still enforce a maxInvocation check during mint.
+     */
+    function setProjectMaxInvocations(
+        uint256 _projectId
+    ) external onlyArtist(_projectId) {
```
jakerockland, line 240: Given that this method is kinda a footgun, I'm inclined to say that we actually just make it a no-op for this minter.. WDYT @ryley-o ?

Are there weird downstream effects of doing so (breaks out artist UI in some weird way)?

jakerockland:
>  It also keeps a bool useLocalMaxInvocations in the ProjectConfig, clearly defining if the minter is limiting maxInvocations locally, or if the core contract should be limiting maxInvocations. 

Would love to double-click together on this @ryley-o – I find it a bit more confusing and error prone if this minter is modified in the future to maintain this state in this branched way (need to check the boolean every time and then the corresponding int) vs. just maintaining one cached value, and am not sure I understand the benefit of this other than the nice semantics around _knowing which method was called based on the bool_ – also if we go w/ the suggestion in https://github.com/ArtBlocks/artblocks-contracts/pull/478#discussion_r1100889782 then this resolves itself I suppose.

Would love to chat more on this point as it feels like a bigger one re: this PR.

ryley-o:
> @ryley-o just want to confirm here as I feel that this can be read in a couple ways – it is not the intention in a normal use case to call setProjectMaxInvocations after an auction completes already; normal usage should always involve withdrawing funds and moving to a new minter type I would think… there isn't a great reason to do a two stage DA where both stages are the same style, but reasons not to do so (weird pricing experience haha).

>Just want to confirm we're on the same page here and I'm not misunderstanding your description to be a suggestion that there is indeed a reason an artist would intend to do this that you have of mind.

Yeah, we are on the same page here. Basically wanted to point out that I could see an artist wanting to e.g. use a DA w/Settlement for first half (100 of 200) of token sales, then selling the second half of all tokens at whatever the first half sold out or settled to. In that case, I could imagine an artist wanting to just use this minter for the second half of token sales (since it acts like a fixed-price after an artist withdraws funds, but I agree that our guidance should be to use a different minter.

contracts/minter-suite/Minters/MinterDAExpSettlementV2.sol
ryley-o: agreed, and this adds complexity to this minter's interface that isn't really useful for artists. pushed an update that removes this function 🙏

ryley-o:
@jakerockland This is ready for a re-review with the `setProjectMaxInvocations` function not implemented 👍 basically everything in commit 1ac9a99
New diff from LPDA V1 -> V2 here: https://www.diffchecker.com/QcyHG2Nw/

ryley-o:
@jakerockland - friendly reminder that this re-review is ready for you 👍

contracts/minter-suite/Minters/MinterDAExpSettlementV2.sol
```diff
@@ -0,0 +1,1296 @@
...
+            _maxInvocations >= coreInvocations,
+            "Cannot set project max invocations to less than current invocations"
+        );
+        // EFFECTS
+        ProjectConfig storage _projectConfig = projectConfig[_projectId];
+        // update storage with results
+        _projectConfig.maxInvocations = uint24(_maxInvocations);
+        // We need to ensure maxHasBeenInvoked is correctly set after manually setting the
+        // local maxInvocations value.
+        _projectConfig.maxHasBeenInvoked = coreInvocations == _maxInvocations;
```
jakerockland, line 266: > This minter exposes a new gotcha that should be prevented by business practices. Put simply, once revenues are ready to be withdrawn, an artist should call withdrawArtistAndAdminRevenues. After a sellout, it is not recommended that the artist do anything with the local max invocations on the minter. If an auction were to sell out and an artist were to increase the local max invocations before withdrawing revenues, the most reasonable state could be interpreted to be that the auction should continue down to the base price, and the last price paid could drop. The result would be a lower settling price than expected. This would be an operational mistake by the artist, and would reduce the artist's revenues for that project while reducing the settling net-price paid by users.

Could we mitigate this even being a possible footgun, by making this line instead:

```
// Note: as a safety measure, never set a already true value to false, only allowing 
// `_maxInvocations`s functionally to be modified if have not already hit sell out
_projectConfig.maxHasBeenInvoked = (!_projectConfig.maxHasBeenInvoked) && coreInvocations == _maxInvocations;
```

jakerockland (approved):

contracts/minter-suite/Minters/MinterDAExpSettlementV2.sol
jakerockland: I believe this suggestion may have related implications for other invocation syncing methods throughout this minter, if this route seems like a good one in the interest of simplifying possible behavior.

jakerockland:
@ryley-o apologies for the review delay here – thank you very much for the additional ping, much appreciated.

One final suggestion / topic here: https://github.com/ArtBlocks/artblocks-contracts/pull/478#discussion_r1106497217 but otherwise LGTM to merge pending that and move forward with re: next steps around deploying. 

Perhaps we batch the deployment plans for this update with those in https://github.com/ArtBlocks/artblocks-contracts/pull/491

contracts/minter-suite/Minters/MinterDAExpSettlementV2.sol
ryley-o: I quite like the concept of making `_projectConfig.maxHasBeenInvoked` a (almost, see last note below) 1-way switch.

To keep assumptions across the minter simpler, I have kept these lines of code the same, but added a require to this function that `_projectConfig.maxHasBeenInvoked` is false. Along with that require, also seems appropriate to require that the new `_maxInvocations` is gt `0`, which would forever prevent a project from using this minter (since it successful, it will immediately toggle the one-way switch) and be a nonsensical call by an artist.

Note that it is intentionally still possible for the local `maxHasBeenInvoked` to flip from true to false in the extreme edge case where the minter's local `maxHasBeenInvoked` is returning a false negative, which can only happen if max invocations were limited on the minter, the project used a different minter to go above this minter's local max invocations, then attempted to increase the local max invocations on this minter. This is not the footgun we are trying to protect against, and in that edge case, the artist's intentions cannot be easily predicted, so we can defer to whatever the artist calls.

ryley-o: @jakerockland - These changes are implemented in [8475993](https://github.com/ArtBlocks/artblocks-contracts/pull/478/commits/8475993e1767f5e87535255ff3cb8e24dd6887b0) - do you mind resolving this if that looks good to you?

ryley-o:
>One final suggestion / topic here: https://github.com/ArtBlocks/artblocks-contracts/pull/478#discussion_r1106497217 but otherwise LGTM to merge pending that and move forward with re: next steps around deploying.

Great callout, and I've pushed an additional commit to address this

>Perhaps we batch the deployment plans for this update with those in https://github.com/ArtBlocks/artblocks-contracts/pull/491

That seems like a great plan 👍

jakerockland (approved):
Update LGTM 🙏 💯

contracts/minter-suite/Minters/MinterDAExpSettlementV2.sol
jakerockland: Fantastic–LGTM ser.

PR #481, opened 2023-02-07
Reduce size of V3_Engine_Flex & V3_Engine

ryley-o:
Reduce contract bytecode size to <24 kb by doing the following:
- change modifiers to internal functions, preventing duplication of the logic throughout the bytecode
- change constants to non-constants, preventing duplication throughout bytecode
- move OpenZeppelin Strings lib into our contract and minimize bytecode of the portions of the library that are used
- change overly long revert strings (>31 characters) to be <= 31 characters where necessary

In addition to the changes above, we could further reduce this contract's bytecode size by:
- removing the "purge bytecode" logic and calls, since selfdestruct is deprecated (maybe 1 kb?)
- eliminating custom errors (that could use error codes) instead of using revert messages (maybe 1-2 kb?)

(the estimates on potential for further kb reduction above are just guesses, so we should study if we want/need to pursue further reductions.

Note that without shortening some of the error messages, which is the only change that isn't a refactor, the contract is only 0.1 kb over the 24 kb limit. Instead of changing error messages (which adds complexity to our testing suite), we could do something else like remove the purge bytecode logic.

============================================================================================

(BEN L UPDATE, SOW @ 2/7/23 5:30PM PST)

After further discussion async with concerned parties, the optimizations for V3_Engine & V3_Engine_FLEX are now as follows:

- Change modifiers to internal functions, preventing duplication of the logic throughout the bytecode
- Change overly long revert strings (>31 characters) to be <= 31 characters where necessary
- Removing the "purge bytecode" logic and calls, since selfdestruct is deprecated

V3_ENGINE_FLEX contract is now ~23.9KB.
Moving forward we are also planning to add the rule of preferring internal functions over modifiers to our contract's styleguide

### DIFF
--- a/contracts/GenArt721CoreV3_Engine.sol
+++ b/contracts/GenArt721CoreV3_Engine.sol
@@ -273,46 +273,46 @@ contract GenArt721CoreV3_Engine is
     /// default base URI to initialize all new project projectBaseURI values to
     string public defaultBaseURI;
 
-    function onlyNonZeroAddress(address _address) internal pure {
+    function _onlyNonZeroAddress(address _address) internal pure {
         require(_address != address(0), "Must input non-zero address");
     }
 
-    function onlyNonEmptyString(string memory _string) internal pure {
+    function _onlyNonEmptyString(string memory _string) internal pure {
         require(bytes(_string).length != 0, "Must input non-empty string");
     }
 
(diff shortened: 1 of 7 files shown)


ryley-o (commented):
This LGTM @lyaunzbe - only minor gas test failing, fine to fold into the flex branch!

PR #482, opened 2023-02-07
Add Deployed event on MinterFilterV1

ryley-o:
## Description of the change

Add a `Deployed` event to IMinterFilterV0 to improve indexing logic available for V3 Engine contracts

We currently have an issue in our subgraph where we would like to determine if we should index an Engine partner's MinterSuite based on if their MinterFilter contract is in the subgraph store or not. This new event, along with a follow-on handler to add the MinterFilter to the store when the `Deployed` event is observed, will enable that process.

Assuming this is implemented, a follow-on PR will be on subgraph, and we will implement block-number based logic to index all MinterFilters before this change is implemented (our current patched behavior), and will only index MinterFilters in a subgraph's config if on a block after this change is implemented.

## Design Alternatives Considered

We could index all Engine contract minters as MinterFilters (which is our current patched behavior), but we could experience subgraph failures if we index all Engine minter systems by default.
We could change our deployment strategy to require a specific ordering such as a minter must be allowlisted on a MinterFilter before we set the MinterFilter to be the minter on a core contract (which adds the minter to the store due to the emitted event), but that system is not very robust.
This seemed like the cleanest system with minimal effort, especially since it will also work with our future plans of a consolidated minter suite.

---
  - https://app.asana.com/0/0/1203931419658369

### DIFF
--- a/contracts/interfaces/0.8.x/IMinterFilterV0.sol
+++ b/contracts/interfaces/0.8.x/IMinterFilterV0.sol
@@ -5,6 +5,12 @@ pragma solidity ^0.8.0;
 
 interface IMinterFilterV0 {
     /**
+     * @notice Emitted when contract is deployed to notify indexing services
+     * of the new contract deployment.
+     */
+    event Deployed();
+
+    /**
      * @notice Approved minter `_minterAddress`.
      */
     event MinterApproved(address indexed _minterAddress, string _minterType);
--- a/contracts/minter-suite/MinterFilter/MinterFilterV1.sol
+++ b/contracts/minter-suite/MinterFilter/MinterFilterV1.sol
@@ -106,6 +106,7 @@ contract MinterFilterV1 is IMinterFilterV0 {
     ) onlyNonZeroAddress(_genArt721Address) {
         genArt721CoreAddress = _genArt721Address;
         genArtCoreContract = IGenArt721CoreContractV3(_genArt721Address);
+        emit Deployed();
     }
 
     /**
--- a/contracts/minter-suite/MinterFilter/MinterFilterV1.sol
+++ b/contracts/minter-suite/MinterFilter/MinterFilterV1.sol
@@ -6,6 +6,8 @@ import "../../interfaces/0.8.x/IFilteredMinterV0.sol";
 import "../../interfaces/0.8.x/IAdminACLV0.sol";
 import "../../interfaces/0.8.x/IGenArt721CoreContractV3.sol";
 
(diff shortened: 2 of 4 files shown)


ryley-o:
>Also–should we consider bumping from V1 to V2 here? I would think probably not but may make our subgraph logic easier to rationalize about..?

@jakerockland - ah, this is a good idea, however we don't have a public variable that broadcasts the minter filter version (doh!). It would be a great use of minor version if we had that implemented.
I think for now, avoiding more ABIs propagating through our infra is probably more valuable than rolling this version, and block-dependent subgraph logic is reasonable. In the future, we should absolutely be broadcasting a major and minor version of the MinterFilter, which will improve our handler logic.
Interested to hear @yoshiwarab's thoughts on this as well, definitely no super strong opinions from me.

jakerockland:
> > Also–should we consider bumping from V1 to V2 here? I would think probably not but may make our subgraph logic easier to rationalize about..?
> 
> @jakerockland - ah, this is a good idea, however we don't have a public variable that broadcasts the minter filter version (doh!). It would be a great use of minor version if we had that implemented. I think for now, avoiding more ABIs propagating through our infra is probably more valuable than rolling this version, and block-dependent subgraph logic is reasonable. In the future, we should absolutely be broadcasting a major and minor version of the MinterFilter, which will improve our handler logic. Interested to hear @yoshiwarab's thoughts on this as well, definitely no super strong opinions from me.

That definitely makes sense to me @ryley-o, thank you for considering–something to follow up on down the road!

Would love to get hear @yoshiwarab's thought as well.

ryley-o:
> @ryley-o this sounds good to me. I think instead of trying to track a specific block number, what do you think about adding the public version field to this minor version and doing a try catch in the subgraph handler assuming that if that field doesn't exist that it's the earlier version and should be indexed?

@yoshiwarab - That would be nice, but an issue I see with that is if, for example, a partner wants to not use our MinterFilter abstraction, our subgraph would default to indexing the unknown minter as a MinterFilter (because the try/catch would fail).

That being said, I think the logic could go something like:
- if before block X, index the minter filter (we know this history, and we want to index those minter filters)
- if after block X, determine if the minter filter should be indexed by determining if it is in store or not. Additionally, determine the MinerFilter's version based on it's exposed public version (and not indexing if it doesn't respond/reverts).

Thus, I'll update this PR to include a pubic version on the MinterFilter, because for future upgrades that occur after block X, we want to be able to NOT index something if it doesn't respond with a Minter Filter version. 

does that sound like a good idea?

PR #484, opened 2023-02-08
V3 Engine deployer script

ryley-o:
## Description of the change

Create generic V3 Engine deployer script that handles deployments via input config file and auto-records deployment details.

Adds following capabilities:
- define a config json file format, so we have a single script to deploy any engine product suite
- try/catch when verifying contracts
- auto-generate deployments markdown file (save in same dir as config json)
- auto-record all deployment stout in log files (filtering nuisance errors)
- keep with new /deployments directory, properly dividing responsibilities of scripts and deployment record keeping

## Design Decisions
- `prompt` library was used to request config file path from user because Hardhat scripts do not play well with varg inputs
  - Hardhat "tasks" do accept standard inputs, however they have many other issues/quirks (e.g. global variables, break to a different format than our tests) that are not preferred

## Testing/Validation
- This script was confirmed to be working on goerli, generating outputs as desired.
  
## Follow-on
A follow-on PR to auto-configure Hasura may be created, which would populate the contract bucket name, contract name, but not type (since that should be implemented on sync server)

---
  - https://app.asana.com/0/0/1203760090368803


deployments/engine/V3/partners/dev-example/deployment-config.dev.ts
```diff
@@ -0,0 +1,45 @@
+// This file is used to configure the deployment of the Engine Partner contracts
+// It is intended to be imported by the generic deployer in `/scripts/engine/V3/generic-v3-engine-deployer.ts`
+export const deployDetailsArray = [
+  {
+    network: "goerli",
+    // environment is only used for metadata purposes, and is not used in the deployment process
+    // Please set to "dev", "staging", or "mainnet", as appropriate
+    environment: "dev",
+    // if you want to use an existing admin ACL, set the address here (otherwise set as undefined)
```
jakerockland, line 9: Nit: "otherwise set as undefined to deploy a new one"

scripts/engine/V3/generic-engine-deployer.ts
```diff
@@ -0,0 +1,541 @@
...
+ * deployment by the configured superAdmin.
+ */
+async function main() {
+  prompt.start();
+  const deplomentConfigFile = (
+    await prompt.get<{ from: string }>(["deployment config file"])
+  )["deployment config file"];
+  // dynamically import input deployment configuration details
+  let deployDetailsArray;
+  const fullImportPath = path.join("../../../", deplomentConfigFile);
```
jakerockland, line 45: I'm not sure if it makes sense to do this vs. just expecting an absolute path – my inclination would be towards the former but would love to better understand your thinking here @ryley-o

scripts/engine/V3/generic-engine-deployer.ts
```diff
@@ -0,0 +1,541 @@
...
+  const inputFileDirectory = path.dirname(fullImportPath);
+  try {
+    ({ deployDetailsArray } = await import(fullImportPath));
+  } catch (error) {
+    throw new Error(
+      `[ERROR] Unable to import deployment configuration file: ${deplomentConfigFile}
```
jakerockland, line 51: If we append to the path, I think we should probably output the full import path here in our error.

scripts/engine/V3/generic-engine-deployer.ts
```diff
@@ -0,0 +1,541 @@
...
+    //////////////////////////////////////////////////////////////////////////////
+
+    //////////////////////////////////////////////////////////////////////////////
+    // VERIFICATION BEGINS HERE
+    //////////////////////////////////////////////////////////////////////////////
+
+    // Output instructions for manual Etherscan verification.
+    const standardVerify = "yarn hardhat verify";
+    // CORE CONTRACT
+    // @dev do not use helper tryVerify here due to complex constructor args
```
jakerockland, line 368: Very fair 😄

jakerockland (approved):
This is awesome @ryley-o – super exciting. Thank you!!! 🔥 🙏 💜

deployments/engine/V3/partners/dev-example/deployment-config.dev.ts
ryley-o: updated 👍

scripts/engine/V3/generic-engine-deployer.ts
ryley-o: good call, updated!

scripts/engine/V3/generic-engine-deployer.ts
ryley-o: good call, updated this to use full import path!

PR #485, opened 2023-02-08
initial version

ryley-o:
## Description of the change

\<add description here\>


ryley-o:
@yoshiwarab - tagged you for a review on this one since it contains a decent amount of graphql 🙏

deployments/engine/V3/internal-testing/dev-example/deployment-config.dev.ts
```diff
@@ -41,5 +41,11 @@ export const deployDetailsArray = [
     // set to true if you want to add an initial token to the initial project
     // (this will only work if you have set addInitialProject to true, and requires a MinterSetPriceV[4-9])
     addInitialToken: true,
+    // optionally define this to set default vertical name for the contract after deployment.
+    // if not defined, the default vertical name will be "unassigned".
+    // common values include `fullyonchain`, `flex`, or partnerships like `artblocksxpace`.
+    // note that if initial project is added, the vertical name of that specific project
```
jakerockland, line 47: Is this in the case of initial project being added via deployment, due to the race condition in that case? IIUC if that is so, would clarify this more explicitly inline, or perhaps there is another case here this intends to reflect?

jakerockland (approved):
Woooo!! 

One question here, does this capture the contract type being properly set to Engine though for Engine contracts? I don't believe so, but maybe that is auto-resolved elsewhere for V3..

So exciting. 

This LGTM, but will defer to @yoshiwarab on the review of the more graphQL heavy components here as my fluency there is definitely more cursory.

deployments/engine/V3/internal-testing/dev-example/deployment-config.dev.ts
ryley-o: Ah, I forgot to update this comment - the initial project is actually updated now (added another graphql update), so I've removed this stale comment.

I also made sure to include the note about if new a new vertical name is needed, it must be added manually/by migration. This is because there are some other details needed when defining a new vertical in Hasura.

ryley-o:
>One question here, does this capture the contract type being properly set to Engine though for Engine contracts? I don't believe so, but maybe that is auto-resolved elsewhere for V3..

@jakerockland This process doesn't update contract type (since it should be synced from our subgraph), but that detail is resolved in https://github.com/ArtBlocks/artblocks/pull/479 (including properly handling V3_Engine types) 👍

jakerockland:
> > One question here, does this capture the contract type being properly set to Engine though for Engine contracts? I don't believe so, but maybe that is auto-resolved elsewhere for V3..
> 
> @jakerockland This process doesn't update contract type (since it should be synced from our subgraph), but that detail is resolved in [ArtBlocks/artblocks#479](https://github.com/ArtBlocks/artblocks/pull/479) (including properly handling V3_Engine types) 👍

Fantastic – TYVM for confirming!

PR #488, opened 2023-02-11
Update Style Guidelines

### LINKED ISSUE #483: Update styleguide to mention non-use of modifier functions & preform cleanup
result of #481 

ryley-o:
## Description of the change

Refactor and update style guidelines to reflect preference to not use Solidity modifiers.

This partially completes #483

---
  - https://app.asana.com/0/0/1203940173166169

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -1,0 +1,15 @@
+# Style Guidelines
+
+> Please read through our [Solidity Gotchas](./solidity-gotchas.md) and [Testing Philosophy](./test/README.md) sections before contributing.
+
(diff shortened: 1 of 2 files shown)


jakerockland (approved):
This is fantastic – thank you @ryley-o 

One note, lets wait to fully close #483 until we've also done the cleanup of usage of existing modifiers (can also split the issue into two if you'd prefer).

ryley-o:
sounds good - the remaining work of removing modifiers from all active contracts has been tagged as a polish week task 🙏 thanks for calling that out!

PR #491, opened 2023-02-13
Minor Max Invocation Patch

ryley-o:
## Description of the change

Patches our minters with local max invocations to handle an edge case that should never be encountered on mainnet.

Specifically, the minters now revert a purchase when invocations on a project are >= max invocations on the minter. 

Tests were updated and drove the development in this PR. 

Also note that for convenience and consistency, a similar design choice as #478 was implemented. When a price/auction is configured, if a project's minter-local max invocations have never been populated, they are automatically synced to the values on the core contract. This removes an extra call that should be performed by most artists to ensure gas efficiency, but is often missed.


jakerockland (approved):
This LGTM ser – thank you very much!

PR #498, opened 2023-02-23
Refactor tests to use Hardhat fixtures

ryley-o:
## Description of the change

Refactor entire test suite to use Hardhat fixtures to reduce time to run tests by >80%.

fixes #371, implements solution based on investigation from #448.

Hardhat fixtures are the recommended way to implement tests due to increased performance. This repository's time to run all tests is very long (~10 minutes, even when ran in parallel in CI), so switching to use Hardhat fixtures will benefit our development process.

Time for CI to run tests (parallel, currently 8 runners):
- before: `~17 minutes` (ref #497 circleci test)
- after: ~3 minutes (ref this PRs circleci test)

Also note that the `yarn test` command can run tests in ~4  minutes on my local macbook (tests are not ran in parallel).

## Design choices

In general, tests were kept similar to our current organization patterns. Fixtures are generally named `_beforeEach`, and return an object named `config`. `config` is used similarly to our previous use of `this`.

Using fixtures has the drawback of not conforming to the more often used mocha `beforeEach` pattern. However, the large performance benefit justifies its use in this repo.

## Note on Coverage

Coveralls is reporting a decrease of 0.05% coverage on the core V3 contract, but I don't know why. No tests should have been removed during this refactoring, but less runs of "beforeEach" are occurring since the fixture only gets ran a single time. Note that we still have full line coverage on the V3 core contract, so I am fine with merging this as-is, especially since I'm wondering if the blip has something to do with how coveralls is comparing this PR's latest commit to main branch...

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203665500354310
  - https://app.asana.com/0/0/1203760090368799

PR #499, opened 2023-02-23
Remove modifiers from all active contracts

### LINKED ISSUE #483: Update styleguide to mention non-use of modifier functions & preform cleanup

ryley-o:
## Description of the change

Remove modifiers from all active contracts.

This fixes #483 

Note that gas impacts are very minimal (maybe ~30 gas total impact), and gas tests will be updated after #498 is merged into main.


ryley-o:
closing in favor of #500

PR #500, opened 2023-02-23
Remove modifiers from all active contracts.

### LINKED ISSUE #483: Update styleguide to mention non-use of modifier functions & preform cleanup

ryley-o:
## Description of the change

Remove modifiers from all active contracts.

This fixes #483 

Benefits are reduced bytecode sizes for contracts, with negligible gas impacts.

## Other minor changes

- Gas tests for each minter are now performed only on flagship contract (since the are only intended to track changes in gas due to minter changes, and testing on one core is sufficient)
- V3_Explorations picked up a couple minor comment updates to align with V3 flagship contract
- V3 flagship and V3_Explorations have their versions bumped to v3.0.2 and v3.0.3, respectively
  - V3 Flagship changelog updated

---
  - https://app.asana.com/0/0/1203940173166170


jakerockland (approved):
Thank you very much @ryley-o

PR #501, opened 2023-02-23
Reduce minter gas test nuisance failures

ryley-o:
## Description of the change

Update minter gas tests to reduce nuisance failures.

Gas tests for minters are updated to:

- only test gas cost of minting with one core contract
  - this is because the gas tests are only meant to alert for changes, and testing with a single core is sufficient to do this
- only fail if expected gas is >1% different than target
  - we shouldn't receive nuisance warnings about negligible changes

---
  - https://app.asana.com/0/0/1203960042742717

### DIFF
--- a/test/util/common.ts
+++ b/test/util/common.ts
@@ -290,14 +290,21 @@ export async function advanceEVMByTime(_timeSeconds: number) {
 }
 
 // utility funciton to compare Big Numbers, expecting them to be within x%, +/-
-export function compareBN(
+export function requireBigNumberIsClose(
   actual: BigNumber,
   expected: BigNumber,
   tolerancePercent: number = 1
-): boolean {
+) {
   const diff = actual.sub(expected);
   const percentDiff = diff.mul(BigNumber.from("100")).div(expected);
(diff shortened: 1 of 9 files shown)


test/util/common.ts
```diff
@@ -290,14 +290,21 @@ export async function advanceEVMByTime(_timeSeconds: number) {
 }
 
 // utility funciton to compare Big Numbers, expecting them to be within x%, +/-
-export function compareBN(
+export function requireBigNumberIsClose(
   actual: BigNumber,
   expected: BigNumber,
   tolerancePercent: number = 1
```
jakerockland, line 296: Nit: Would suggest that we either do not have a default value here at all, or if we do keep this default value (my preference to keep, and be `1`) that we don't also pass in `1` in all the call-sites above but instead make use of the default.

jakerockland (approved):

test/util/common.ts
ryley-o: good call, updated to not pass in `1` in all the call-sites, and kept default value of `1` in commit 3159566

PR #503, opened 2023-02-25
Refactor hash generation to base class

ryley-o:
## Description of the change

Move pseudorandom hash generation to a base class to improve separation of changes.

Note that a base class is used instead of a library to enable use of randomizer state during pseudorandom hash generation.

---
  - https://app.asana.com/0/0/1204024759451481


jakerockland (approved):
🤌 🤌 🤌

PR #504, opened 2023-02-28
Serial English Auction (nouns-style) minter

### LINKED ISSUE #185: Nouns Style Minter
nouns style auction for an entire project, ie artist does a small mint size collection like 32 mints, each mint auctions for 15min and the visual outcome is visible during that time period, so it’s like an 8hr event essentially. params would be configurable 

[add more...]

ryley-o:
## Description of the change

Add a serial English auction ([nouns-style](https://nouns.wtf/)) minter.

Fixes #185 

## Overview

This minter provides an option to mint and distribute a project's tokens one at a time via consecutive, on-chain English auctions. 

## Tests

- Tests are added to achieve write tests for full coverage + integration tests of this minter
- note that full branch coverage is achieved except for the case of two of three `nonReentrant` modifiers. This was because the two untested `nonReentrant` modifiers are only providing dual redundancy, and are not our primary layer of defense.

---
  - https://app.asana.com/0/0/1203665500354290


ryley-o:
@jakerockland - this is ready for review. Would appreciate thorough feedback before writing all of the tests 🙏 Also note that our internal design document has been created and is ready for review 👍

ryley-o:
@jakerockland @lyaunzbe - this is ready for a complete review, tests have been fully implemented 🙏

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,987 @@
...
+        require(
+            _maxAuctionDurationSeconds > _minAuctionDurationSeconds,
+            "Only max gt min"
+        );
+        require(_minAuctionDurationSeconds > 0, "Only min gt 0");
+        // EFFECTS
+        minAuctionDurationSeconds = _minAuctionDurationSeconds.toUint32();
+        maxAuctionDurationSeconds = _maxAuctionDurationSeconds.toUint32();
+        emit AuctionDurationSecondsRangeUpdated(
+            _minAuctionDurationSeconds.toUint32(),
```
lyaunzbe, line 315: nit: potentially worth just passing in `minAuctionDurationSeconds` & `maxAuctionDurationSeconds` to avoid extra `toUint32` function calls?

lyaunzbe: i guess i'm also curious why you can't just intake uint32 (instead of uint256) here for the external function arguments, given that you do that below with `updateMinterTimeBufferSeconds`

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,987 @@
...
+        // update auction state
+        _auction.currentBid = msg.value;
+        _auction.currentBidder = payable(msg.sender);
+        uint256 minEndTime = block.timestamp + minterTimeBufferSeconds;
+        if (_auction.endTime < minEndTime) {
+            _auction.endTime = minEndTime.toUint64();
+        }
+
+        // INTERACTIONS
+        // refund previous highest bidder
```
lyaunzbe, line 615: more of a design question and definitely not blocking: were user claim/settlement style bid refunds considered? Aside from reducing gas cost of this bidding, this would also potentially allow for an implementation where someone that is outbid can add create a new active bid by adding funds to a previous bid that was outbid but has not been claimed and is still being held by the contract. Totally understand if the thinking here is that the risk of holding funds outweighs the potential mentioned benefits though.

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,987 @@
...
+        (
+            uint256 invocations,
+            uint256 maxInvocations,
+            ,
+            ,
+            ,
+
+        ) = genArtCoreContract_Base.projectStateData(_projectId);
+        // revert if next invocation would exceed max invocations
+        require(
```
lyaunzbe, line 800: would this be ignoring a manually set max project invocations limit?

lyaunzbe (approved):
left some super minor nits/questions. very awesome overall, great inline documentation made this nice to read and the tests are super thorough 👍 👍 

I also wonder if some high level diagram/overview would be useful for this minter as it has a good bit more complexity compared to most of the other minters in the suite. This also may make sense to go in our actual hosted docs vs github though

contracts/minter-suite/Minters/MinterSEAV0.sol
jakerockland: This is a very interesting design consideration. I do think though that the primary value proposition in this case though would just be gas costs, no?

> this would also potentially allow for an implementation where someone that is outbid can add create a new active bid by adding funds to a previous bid that was outbid but has not been claimed and is still being held by the contract. 

Put differently, what is the benefit of this itself other than the gas savings angle? The rebidding flow is still there as an option either way, and needs to have explicit user action associated with it.

lyaunzbe: agree,  practically speaking it all boils down to gas cost savings. The rebidding flow is there regardless and explicit regardess, but with this design the rebidding flow maybe has a different psychological user experience (adding a small amount to get back "in the game" vs re-adding the entire amount from your previous bid + amount to make your new bid valid)

ryley-o: Good discussion topic here.

Re: gas efficiency, I think the current implementation is actually slightly more globally gas-efficient, because it will require at least one less tx per auction (the second-place bidder doesn't have to reclaim their ETH at the end of an auction), and each tx has a base cost of 21000 gas.

I think probably more important than gas efficiency is UX. While it can be debated that a pull mechanism is sometimes preferred relative to a push, I think following the nouns team's implementation of a push mechanism is slightly more user friendly for this type of auction. The single biggest benefit of immediately refunding the losing bidder is that it eliminates the need for a user to worry about the question "Do I have any funds to go reclaim in Art Blocks auctions?". It also eliminates the need for us to design and implement an entire notification system for a user to communicate which project(s) they have funds available to reclaim on.

These are just my instincts though, and definitely consider this topic worth discussing further across teams!

lyaunzbe: Ah ya good points on the global gas efficiency + extra work around notification system 👍 

Though global gas efficiency aside, I would say it's potentially nicer if the refund gas costs are associated/billed to the bidder receiving the refund vs a different bidder entirely. That being said, the points mentioned above are pretty compelling to the side of it being not worth it

ryley-o:
important update added in [65b3f12](https://github.com/ArtBlocks/artblocks-contracts/pull/504/commits/65b3f12036fb4c930fb893aba7f3933ed20fd0ce) to prevent gas cost/DoS attack.

Cc @lyaunzbe @jakerockland

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: >also just feels not worth it for locking funds

This is a very good point 😅

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: good call - added additional check for minter max invocations in 8684a99 🙏

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: absolutely can do this 🙏 good call, updated in 8389aa1

ryley-o:
Folded in PR #546 to this PR 🙏

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,1186 @@
...
+    address public immutable minterFilterAddress;
+
+    /// Minter filter this minter may interact with.
+    IMinterFilterV0 private immutable minterFilter;
+
+    /// minterType for this minter
+    string public constant minterType = "MinterSEAV0";
+
+    /// minter version for this minter
+    string public constant minterVersion = "v0.0.0";
```
jakerockland, line 105: Nit: should we start this at `v0.0.1` or `v1.0.0`

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,1186 @@
...
+            _maxAuctionDurationSeconds > _minAuctionDurationSeconds,
+            "Only max gt min"
+        );
+        require(_minAuctionDurationSeconds > 0, "Only min gt 0");
+        // EFFECTS
+        minAuctionDurationSeconds = _minAuctionDurationSeconds;
+        maxAuctionDurationSeconds = _maxAuctionDurationSeconds;
+        emit AuctionDurationSecondsRangeUpdated(
+            _minAuctionDurationSeconds,
+            _minAuctionDurationSeconds
```
jakerockland, line 348: Suggested fix for copypaste error.

```suggestion
            _maxAuctionDurationSeconds
```

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,1186 @@
...
+        uint32 _minAuctionDurationSeconds,
+        uint32 _maxAuctionDurationSeconds
+    ) external {
+        _onlyCoreAdminACL(this.updateAllowableAuctionDurationSeconds.selector);
+        // CHECKS
+        require(
+            _maxAuctionDurationSeconds > _minAuctionDurationSeconds,
+            "Only max gt min"
+        );
+        require(_minAuctionDurationSeconds > 0, "Only min gt 0");
```
jakerockland, line 342: Nit: we do a lot of >0 checks, wonder if we should make them into a helper `_onlyNonZero` method?

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,1186 @@
...
+        // CHECKS
+        // @dev this check is not strictly necessary, but is included for
+        // clear error messaging
+        require(_auction.initialized, "Auction not initialized");
+        if (_auction.settled || (_auction.tokenId != _tokenId)) {
+            // auction already settled or is for a different token ID, so
+            // return early and do not modify state
+            return;
+        }
+        require(block.timestamp > _auction.endTime, "Auction not yet ended");
```
jakerockland, line 632: Nit: Would add a comment here clarifying that the order of these "require"-->"check for return early cases"-->"require" checks is important – as someone could accidentally move the ordering of these to consolidate requires and then return-early-checks and it would modify the intended behavior here.

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,1186 @@
...
+     *   auction is initialized
+     * - sufficiently greater than the current highest bid, according to the
+     *   minter's bid increment percentage `minterMinBidIncrementPercentage`,
+     *   if an auction for the token already exists
+     * If the bid is unsuccessful, the transaction will revert.
+     * If the bid is successful, but outbid by another bid before the auction
+     * ends, the funds will be noncustodially returned to the bidder's address,
+     * `msg.sender`. A fallback method of sending funds back to the bidder via
+     * WETH is used if the bidder address is not accepting ETH (preventing
+     * denial of service attacks) within a 30_000 gas limit.
```
jakerockland, line 668: Any reason we would want to consider to make this gas limit configurable for future-proofing purposes?

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,1186 @@
...
+     *   - project has a current active auction that is not settled
+     *   - insufficient bid amount (msg.value < basePrice)
+     *   - no next token has been minted for the project (artist may need to
+     *     call `tryPopulateNextToken`)
+     *   - `_targetTokenId` does not match the next token ID for the project
+     * This function attempts to mint a new token and assign it to the
+     * project's next token slot. However, if the project has reached its
+     * maximum invocations on either the core contract or minter, this function
+     * will not mint a new token, and the next token slot for the project will
+     * remain empty.
```
jakerockland, line 964: IIUC in this case we still revert though, no?

contracts/minter-suite/Minters/MinterSEAV0.sol
jakerockland: I know we had talked about this more synchronously since this async convo, but upon more conversation definitely understand how the variation suggested here could have a difference in perceptual effect when interacting with this minter. That said, for our initial MVP of this minter I don't think the additional complexity that this would introduce is warranted given that the potential perceptual benefits are definitely hypothetical and could always be introduced via an iterative variant if we had signal that there was demand to warrant this complexity in the future.

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,1186 @@
...
+     * allowed on this minter
+     * @return timestampStart The project's start timestamp, after which new
+     * auctions may be created (one at a time)
+     * @return auctionDurationSeconds The project's default auction duration,
+     * before any extensions due to buffer time
+     * @return basePrice The project's minimum starting bid price
+     * @return nextTokenNumberIsPopulated Whether or not the project's next
+     * token number has been populated
+     * @return nextTokenNumber The project's next token number to be auctioned,
+     * dummy value of 0 if `nextTokenNumberIsPopulated` is false
```
jakerockland, line 849: Nit: 

`0` may not be the best dummy value here, as could be a real value for a new Engine project starting w/ project 0...

That said, it is probably OK if this caveat is just noted as a quirk of this function in the non-initialized state via dev comment, as I'm not sure what a better dummy value would be...

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -0,0 +1,1186 @@
...
+            return _auction.tokenId;
+        }
+        // otherwise, return the next expected token ID to be auctioned
+        if (!_projectConfig.nextTokenNumberIsPopulated) {
+            revert("Next token number not populated");
+        }
+        // @dev overflow automatically checked in Solidity ^0.8.0
+        uint256 nextTokenId = (_projectId * ONE_MILLION) +
+            _projectConfig.nextTokenNumber;
+        return nextTokenId;
```
jakerockland, line 925: Nit: For code-readability purposes, I'm tempted to just return value of `getNextTokenId` here instead.

jakerockland (approved):
So exciting – excellent work here @ryley-o!

Handful of minor nits, but generally LGTM! 🙌 🔥

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: thank you, updated!

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: I like this idea. I broke it out into two functions since I think we kind of have two types of common non-zero tests in this minter:
- a modifier-like `_onlyNonZero` function that may be used to require function inputs be non-zero
- an informative `_projectIsConfigured` internal method, that checks if a project is configured via checking `projectConfig.basePrice`
These updates were pushed in commit [12dfb02](https://github.com/ArtBlocks/artblocks-contracts/pull/504/commits/12dfb0208541f9a40f3483d1b4ae727f8860c20f)

jakerockland: Awesome – thank you! LGTM.

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: ⚠️ code smell since this wasn't caught in tests. Realized that quite a few events needed to be added to this minter's tests, so I've patch this and updated our tests to ensure all event values are as-expected in [bd8520b](https://github.com/ArtBlocks/artblocks-contracts/pull/504/commits/bd8520b582adf0709a3c23720767d36e4e6d5d63)

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: good call, updated with dev comment in commit [f322346](https://github.com/ArtBlocks/artblocks-contracts/pull/504/commits/f3223464d420c5224ae4726e28074b619cf38f84)

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: agree that `0` could certainly be a real token's ID. I think `0` is probably the most ideal dummy value here, since it is Solidity's default value (and most gas-efficient value to return).
Definitely a quirk though 😅

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: Love this, updated to do this in commit [4efdf77](https://github.com/ArtBlocks/artblocks-contracts/pull/504/commits/4efdf773a1d4054e5cdb8512cc8a6c4c81e3b71d)

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: I've updated these comments to help clarify in commit [fce0456](https://github.com/ArtBlocks/artblocks-contracts/pull/504/commits/fce0456f802873a234c21bccecfeaded7a163556).

This is trying to say that _after_ a new auction is created for the token in the `nextToken` slot, the function will attempt to re-fill the `nextToken` slot with a new token (by minting a new token). It will not re-fill the `nextToken` slot if the project has reached max invocations (since it needs to mint a new token to do so), but the transaction won't revert (since we still want to create an auction).

We do revert if this is called and there is no token in the `nextToken` slot at the starting state. However, the `nextToken` slot is automatically filled when an artist configures their project, adjusts max invocations, etc., so we don't expect that case to be encountered (in fact, I believe it is unreachable).

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: I think this is reasonable to be an admin-configurable value (if even for improved cross-chain compatibility).
The value is easily packed with other minter-configuration values also accessed during a new bid, so the gas impacts are minimal.
I've updated in commit [98394da](https://github.com/ArtBlocks/artblocks-contracts/pull/504/commits/98394da2c3fdce41e609d2a677a683ca85081e82) with a solution that emits an event and includes the value in our `minterConfigurationDetails` view function.
I would recommend that we don't bother indexing this value in our subgraph, just because it would be seldomly changed. Thoughts?

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -393,6 +397,27 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
...
+     * @dev max gas limit is 63,535, which is considered a future-safe upper
+     * bound.
+     * @param _minterRefundGasLimit Gas limit to set for refunds. Must be between
+     * 5,000 and max uint16 (63,535).
+     */
+    function updateRefundGasLimit(uint16 _minterRefundGasLimit) external {
+        _onlyCoreAdminACL(this.updateRefundGasLimit.selector);
+        // CHECKS
+        // @dev max gas limit implicitly checked by using uint16 input arg
+        require(_minterRefundGasLimit >= 5000, "Only gte 5_000");
```
jakerockland, line 415: Why is this 5k, vs. e.g. 21k?

jakerockland (approved):

contracts/minter-suite/Minters/MinterSEAV0.sol
jakerockland: Fantastic – that rationale SGTM, I don't think we need to index this info right now.

One minor follow up comment on min value being checked, but can follow up on that in the separate comment thread where I flagged it – resolving!

contracts/minter-suite/Minters/MinterSEAV0.sol
jakerockland: Roger roger – I think worth just noting in dev comment that `0` could be a real value so shouldn't be relied on as way to check false-y/dummy value state (though that this _is_ the dummy value).

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: This is mostly to prevent a footgun when an admin is configuring, so was trying to just impose limits that seemed somewhat reasonable.

The `5_000` gas limit is a nice round number that should still support transferring to most smart contract wallets. Solidity's `.transfer` function imposes a gas limit of `2_300`, but as we know, that isn't enough gas for sending ETH to things like a gnosis safe after London hardfork. With `5_000` gas, transfers should still be successful to a gnosis safe, but receive logic would still be pretty limited. `21_000` seems like it would be quite a lot for a lower bound (although it might be an appropriate value for an admin to configure as the actual refund gas limit).

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: good call, updated the comment!

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: @jakerockland - I've simplified the code that requires next token number is populated in commit [5b9eeaa](https://github.com/ArtBlocks/artblocks-contracts/pull/504/commits/5b9eeaab355a886868e9c5b1a91d3fd9874fba70).

Thanks for the quick chat today!

contracts/minter-suite/Minters/MinterSEAV0.sol
jakerockland: Makes sense ser. 21k from being standard for EOA initiated transfers IIUC, but perhaps that is also bad limit here.

5k does feel somewhat random/arbitrary, so wonder if we should anchor lower (e.g. `2600` based on https://github.com/wolflo/evm-opcodes/blob/main/gas.md#aa-call-operations)  – where is `2300` from?

ryley-o: @jakerockland  good point, 5k is too random. but 2600 is just the base cost of accessing the cold address, so things would fail there too.

I think at a bare-minimum, we would want to at least not remove support for gnosis safes.

Gnosis safes emit an event upon receiving ETH (ref https://github.com/safe-global/safe-contracts/blob/cb22537c89ea4187f4ad141ab2e1abf15b27416b/contracts/common/NativeCurrencyPaymentFallback.sol#L16)

Here is a transaction where someone sent our very own Gnosis safe 0.21 ETH in Nov '22: https://etherscan.io/tx/0x39fb07b63552d9c62b369060509f1b76b412311e7398e06ecf691481d8b2add7

That transaction used `27_329` gas. Subtracting off base tx `21_000` gas means `6_329` gas was used to access the cold address, execute the receive function, and emit the event. The gas used to emit an event is variable and can be based on the value sent, so I would recommend we set the minimum to something like `7_000` after looking at this example.

I've updated min limit to be 7_000, and added a dev comment for logic behind the number, in commit [a9b9af0](https://github.com/ArtBlocks/artblocks-contracts/pull/504/commits/a9b9af0be0fc7b5e36521d4a01b8738fc56bee66)

jakerockland: Rad – that SGTM ser. Thank you for the thoughtful responses here!

jakerockland:
:shipit: ?

PR #513, opened 2023-03-08
Simplify bidding interface of SEA minter

ryley-o:
## Description of the change

This is an iteration on #504 (that may be folded into that PR) that simplifies the bidding interface of the SEA minter by abstracting the concept of "initializing" an auction.

I think this is a meaningful improvement relative to the implementation in #504 because it simplifies the collector actions:

**with this change:**
- `Settle`
- `Bid`

**without this change**
- `Settle`
- `Initialize`(which would bid if already initialized for protection against front-running)
- `Bid`

Variable and function names are updated as part of this change.

## downstream effects

Elected to keep this as a PR to fold into #504, and will update design document if this is LGTM'd 🙏 

This has no downstream effects in indexing/database, other than how the frontend purchase flow will interact with the minter (which has not been built yet), and our design document'

# Tests
Tests are updated for this change. We have a redundant check that cannot achieve branch coverage, which I prefer to keep to avoid introducing bugs in future changes. Unfortunately there isn't a "solcover skip next line" command, so coveralls yells about the new missed branch.

### DIFF
--- a/contracts/interfaces/0.8.x/IFilteredMinterSEAV0.sol
+++ b/contracts/interfaces/0.8.x/IFilteredMinterSEAV0.sol
@@ -88,7 +88,7 @@ interface IFilteredMinterSEAV0 is IFilteredMinterV2 {
 
     function settleAndCreateBid(
         uint256 _settleTokenId,
-        uint256 _initializeTokenId
+        uint256 _bidTokenId
     ) external payable;
 
     function settleAuction(uint256 _tokenId) external;
--- a/contracts/interfaces/0.8.x/IFilteredMinterSEAV0.sol
+++ b/contracts/interfaces/0.8.x/IFilteredMinterSEAV0.sol
@@ -86,14 +86,12 @@ interface IFilteredMinterSEAV0 is IFilteredMinterV2 {
 
     function resetAuctionDetails(uint256 _projectId) external;
 
-    function settleAndInitializeAuction(
+    function settleAndCreateBid(
         uint256 _settleTokenId,
         uint256 _initializeTokenId
     ) external payable;
 
     function settleAuction(uint256 _tokenId) external;
-
-    function initializeAuction(uint256 _targetTokenId) external payable;
 
     function createBid(uint256 _tokenId) external payable;
 
@@ -126,7 +124,5 @@ interface IFilteredMinterSEAV0 is IFilteredMinterV2 {
         uint256 _projectId
     ) external view returns (Auction memory);
 
-    function getTokenToBidOrInitialize(
-        uint256 _projectId
-    ) external view returns (uint256);
+    function getTokenToBid(uint256 _projectId) external view returns (uint256);
 }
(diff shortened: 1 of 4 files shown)


contracts/interfaces/0.8.x/IFilteredMinterSEAV0.sol
```diff
@@ -86,15 +86,13 @@ interface IFilteredMinterSEAV0 is IFilteredMinterV2 {
 
     function resetAuctionDetails(uint256 _projectId) external;
 
-    function settleAndInitializeAuction(
+    function settleAndCreateBid(
         uint256 _settleTokenId,
         uint256 _initializeTokenId
```
jakerockland, line 91: Should be named `_bidTokenId` I believe.

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -450,30 +450,32 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
...
     /**
      * @notice Settles any complete auction for token `_settleTokenId` (if
-     * applicable), then initializes an auction for token `_initializeTokenId`
-     * with bid amount and bidder address equal to `msg.value` and
-     * `msg.sender`, respectively.
+     * applicable), then attempts to create a bid for token
+     * `_initializeTokenId` with bid amount and bidder address equal to
+     * `msg.value` and `msg.sender`, respectively.
+     * Intended to gracefully handle the case where a user is front-run by
+     * one or more transactions to settle and/or initialize a new auction.
```
jakerockland, line 457: How exactly does this more gracefully handle that case? In the case of front run, one's bid is still going to be too low even if it is an "initialize or bid" type logic vs. just "bid" and "initialize" being separate.

If this doesn't actually more gracefully handle the case, are the changes in this PR still preferable?

contracts/interfaces/0.8.x/IFilteredMinterSEAV0.sol
ryley-o: good call, updated!

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: I've added some additional language here to help clarify.

Regarding handling front-runs, this PR doesn't actually change behaviors of the minter. In either this implementation or in #504, if a user's initial bid is front-run by a different bid, and the user's bid is not sufficiently higher than the other bid, the tx will revert. They also both gracefully handle the case where a collector wants to initialize a new auction with a bid at e.g. 3x initial bid value, and someone else front-runs an initial bid equal to initial bid value. In that case, both implementations place a new bid at 3x the initial bid value, and return funds to the front-runner.

The benefits of this PR are entirely about simplifying the interface to be "settle", "bid", and "settle&bid", removing the concept of "initialize" (since initializing is performed by bidding anyway).

jakerockland: That makes sense to me ser – appreciate the additional clarification here re: rationale.

jakerockland (approved):

PR #520, opened 2023-03-09
Adds getTokenHtml function to dependency registry contract

yoshiwarab:
## Description of the change
Taking advantage of scripty.sol, this PR adds a new version of the DependencyRegistry contract that enables a form of fully on chain generator via an added getTokenHtml function. Currently this only works for V3 and above core contracts which have a dependency that has been fully uploaded to the dependency registry.

To see this in action go to https://goerli.etherscan.io/address/0x47AF6d74AB159A7E95523f14424338980f434a0A#readProxyContract#F18 and enter `0x5702797ff45fcb0a70eb6ae1e4563299dcfa9dd6` in the `_contractAddress` field and `14000000` in the `_tokenId` field. Copy the response into your browser address bar to see a fully on-chain Ringers iteration. 

Diff betweend v1 and v0:
https://www.diffchecker.com/TdANsM3w/


contracts/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1074 @@
...
+        uint256 projectId = _tokenId / ONE_MILLION;
+        bytes32 dependencyType = stringToBytes32(
+            this.getDependencyTypeForProject(_contractAddress, projectId)
+        );
+        bytes32 tokenHash = IDependencyRegistryCompatibleV0(_contractAddress)
+            .tokenIdToHash(_tokenId);
+
+        WrappedScriptRequest[] memory requests = new WrappedScriptRequest[](5);
+
+        requests[0]
```
yoshiwarab, line 906: Note: Scripty.sol uses a single hardcoded html template so we use an extra script to setup our default template style. Alternatively we could fork the scripty contract and update the template there to include the style tag.

contracts/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1074 @@
...
+        requests[1].scriptContent = abi.encodePacked(
+            'let tokenData = {"tokenId":"',
+            Strings.toString(_tokenId),
+            '"',
+            ',"hash":"',
+            bytesToHexString(abi.encodePacked(tokenHash)),
+            '"}'
+        );
+
+        requests[2].name = "dependency";
```
yoshiwarab, line 918: Note: I initially uploaded just the minified library but ran into gas limit issues so I had to upload a gzipped version instead.

contracts/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1074 @@
...
+        requests[2].contractData = abi.encode(dependencyType);
+        requests[2].contractAddress = address(this);
+        requests[2].wrapType = 2; // <script type="text/javascript+gzip" src="data:text/javascript;base64,[script]"></script>
+
+        requests[3].name = "gunzipScripts-0.0.1.js";
```
yoshiwarab, line 923: This is required to unzip the gzipped lib

scripts/util/upload-dependency-script-to-dependency-registry.ts
```diff
@@ -0,0 +1,133 @@
...
+program.parse(process.argv);
+
+export const maxContentSize = 24000;
+
+function chunk(s, maxBytes) {
+  let buf = Buffer.from(s);
+  const result = [];
+
+  // Use Graphemer to split on grapheme boundaries
+  const splitter = new Graphemer();
```
yoshiwarab, line 36: Note: I wanted this to be set up to handle chunking of any sort of large file so this function is a bit more complex than it needs to be if we expect to only upload base64 encoded, gzipped files which should only really have single byte characters.

jakerockland:
> Currently this only works for V3 and above core contracts which have a dependency that has been fully uploaded to the dependency registry.

IIUC, won't < V3 also be supported if an override is provided?

contracts/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1074 @@
...
+    EnumerableSet.Bytes32Set private _dependencyTypes;
+    mapping(bytes32 => Dependency) dependencyDetails;
+
+    EnumerableSet.AddressSet private _supportedCoreContracts;
+    mapping(address => mapping(uint256 => bytes32)) projectDependencyTypeOverrides;
+
+    IScriptyBuilder public scriptyBuilder;
+    address public ethFsAddress;
+
+    modifier onlyNonZeroAddress(address _address) {
```
ryley-o, line 78: nit - do you mind switching these to internal functions? recently switched over all active contracts to not use modifiers as a best practice, described here: https://github.com/ArtBlocks/artblocks-contracts/blob/main/GUIDELINES.md

contracts/DependencyRegistryV1.sol
ryley-o: I don't have super strong feelings here, but if we consider forking, perhaps @thedude would be willing to help figure out a more globally optimal solution here 👥

contracts/DependencyRegistryV1.sol
ryley-o: curious - do we support uploading in multiple chunks? feels like we should be able to handle arbitrarily large dependencies, even if it means uploading across multiple txs/blocks?

contracts/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1074 @@
...
+import "@openzeppelin-4.5/contracts/utils/math/SafeCast.sol";
+
+import {IContractScript} from "scripty.sol/contracts/scripty/IContractScript.sol";
+import {AddressChunks} from "./AddressChunks.sol";
+import {IScriptyBuilder, WrappedScriptRequest} from "scripty.sol/contracts/scripty/IScriptyBuilder.sol";
+
+import "./libs/0.8.x/BytecodeStorage.sol";
+import "./libs/0.8.x/Bytes32Strings.sol";
+
+import "hardhat/console.sol";
```
ryley-o, line 23: interesting I've never used this 👀  are you using this functionality at all?

ryley-o (commented):
this is really really really really really cool!

A few thoughts based on current draft:
- Do you mind diffing DependencyRegistry V0 vs. V1, pulling in some of the changes from V0 to V1 (I think this is mostly limited to modifiers being switched to internal functions per a recent style guide update)
- Would there be any appetite at all to return html text instead of base64? might be a slightly cleaner UX for someone non-technical?
- What does the getScript function do?

contracts/DependencyRegistryV1.sol
yoshiwarab: Oops, didn't mean to leave this in, was just using for debugging

yoshiwarab:
> Do you mind diffing DependencyRegistry V0 vs. V1, pulling in some of the changes from V0 to V1 (I think this is mostly limited to modifiers being switched to internal functions per a recent style guide update)

Yes, will do!

> Would there be any appetite at all to return html text instead of base64? might be a slightly cleaner UX for someone non-technical?

Yeah definitely! I chose the output method here with the idea that we could potentially insert the output of the function directly into the animation_url of the token api response but could for sure see it being nice to have decoded html. Would probably be fairly easy to break out the shared part of the logic for that.

> What does the getScript function do?

`getScript` is the function that ScriptyBuilder expects all on-chain storage sources to implement (see https://github.com/intartnft/scripty.sol/blob/81e8f48663dbd0e5b94cc3ecade1ad1db1e29f83/contracts/scripty/IContractScript.sol). Here it combines all project/contract scripts together, so they can be fetched all at once by scripty builder.

contracts/DependencyRegistryV1.sol
yoshiwarab: To clarify, we do upload dependencies in chunks but the getTokenHtml call was failing to return the full expected output because of provider enforced gas limit (e.g. https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_call)

contracts/DependencyRegistryV1.sol
0xdude: I was actually planning to add another request type for the headers. Something like this:

HeaderTagRequest
- tagPrefix
- tagSuffix
- content

This way you can add custom tags in `<header></header >` tags. In your case you can define your css like this:
```
HeaderTagRequest[] memory headerTagRequests = new HeaderTagRequest[](1);
headerTagRequests[0].tagPrefix = "<style>";
headerTagRequests[0].tagPrefix = "</style>";
headerTagRequests[0].content = "html{height:100%}body{min-height:100%;margin:0;padding:0}canvas{padding:0;margin:auto;display:block;position:absolute;top:0;bottom:0;left:0;right:0}";
```

Maybe this way you can even add `<metadata>` to set description author title etc?
```
scriptyBuilder.getEncodedHTMLWrapped(
    headerTagRequests, 
    scriptRequests,
    bufferSize
);
```

scripty.sol is initially designed to focus on scripts but I think having this feature is valuable and makes scripty even more flexible. We wanted scripty.sol to be immutable so it has no owner and it's not upgradable but since it doesn't store anything and it has no dependency, deploying scriptyv2.sol shouldn't be a problem.

What do you think?

yoshiwarab:
> IIUC, won't < V3 also be supported if an override is provided?
The way this works now currently leans on AddressChunks.sol to concatenate data from BytecodeStorage contracts into a single script but I don't think it would be too hard to extend our logic to concatenate strings from older contracts as well. Not totally sure what the effect on gas efficiency here would be though

jakerockland:
> > IIUC, won't < V3 also be supported if an override is provided?
> 
> The way this works now currently leans on AddressChunks.sol to concatenate data from BytecodeStorage contracts into a single script but I don't think it would be too hard to extend our logic to concatenate strings from older contracts as well. Not totally sure what the effect on gas efficiency here would be though

Ahhhh–yes, I see I see. 

I do think its worth extending for that purpose – despite the bit of rigamarole required 😅 😄

contracts/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1074 @@
...
+        requests[3].contractAddress = ethFsAddress;
+
+        requests[4].name = "project";
+        requests[4].contractData = abi.encode(_contractAddress, projectId);
+        requests[4].contractAddress = address(this);
+        requests[4].wrapType = 0; // <script>[script]</script>
+
+        uint256 bufferSize = scriptyBuilder.getBufferSizeForEncodedHTMLWrapped(
+            requests
+        );
```
0xdude, line 934: This here is a bit problematic. 

Just to give a little bit context.. scripty.sol gets bufferSize to allocate a big amount of container memory
which then is subsequently filled without needing to reallocate. This way scripty can stitch scripts together in a gas efficient way.

`scriptyBuilder.getBufferSizeForEncodedHTMLWrapped()` -> This method actually builds the HTML and then returns the buffer size of final HTML. So calling this method on-chain basically doubles the HTML building operation which might explain why you had gas issue initially. It's intended to be used mostly off-chain.

So I wondering if this approach would be better for AB:
- getDependencyScript(params..) -> String  - Returns full dependency script text. (Similar to getScript method but tuned for dependency)
- getArtScript(params..) -> String - Returns full art script. (Similar to getScript but tuned for art scripts)
- getArtControllerScript(params...) -> String returns the let tokenData = {"tokenData script.

Then you can get the final buffer size:
- dependencyScript.length + artScript.length + artControllerScript.length

and pass it to scripty.

These are few benefits I see:
- All components needed to build final HTML would be separated and can be called publicly in a clean way. 
- No extra contract call gas cost from scripty. (Right now registry calls scripty and then scripty calls registry back to reach the scripts.)

What do you think?

0xdude:
Hello, this looks amazingly cool! Thanks for digging into scripty :)

I just left few comments for registry contract implementation. I would love to contribute and collaborate so I'll be following this implementation and if you need anything else you can always ping me.

contracts/DependencyRegistryV1.sol
ryley-o: got it, makes sense and I think natural next question would be if we can make it more efficient. Sounds like some recommendations fom @0xdude below will help with that. We could probably use some assembly to convert bytes to string as well, if it enables us to avoid some of these provider-imposed read limits!

ryley-o: I'll resolve this thread and address gas efficiency details as separate comment threads!

contracts/DependencyRegistryV1.sol
yoshiwarab: Ah good call, that makes a lot of sense!

contracts/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1074 @@
...
+
+        requests[0]
+            .scriptContent = 'let css="html{height:100%}body{min-height:100%;margin:0;padding:0}canvas{padding:0;margin:auto;display:block;position:absolute;top:0;bottom:0;left:0;right:0}",head=document.head,style=document.createElement("style");head.appendChild(style),style.type="text/css",style.appendChild(document.createTextNode(css));';
+
+        requests[1].scriptContent = abi.encodePacked(
+            'let tokenData = {"tokenId":"',
+            Strings.toString(_tokenId),
+            '"',
+            ',"hash":"',
+            bytesToHexString(abi.encodePacked(tokenHash)),
```
ryley-o, line 914: nit: probably easier to cast `tokenHash` as a `uint256` here, then use OpenZeppelin Strings.toHexString function?

contracts/DependencyRegistryV1.sol
yoshiwarab: Yes, I think this would be super useful and much cleaner! Just to be clear, you mean `<head></head>` tags and not header tags right?

lyaunzbe: ++ this is awesome and a nice way to handle the potential of supporting multiple templates (in the case we don't end up consolidating into a singular one)

contracts/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1074 @@
...
+        requests[4].wrapType = 0; // <script>[script]</script>
+
+        uint256 bufferSize = scriptyBuilder.getBufferSizeForEncodedHTMLWrapped(
+            requests
+        );
+
+        bytes memory base64EncodedHTMLDataURI = scriptyBuilder
+            .getEncodedHTMLWrapped(requests, bufferSize);
+
+        return string(abi.encodePacked(base64EncodedHTMLDataURI));
```
ryley-o, line 939: for more gas efficiency, I think the `abi.encodePacked` here is unnecessary.
Additionally, since bytes and strings are the same thing, you should be able to optimally use:
```
assembly {
    output_ := base64EncodedHTMLDataURI
}
```
where `output_` can be changed to be the string that is returned.

contracts/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1074 @@
...
+        requests[4].contractAddress = address(this);
+        requests[4].wrapType = 0; // <script>[script]</script>
+
+        uint256 bufferSize = scriptyBuilder.getBufferSizeForEncodedHTMLWrapped(
+            requests
+        );
+
+        bytes memory base64EncodedHTMLDataURI = scriptyBuilder
+            .getEncodedHTMLWrapped(requests, bufferSize);
+
```
0xdude, line 938: Regarding having base64/raw HTML returning methods. 

If you want to support both versions you can use scripty's these methods together:
- getEncodedHTMLWrapped() -> bytes | base64 data uri 
- getHTMLWrapped() -> bytes | raw html

Maybe an internal method can return the requests + bufferSize and then you can have 2 separate methods that can return raw HTML and base64 encoded one. scripty internally calculates the buffer size offset for base64 conversion so you can use same requests and bufferSize for both methods.

contracts/DependencyRegistryV1.sol
0xdude: Yup I meant `<head></head>` tags :) alright I'll pick this up and let you know when I draft it.

contracts/dependency-registry/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1092 @@
...
+        requests[4].wrapType = 0; // <script>[script]</script>
+
+        IScriptyBuilder scriptyBuilder = IScriptyBuilder(scriptyBuilderAddress);
+
+        uint256 bufferSize = requests[0].scriptContent.length +
+            requests[1].scriptContent.length +
+            requests[2].scriptContent.length +
+            requests[3].scriptContent.length +
+            requests[4].scriptContent.length +
+            500;
```
yoshiwarab, line 1023: @0xdude I was getting `DynamicBuffer: Appending out of bounds.` reverts without adding this arbitrary 500 bytes to the buffer, anything I'm doing obviously wrong here?

contracts/dependency-registry/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1123 @@
...
+            '"}'
+        );
+
+        bytes memory dependencyScript = this.getDependencyScript(
+            dependencyType
+        );
+        requests[2].scriptContent = dependencyScript;
+        requests[2].wrapType = 2; // <script type="text/javascript+gzip" src="data:text/javascript;base64,[script]"></script>
+
+        bytes memory gunzipScript = IContractScript(ethFsAddress).getScript(
```
yoshiwarab, line 1003: @jakerockland @ryley-o Are we good with gzipping our dependency scripts? This should allow the project scripts to be longer before hitting any provider enforced gas limit so seems like a good idea to me. Also, how do we feel about depending on ethfs for pulling this script?

contracts/dependency-registry/Scripty.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: Unlicense
```
yoshiwarab, line 1: Added this file so that hardhat would compile ScriptyBuilder for testing. @jakerockland @ryley-o let me know if there's a better way to do this

contracts/dependency-registry/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1127 @@
...
+            '"',
+            ',"hash":"',
+            Strings.toHexString(uint256(tokenHash)),
+            '"}'
+        );
+
+        Dependency storage dependency = dependencyDetails[dependencyType];
+        if (dependency.scriptCount == 0) {
+            requests[2].wrapPrefix = abi.encodePacked(
+                '<script type="text/javascript" src="',
```
yoshiwarab, line 1000: @0xdude this definitely goes against the fully on-chain intention of scripty.sol, but what do you think about adding a wrapType for loading off-chain libraries? Definitely convenient for us while we're in the process of storing our dependencies on-chain.

yoshiwarab:
@jakerockland @ryley-o still need to add more comments + tests but I think this is ready for another round of review. I've added the ability to load pre-v3 scripts + use the preferred cdn if a dependency script isn't available on-chain

contracts/dependency-registry/DependencyRegistryV1.sol
0xdude: Sorry for misleading previous comment. I believe, since all requests use `scriptContent` you can use this now:
`getBufferSizeForHTMLWrapped(requests)`

Scripty won't do contract call and just calculate all the script length and tag sizes. It should match but let me know if you have any problem.

contracts/dependency-registry/DependencyRegistryV1.sol
0xdude: From scripty perspective this is fine. Those wrapPrefix and wrapSuffix are there to have custom script tags so it's fine to me 👍

0xdude:
Maybe I'm overthinking but, considering the size of registry contract and its responsibilities, maybe separating dependency script management from HTML rendering would be better. 

In my opinion, dependency script management has a lower chance of needing an upgrade in the future. But HTML rendering may need to be updated in the future to support browser changes, new templates, and new HTML properties.

Would this be possible?
ArtBlockHTMLRenderer - depends on:
- GenArtCore contracts
- DependencyRegistry 

What do you think?

contracts/dependency-registry/Scripty.sol
ryley-o: we sometimes put those files in `/contracts/libs/integration-refs/<reference-name>/`, just to clearly designate them as references that we didn't design

ryley-o: probably worth adding a dev comment to the file indicating what you did in this comment?

contracts/dependency-registry/DependencyRegistryV1.sol
```diff
@@ -0,0 +1,1127 @@
...
+     * @return result bytes32 representation of string.
+     */
+    function stringToBytes32(
+        string memory source
+    ) public pure returns (bytes32 result) {
+        bytes memory tempEmptyStringTest = bytes(source);
+        if (tempEmptyStringTest.length == 0) {
+            return 0x0;
+        }
+
```
ryley-o, line 853: This should perform the following check to ensure the string is not too long to fit into a bytes32:
```
require(bytes(source).length > 31, "String too long");
```

ryley-o (commented):
Looks good, mostly minor nits from me!

In addition to comments, please resolve the compiler warnings (SPDX-License-Identifier warnings, unused variables, shadow declarations, and pure function suggestion).

.openzeppelin/goerli.json
```diff
@@ -9,6 +9,16 @@
       "address": "0x32CAA016dd6e8d5750517D5777e2EB70287B023e",
       "txHash": "0xc61bd5a1a84231c98a4cd206d7f8cceda4058f6cad673d389afee1d38d7e7880",
       "kind": "transparent"
+    },
```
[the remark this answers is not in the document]
yoshiwarab: This gets auto-generated when deploying an upgradeable contract. Was considering doing a new deploy entirely and wiping this once this PR is in a finalized state.

contracts/DependencyRegistryV1.sol
xtremetom: @0xdude if you haven't started on this yet, I'll get cracking on it tomo

0xdude:
Just let you know, @xtremetom and I are currently working on ScriptyV2. He already completed most of the heavy work. 

We are aiming to:
- Support <head> tags. (done)
- Eliminate need of calculating buffer size. (almost done)
- Separating builder contract into different sub components. (almost done)

Once it's mature enough we will share it here.

0xdude:
@yoshiwarab we completed the V2 and we think it's almost ready to be released. There are small syntax changes but general idea is the same. Major changes:
- No buffer size calculation needed before requesting HTML.
- Supports adding any tags under <head> tag. Supports storage as well. For testing, I stored your full screen canvas CSS on EthFS. You can directly read it from there. But a huge win as CSS is pretty small but it's something to consider I guess.

You can check the v2 repo here: 
https://github.com/intartnft/scripty.sol/tree/scripty-v2/contracts/scripty

V2 docs are here (Please note that it's not done yet. I just added couple of examples to show new APIs)
https://int-art.gitbook.io/scripty.sol-v2/building-html/getting-html-on-chain

Any feedback is welcomed!

yoshiwarab:
@0xdude, closing this in favor of https://github.com/ArtBlocks/artblocks-contracts/pull/1320 if you'd like to take a look!

PR #532, opened 2023-03-13
Fix mismatched selectors on Flex adminACL check

ryley-o:
## Description of the change

Fix mismatched selectors on Flex adminACL check.

This issue was caught prior to any meaningful deployments of the flex contract. However, interested in discussion on if we should add tests to catch bugs of this class for all adminACL-gated functions.


lyaunzbe (approved):
😅 thank you for catching this (probably copy-paste related) bug 🙏 🙏 

IMO, if we can think of some sort of reusable (doesn't duplicate a lot of code) way to test these gates it would be worth it to avoid these kinds of bugs for future

jakerockland (approved):
Great catch.

> However, interested in discussion on if we should add tests to catch bugs of this class for all adminACL-gated functions.

Is there a good way to do this in a generalized/automated way? 🤔

PR #546, opened 2023-03-16
SEA minter - hold n+1 token to improve UX

ryley-o:
## Description of the change

Update SEA minter smart contracts to hold "next token" to improve UX when shifting from one auction to the next.

**Proposed to be folded into #504**

This iteration adds smart contract complexity and minter-state complexity relative to #504, but results in an improved UX. The main benefit is that when transitioning from one auction to the next, the next token to be auctioned is known before the initial bid. Additionally, since the auction holds the next token to be minted, it is possible for a single transaction to settle the previous auction, initialize the auction for the known, next token, and place a bid on said token.

## Tests
Tests are updated for these changes, and full coverage is reached for the portions of code that are considered reachable.
 
_a couple lines are "unreachable" and are not covered during testing. These "unreachable" lines are included in the minter to provide redundant protection against a state that is thought to be impossible, but would be dangerous if possible_

### DIFF
--- a/contracts/minter-suite/Minters/MinterSEAV0.sol
+++ b/contracts/minter-suite/Minters/MinterSEAV0.sol
@@ -13,6 +13,7 @@ import "./MinterBase_v0_1_1.sol";
 import "@openzeppelin-4.7/contracts/token/ERC721/IERC721.sol";
 import "@openzeppelin-4.7/contracts/security/ReentrancyGuard.sol";
 import "@openzeppelin-4.7/contracts/utils/math/SafeCast.sol";
+import "@openzeppelin-4.7/contracts/utils/math/Math.sol";
 
 /**
  * @title Filtered Minter contract that allows tokens to be minted with ETH.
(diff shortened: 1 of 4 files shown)


contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -457,59 +519,59 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
...
+            (_auction.tokenId != _tokenId)
+        ) {
+            // auction not initialized, already settled, or is for a different
+            // token ID, so return early
             return;
         }
         require(block.timestamp > _auction.endTime, "Auction not yet ended");
-        // @dev this check is not strictly necessary, but is included for
-        // clear error messaging
-        require(_auction.initialized, "Auction not initialized");
```
jakerockland, line 512: What's the rationale for pulling this out of the require (and up above instead)?

jakerockland (approved):
I won't have a chance to get more into the details of #504 until next week unfortunately, but this delta against that PR LGTM.

One minor ask – I do think we should follow up on adding the "eject mint" type functionality as an artist (or perhaps admin) only functionality that can only be used when auction process has already been stopped. 

Not sure if you want to add that in this PR or in an alternative follow up (or in #504 directly perhaps already and I've just not seen it yet).

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: good callout - seems clearer to revert with this error message. Updated in fc3b08d 🙏

ryley-o:
>One minor ask – I do think we should follow up on adding the "eject mint" type functionality as an artist (or perhaps admin) only functionality that can only be used when auction process has already been stopped.

>Not sure if you want to add that in this PR or in an alternative follow up (or in https://github.com/ArtBlocks/artblocks-contracts/pull/504 directly perhaps already and I've just not seen it yet).

Good call - I think best to add it in a follow-up PR and fold it into #504, just to keep discussion on that functionality focused (since there are a few ways we could design it). I'll toss a follow-on PR sometime today 🙏

PR #549, opened 2023-03-17
add ejectNextTokenTo admin-only function

ryley-o:
## Description of the change

\<add description here\>


jakerockland (approved):
This LGTM – would encourage expanding on the inline documentation here to clarify why only one token will ever be in a "needs ejecting" state for a given project.

ryley-o:
updated some comments and folding into #504 🙏

PR #571, opened 2023-03-23
Update royalties docs, add diagram

ryley-o:
## Description of the change

Update royalties documentation and add royalty registry override diagram.

This somewhat simplifies the information contained in this repository, and also points to our Art Blocks Docs site to point to a single source of truth for instructions on how to configure royalty registry override contracts.


jakerockland (approved):
This is fantastic – thank you @ryley-o 

Worth also getting @ABDruid's eyes on this and iterating as needed, though I don't think it needs to block this PR.

🔥 🙌

PR #581, opened 2023-03-27
Engine Flex contract mod

cxkoda:
## Description of the change

Adding a modified version of `GenArt721CoreV3_Engine_Flex` that replaces all `onlyArtist` with `onlyArtistAndAdminACL`.

### DIFF
--- a/contracts/GenArt721CoreV3_Engine_Flex_PROOF.sol
+++ b/contracts/GenArt721CoreV3_Engine_Flex_PROOF.sol
@@ -1087,6 +1087,7 @@ contract GenArt721CoreV3_Engine_Flex_PROOF is
      * @param _projectId Project ID to be toggled.
      */
     function toggleProjectIsPaused(uint256 _projectId) external {
+        _onlyValidProjectId(_projectId);
         _onlyArtistOrAdminACL(_projectId, this.toggleProjectIsPaused.selector);
         projects[_projectId].paused = !projects[_projectId].paused;
         emit ProjectUpdated(_projectId, FIELD_PROJECT_PAUSED);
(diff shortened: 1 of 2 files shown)


contracts/GenArt721CoreV3_Engine_Flex_PROOF.sol
```diff
@@ -0,0 +1,2278 @@
...
+ * the Admin ACL contract, only when the project is not locked:
+ * - updateProjectName
+ * - updateProjectArtistName
+ * - updateProjectLicense
+ * - Change project script via addProjectScript, updateProjectScript,
+ *   and removeProjectLastScript
+ * - updateProjectScriptType
+ * - updateProjectAspectRatio
+ * ----------------------------------------------------------------------------
+ * The following functions are restricted to only the Artist address:
```
ryley-o, line 57: Let's update this section's description to be something like `The following functions are restricted to only the Artist or Admin ACL contract of a valid project ID:`

Additionally, since the `onlyArtist` implicitly checked for a valid project ID, please add `_onlyValidProjectId(_projectId)` to each of the adjusted functions in this section 🙏

ryley-o (changes requested):
A couple of minor nits in the comment below, and then this can be approved 👍 thank you!

contracts/GenArt721CoreV3_Engine_Flex_PROOF.sol
cxkoda: Added the former. Also happy to add the latter, however `onlyArtist` does not imply `_onlyValidProjectId` as far as I can tell from https://github.com/ArtBlocks/artblocks-contracts/blob/b2c4e24848fee440226bc56302edc1c72c2c5b3f/contracts/GenArt721CoreV3_Engine_Flex.sol#L329-L345

ryley-o: resolved in 4b7f5bb & 48be657 🎉

jakerockland (approved):
LGTM 🙏

PR #587, opened 2023-03-28
bug fix - SEA minter - mint next token with auth of msg.sender

ryley-o:
## Description of the change

Update SEA minter to mint next token with auth of `msg.sender` to enable artists to configure projects while paused.

This correctly handles the scenario where an artist is configuring their auction or minter max invocations while their project is still paused.

>note: this bug fix does not patch anything that would result in loss of funds. additionally, the SEA minter has only been deployed to the dev environment, and the deployed minter will be removed from the minter filter to prevent non-fixed versions of this minter being used (even on dev).


jakerockland (approved):
🙏 🙏

PR #617, opened 2023-04-12
Minter Filter V2

ryley-o:
## Description of the change

Add new minter filter contract that is compatible with multiple core contracts.

Reviewers should refer to the internal design document that describes design choices and implementation details in more detail: https://docs.google.com/document/d/1Ix0LhPHqeDgCouKPSYZufQY9zQ5cNqf8yXHbus8y3w8/edit?usp=sharing

## Changed Items:
- Minter Filter
- Engine Registry (require ability to query the registry to see if contract is approved in registry)

## Items that will require changes, but are not included in this PR:
- All Minter contracts

## Tests

- Full test coverage of new contracts is achieved
- Additional tests that ensure events are emitted, etc. are also added

---
  - https://app.asana.com/0/0/1204365671969089


ryley-o:
This is ready for an initial review, with tests to follow after any iterations 🙏

contracts/engine-registry/future/EngineRegistryV1.sol
jakerockland, line 1: Should we be deleting this vs. moving it to a deprecated folder? Seems wrong to delete IIUC.

contracts/engine-registry/CoreRegistryV1.sol
```diff
@@ -0,0 +1,240 @@
...
+ * - Collaboration contracts
+ * - Engine contracts
+ * - Engine Flex contracts
+ *
+ * Note that not all contracts will be registered in this registry, as some
+ * contracts may not need to interact with a MinterFilterV2 contract. For
+ * example, the original Art Blocks V0 contract does not need to interact with
+ * a MinterFilterV2 contract, as it uses a different minting mechanism.
+ *
+ * A view funciton is provided to determine if a contract is registered.
```
jakerockland, line 36:
```suggestion
 * A view function is provided to determine if a contract is registered.
```

contracts/engine-registry/CoreRegistryV1.sol
```diff
@@ -0,0 +1,240 @@
...
+    function unregisterContract(address _contractAddress) external {
+        // CHECKS
+        // Prevent unregistration of a contract by an unrelated contract
+        require(
+            msg.sender == _contractAddress || msg.sender == owner(),
+            "Only calls by owner or contract"
+        );
+        // EFFECTS
+        _unregisterContract(_contractAddress);
+    }
```
jakerockland, line 112: @ryley-o I don't think we want to allow contracts to register/unregister _themselves_ 🤔

jakerockland (changes requested):
Some areas for a bit more discussion / back-and-forth before I think worth adding full test coverage here ser (comments added where I think relevant for further double-clicking) – but general approach LGTM. 🙏 💜 🔥 

Diff that I used during my review: https://www.diffchecker.com/Q8oGgh7R

contracts/engine-registry/CoreRegistryV1.sol
ryley-o: hmm...

For registering, I think we want to allow a contract to register itself if `tx.origin == owner`, because that enables the deployment process where the contract register's itself via a call from the contract's constructor. `EngineRegistryV0` only checks that `tx.origin` is the deployer address (aka `owner()` in this update), which we still enforce. Adding an additional constraint on `msg.sender` adds a small amount of security benefit (albeit minor).

For unregistering, `EngineRegistryV0` only checks that `tx.origin` is the deployer address (aka `owner()`), which is not ideal from a security perspective. I think this version should definitely change that check to be `msg.sender == owner()`. I would be okay with not allowing contracts to unregister themselves (since there isn't really a reason I can think of why someone would fork the V3 Engine contract to do that).

contracts/engine-registry/future/EngineRegistryV1.sol
ryley-o: we ended up never implementing or deploying this Engine Registry V1 contract, and this new shared Core Registry is substantially different.
Since it was never used, I'm totally fine scrapping it entirely if you are okay with that.

contracts/engine-registry/CoreRegistryV1.sol
jakerockland: Hrm, I have some thoughts on why this may create somewhat of a problematic pattern – let's discuss more on Monday! 🙏

contracts/engine-registry/future/EngineRegistryV1.sol
jakerockland: AH – yeah that makes sense, SGTM ser.

contracts/engine-registry/CoreRegistryV1.sol
ryley-o: TODO: I'll remove the ability for a contract to eject itself 🙏

contracts/engine-registry/CoreRegistryV1.sol
```diff
@@ -0,0 +1,240 @@
...
+    ) external {
+        // CHECKS
+        // Validate against `tx.origin` rather than `msg.sender` as it is intended that this registration be
+        // performed in an automated fashion *at the time* of contract deployment for the `_contractAddress`.
+        require(tx.origin == owner(), "Only tx origin of owner");
+        // Prevent registration of a contract by an unrelated contract
+        require(
+            msg.sender == _contractAddress || msg.sender == owner(),
+            "Only calls by owner or contract"
+        );
```
ryley-o, line 91: improve readability here, improve documentation, maybe its own modifier-like gating function

ryley-o: updated

contracts/engine-registry/CoreRegistryV1.sol
ryley-o: updated and resolving, based on today's conversation with @jakerockland 🙏

jakerockland (approved):
TYVM for the updates here – LGTM, pending tests.

lyaunzbe (approved):

PR #629, opened 2023-04-13
fix CI issue

ryley-o:
## Description of the change

\<add description here\>


ryley-o:
locking to node version in CI fixes the issue.closing this in favor of a simpler commit history

PR #631, opened 2023-04-13
Engine deploys (2023-04-13)

jakerockland:
*Note*: I am waiting to run deploy scripts on mainnet until _after_ this PR is reviewed, to give chance for second set of eyes on this PR and config files added before doing production mainnet deploy. 

## Description of the change

_Engine deploys for the following tickets:_

(Goerli/Staging):
- https://github.com/ArtBlocks/artblocks/issues/551
- https://github.com/ArtBlocks/artblocks/issues/557

(Mainnet):
- https://github.com/ArtBlocks/artblocks/issues/547
- https://github.com/ArtBlocks/artblocks/issues/550
- https://github.com/ArtBlocks/artblocks/issues/552
- https://github.com/ArtBlocks/artblocks/issues/553

PR #634, opened 2023-04-14
try/catch generic minter deployment minter filter allowlisting

ryley-o:
## Description of the change

Add try/catch on generic minter deployment minter filter allowlisting.

This also reports+logs any necessary follow-on actions to the user running the script.

Tested on goerli to ensure WAI

follow on from #631, so it also fixes the bugged DEPLOYMENTS.md from the deployments in that PR.


jakerockland (approved):
🐐

PR #637, opened 2023-04-17
Initial implementation of new minter architecture with libs (SetPriceMinterV5, SetPriceMinterV5Merkle, SetPriceMinterV5Holder)

lyaunzbe:
## Description of the change

This is the first implementation of our new minter-suite architecture re-design, specifically in the context of creating a new set price minter + gated variations of this minter + associated internal libraries that contain shared functionality + any associated shared interfaces. 

New minters include:
- MinterSetPriceV5
- MinterSetPriceV5Holder
- MinterSetPriceV5Merkle

Minter tests generally have the following structure:
- configure.tests.ts
- events.tests.ts
- integration.test.ts
- views.tests.ts

Shared common test files are used where it makes sense

---
  - https://app.asana.com/0/0/1204892884069010


jakerockland:
Should this be rebased off of `[minter-filter-v2](https://github.com/ArtBlocks/artblocks-contracts/tree/minter-filter-v2)` instead of `main` @lyaunzbe ?

contracts/interfaces/0.8.x/IFilteredSharedMerkle.sol
```diff
@@ -0,0 +1,44 @@
...
+ * @title This interface extends the IFilteredMinterV1 interface in order to
+ * add support for including Merkle proofs when purchasing.
+ * @author Art Blocks Inc.
+ */
+interface IFilteredSharedMerkle {
+    // /**
+    //  * @notice Notifies of the contract's default maximum mints allowed per
+    //  * user for a given project, on this minter. This value can be overridden
+    //  * by the artist of any project at any time.
+    //  */
```
jakerockland, line 16: Nit: don't need the `//`s with the `/**`s
```suggestion
    /**
     * @notice Notifies of the contract's default maximum mints allowed per
     * user for a given project, on this minter. This value can be overridden
     * by the artist of any project at any time.
     */
```

contracts/interfaces/0.8.x/IFilteredSharedMerkle.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title This interface extends the IFilteredMinterV1 interface in order to
```
jakerockland, line 7: IIUC this is a separate interface, no? Rather than actually needing to extend?

contracts/libs/0.8.x/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    function updateMerkleRoot(
+        mapping(address => mapping(uint256 => bytes32))
+            storage projectMerkleRoot,
+        uint256 _projectId,
+        address _coreContract,
+        bytes32 _root
+    ) internal {
+        require(_root != bytes32(0), "Root must be provided");
+        projectMerkleRoot[_coreContract][_projectId] = _root;
```
jakerockland, line 44: I think we may want to start exposing an event here, and include it in our interface?

contracts/interfaces/0.8.x/IFilteredMinterV3.sol
```diff
@@ -0,0 +1,223 @@
...
+        uint256 _projectId,
+        address _coreContract
+    ) external payable returns (uint256 tokenId);
+
+    // Toggles the ability for `purchaseTo` to be called directly with a
+    // specified receiving address that differs from the TX-sending address.
+    function togglePurchaseToDisabled(
+        uint256 _projectId,
+        address _coreContract
+    ) external;
```
jakerockland, line 199: I wonder if we want to continue to expose this in this interface here, vs. moving it to the `SharedMerkle` and related token-gated interfaces?

contracts/interfaces/0.8.x/IFilteredSharedProjectMaxInvocationsLimit.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title This interface extends the IFilteredMinterV1 interface in order to
+ * add support for manually setting project max invocations.
+ * @author Art Blocks Inc.
+ */
```
jakerockland, line 10: IIUC this is a separate interface, no? Rather than actually needing to extend?

contracts/interfaces/0.8.x/IFilteredSharedProjectMaxInvocationsLimit.sol
```diff
@@ -0,0 +1,30 @@
...
+interface IFilteredSharedProjectMaxInvocationsLimit {
```
jakerockland, line 11: Curious the rationale for having this be its own interface rather than just bundling into `IFilteredMinterV3`?

contracts/libs/0.8.x/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+import "@openzeppelin-4.7/contracts/utils/cryptography/MerkleProof.sol";
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title Art Blocks Merkle Library
+ * @notice [TO FILL OUT]
```
jakerockland, line 10: When we fill this out, lets make sure we also have natspec comments for the `ProjectConfig` struct and all our methods here as well (ChatGPT is great at writing initial comments btw hah 😄).

contracts/libs/0.8.x/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+    ) internal pure returns (bool) {
+        return _proof.verifyCalldata(_proofRoot, hashAddress(_address));
+    }
+
+    function projectMaxInvocationsPerAddress(
+        uint256 _projectId,
+        address _coreContract,
+        mapping(address => mapping(uint256 => ProjectConfig))
+            storage projectConfig
+    ) public view returns (uint256) {
```
jakerockland, line 71: I think we probably want this to be internal, given that we intend to use this as a bundled internal library ya?

contracts/libs/0.8.x/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+library MerkleLib {
+    using MerkleProof for bytes32[];
+    uint256 public constant DEFAULT_MAX_INVOCATIONS_PER_ADDRESS = 1;
+    bytes32 public constant CONFIG_MERKLE_ROOT = "merkleRoot";
+    bytes32 public constant CONFIG_USE_MAX_INVOCATIONS_PER_ADDRESS_OVERRIDE =
+        "useMaxMintsPerAddrOverride"; // shortened to fit in 32 bytes
+    bytes32 public constant CONFIG_MAX_INVOCATIONS_OVERRIDE =
+        "maxMintsPerAddrOverride"; // shortened to match format of previous key
```
jakerockland, line 21: I think we probably want these to be internal, given that we intend to use this as a bundled internal library ya?

contracts/interfaces/0.8.x/IFilteredMinterV3.sol
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     * @notice Local max invocations for project `_projectId`, tied to core contract `_coreContractAddress`,
+     * updated to `_maxInvocations`.
+     */
+    event ProjectMaxInvocationsLimitUpdated(
+        uint256 indexed _projectId,
+        address indexed _coreContract,
+        uint256 _maxInvocations
+    );
```
jakerockland, line 172: Related to my comment in the other file, this actually does look to be defined twice IIUC.

contracts/libs/0.8.x/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        address _address
+    ) internal pure returns (bool) {
+        return _proof.verifyCalldata(_proofRoot, hashAddress(_address));
+    }
+
+    function projectMaxInvocationsPerAddress(
+        uint256 _projectId,
+        address _coreContract,
+        mapping(address => mapping(uint256 => ProjectConfig))
+            storage projectConfig
```
jakerockland, line 70: Nit: would rename this to be `projectConfigMapping`
```suggestion
            storage projectConfigMapping
```

contracts/libs/0.8.x/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+            return DEFAULT_MAX_INVOCATIONS_PER_ADDRESS;
+        }
+    }
+
+    function setProjectInvocationsPerAddress(
+        uint256 _projectId,
+        address _coreContract,
+        uint24 _maxInvocationsPerAddress,
+        mapping(address => mapping(uint256 => ProjectConfig))
+            storage projectConfig
```
jakerockland, line 87: Nit: would rename this to be `projectConfigMapping`
```suggestion
            storage projectConfigMapping
```

contracts/libs/0.8.x/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        // a value of 0 means no limit
+        // (only used if `useMaxInvocationsPerAddressOverride` is true)
+        uint24 maxInvocationsPerAddressOverride;
+        uint24 maxInvocations;
+        uint256 pricePerTokenInWei;
+    }
+
+    function updateMerkleRoot(
+        mapping(address => mapping(uint256 => bytes32))
+            storage projectMerkleRoot,
```
jakerockland, line 38: Nit: would rename this to be `projectMerkleRootMapping`
```suggestion
            storage projectMerkleRootMapping,
```

contracts/libs/0.8.x/SplitFundsLib.sol
```diff
@@ -0,0 +1,275 @@
...
+import "../../interfaces/0.8.x/IGenArt721CoreContractV3.sol";
+import "../../interfaces/0.8.x/IGenArt721CoreContractV3_Engine.sol";
+
+import "@openzeppelin-4.7/contracts/token/ERC20/IERC20.sol";
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title Art Blocks Split Funds Library
+ * @notice [TO FILL OUT]
```
jakerockland, line 15: When we fill this out, lets make sure we also have natspec comments for the `ProjectConfig` struct and all our methods here as well (ChatGPT is great at writing initial comments btw hah 😄).

contracts/libs/0.8.x/TokenHolderLib.sol
```diff
@@ -0,0 +1,121 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+import "@openzeppelin-4.5/contracts/utils/structs/EnumerableSet.sol";
+
+pragma solidity ^0.8.0;
+
+library TokenHolderLib {
```
jakerockland, line 8: I think there are some compile issues in this one still.

contracts/libs/0.8.x/TokenHolderLib.sol
```diff
@@ -0,0 +1,121 @@
...
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    uint256 constant ONE_MILLION = 1_000_000;
+
+    function isAllowlistedNFT(
+        mapping(uint256 => mapping(address => mapping(uint256 => bool)))
+            storage allowedProjectHolders,
```
jakerockland, line 15: Nit: would rename to `allowedProjectHoldersMapping`
```suggestion
            storage allowedProjectHoldersMapping,
```

contracts/libs/0.8.x/TokenHolderLib.sol
```diff
@@ -0,0 +1,121 @@
...
+        uint256 _contractProjectId,
```
jakerockland, line 16: Should this just be `projectId`?

contracts/libs/0.8.x/TokenHolderLib.sol
```diff
@@ -0,0 +1,121 @@
...
+    function unregisterNFTAddress(
+        EnumerableSet.AddressSet storage _registeredNFTAddresses,
+        address _NFTAddress
+    ) external {
+        _registeredNFTAddresses.remove(_NFTAddress);
+    }
+
+    function allowHoldersOfProjects(
+        mapping(uint256 => mapping(address => mapping(uint256 => bool)))
+            storage allowedProjectHolders,
```
jakerockland, line 43: Nit: would rename to `allowedProjectHoldersMapping`
```suggestion
            storage allowedProjectHoldersMapping,
```

contracts/libs/0.8.x/TokenHolderLib.sol
```diff
@@ -0,0 +1,121 @@
...
+            );
+            allowedProjectHolders[_contractProjectId][_ownedNFTAddresses[i]][
+                _ownedNFTProjectIds[i]
+            ] = true;
+        }
+    }
+
+    function removeHoldersOfProjects(
+        mapping(uint256 => mapping(address => mapping(uint256 => bool)))
+            storage allowedProjectHolders,
```
jakerockland, line 66: Nit: would rename to `allowedProjectHoldersMapping`
```suggestion
            storage allowedProjectHoldersMapping,
```

contracts/libs/0.8.x/TokenHolderLib.sol
jakerockland, line 43: Nit: would rename to `allowedProjectHoldersMapping`
```suggestion
            storage allowedProjectHoldersMapping,
```

contracts/libs/0.8.x/TokenHolderLib.sol
```diff
@@ -0,0 +1,121 @@
...
+            );
+            allowedProjectHolders[_contractProjectId][_ownedNFTAddresses[i]][
+                _ownedNFTProjectIds[i]
+            ] = false;
+        }
+    }
+
+    function allowRemoveHoldersOfProjects(
+        mapping(uint256 => mapping(address => mapping(uint256 => bool)))
+            storage allowedProjectHolders,
```
jakerockland, line 89: Nit: would rename to `allowedProjectHoldersMapping`
```suggestion
            storage allowedProjectHoldersMapping,
```

contracts/libs/0.8.x/TokenHolderLib.sol
```diff
@@ -0,0 +1,121 @@
...
+        removeHoldersOfProjects(
+            _contractProjectId,
+            _ownedNFTAddressesRemove,
+            _ownedNFTProjectIdsRemove
+        );
+    }
+
+    function isAllowlistedNFT(
+        mapping(uint256 => mapping(address => mapping(uint256 => bool)))
+            storage allowedProjectHolders,
```
jakerockland, line 110: Nit: would rename to `allowedProjectHoldersMapping`
```suggestion
            storage allowedProjectHoldersMapping,
```

jakerockland (commented):
From an architectural / design perspective – this approach overall generally LGTM @lyaunzbe 

Handful of comments added from my initial first review pass here some minor suggestions and other clarifying points of conversation.

Overall, would suggest that we get this PR to a state where `SetPriceV5` is implemented in the "vanilla", `Merkle` and `TokenHolder` permutations, with full test coverage, and then can get that merged into `main` and from there go plug-away at porting over each of the other additional minter types in new separate PRs off of `main`. 

That generally seem right ser?

lyaunzbe:
@jakerockland yes that PR approach seems good to me! and thank you for the current comments. This initial pass is much appreciated (I should've mentioned that tokenholder lib was not really in a ready state, but everything else is super helpful)

contracts/interfaces/0.8.x/IFilteredSharedProjectMaxInvocationsLimit.sol
lyaunzbe: yes this is definitely something I wasn't completely sure of, but the initial rationale is related to there being a few minters (polyptych namely if i recall correctly?) that don't/can't really utilize this project max invocations logic. will definitely revisit this as I flesh this out

contracts/libs/0.8.x/MerkleLib.sol
lyaunzbe: the implementation of this emits an event after calling this lib func, https://github.com/ArtBlocks/artblocks-contracts/pull/637/files/88781f12f6fb1a27648c7a4da918feccf6bbacca#diff-28de2faccfae590a400e7771f763d7290f6dc5d5f5fdefd44f6c232a05cba61dR606

ryley-o: Oof, I really want the library to emit the event, but it appears that the only way we can accomplish that is for the library to define its own (duplicate) event _in addition_ to the contract defining the same event ([ref](https://medium.com/aragondec/library-driven-development-in-solidity-2bebcaf88736)). I really don't want to go down the path of defining the same event in multiple places, but I also really want these libraries to handle emitting events.

There is also a possibility that we don't use the library to update state here, but instead only use it to `validateMerkleRoot` (in this case, check that it is non-zero), then update state and emit the event in the core contract. In that architecture, I think it would make more sense for the library to not emit the event.

I think this is worth a quick sync.

contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,684 @@
...
+    string public constant minterType = "MinterSetPriceV5";
+
+    /// minter version for this minter
+    string public constant minterVersion = "v5.0.0";
+
+    uint256 constant ONE_MILLION = 1_000_000;
+
+    /// contractAddress => projectId => project config
+    mapping(address => mapping(uint256 => MerkleLib.ProjectConfig))
+        public projectConfig;
```
ryley-o, line 61: as noted above, we should figure out where we want to define projectConfig and update here accordingly

contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,684 @@
...
+ * ----------------------------------------------------------------------------
+ * The following functions are restricted to a project's artist:
+ * - updatePricePerTokenInWei
+ * - setProjectMaxInvocations
+ * - manuallyLimitProjectMaxInvocations
+ * ----------------------------------------------------------------------------
+ * Additional admin and artist privileged roles may be described on other
+ * contracts that this minter integrates with.
+ */
+contract MinterSetPriceV5Merkle is
```
ryley-o, line 39: nit: I would be okay if we called this `minterSharedMerkleV0` or something...

contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,684 @@
...
+                vault
+            ),
+            "Invalid Merkle proof"
+        );
+
+        // limit mints per address by project
+        uint256 _maxProjectInvocationsPerAddress = _projectConfig
+            .useMaxInvocationsPerAddressOverride
+            ? _projectConfig.maxInvocationsPerAddressOverride
+            : MerkleLib.DEFAULT_MAX_INVOCATIONS_PER_ADDRESS;
```
ryley-o, line 488: I think we could shove logic like this into `MerkleLib.sol`

contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,684 @@
...
+        bytes32[] calldata _proof,
+        address _vault
+    ) external payable returns (uint256 tokenId) {
+        return purchaseTo_do6(_to, _projectId, _coreContract, _proof, _vault);
+    }
+
+    /**
+     * @notice gas-optimized version of purchaseTo(address, uint256).
+     */
+    function purchaseTo_do6(
```
ryley-o, line 419: since we haven't really ended up calling the gas-optimized function names during purchases, I would be in favor in removing them from these new minters if we all agree on that. Would make things a bit more readable.

contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,684 @@
...
+                tokenInvocation <= localMaxInvocations,
+                "Maximum invocations reached"
+            );
+            // in typical case, update the local maxHasBeenInvoked value
+            // to true if the token invocation == minter local max invocations
+            // (enables gas efficient reverts after sellout)
+            if (tokenInvocation == localMaxInvocations) {
+                _projectConfig.maxHasBeenInvoked = true;
+            }
+        }
```
ryley-o, line 534: I think this could be shoved into a `MaxInvocationsLib.sol` if we made one, since possible to do with only a `tokenID` and reference to the `ProjectConfig`

contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,684 @@
...
+    }
+
+    /**
+     * @notice Returns hashed address (to be used as merkle tree leaf).
+     * Included as a public function to enable users to calculate their hashed
+     * address in Solidity when generating proofs off-chain.
+     * @param _address address to be hashed
+     * @return bytes32 hashed address, via keccak256 (using encodePacked)
+     */
+    function hashAddress(address _address) public pure returns (bytes32) {
```
ryley-o, line 621: This is now purely a convenience function for off-chain checks
```suggestion
    function hashAddress(address _address) external pure returns (bytes32) {
```

ryley-o (commented):
this is a good start - I think would be good to set up a quick sync about some of the comments!

apologies for some of the minor nits and quantity of comments - feel free to just resolve if you prefer to focus on the higher-level items instead of minor things like naming, etc. 👍

ryley-o (commented):
this is a good start - I think would be good to set up a quick sync about some of the comments!

apologies for some of the minor nits and quantity of comments - feel free to just resolve if you prefer to focus on the higher-level items instead of minor things like naming, etc. 👍

contracts/libs/0.8.x/MerkleLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        bool priceIsConfigured;
+        // initial value is false, so by default, projects limit allowlisted
+        // addresses to a mint qty of `DEFAULT_MAX_INVOCATIONS_PER_ADDRESS`
+        bool useMaxInvocationsPerAddressOverride;
+        // a value of 0 means no limit
+        // (only used if `useMaxInvocationsPerAddressOverride` is true)
+        uint24 maxInvocationsPerAddressOverride;
+        uint24 maxInvocations;
+        uint256 pricePerTokenInWei;
+    }
```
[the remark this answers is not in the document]
lyaunzbe: as discussed: starting with option 2, but general struct goes in splitfund lib, merkle specific project config goes in merkle lib

contracts/libs/0.8.x/MerkleLib.sol
lyaunzbe: after discussion: deciding to carry-on with current method

lets make it very clear in any library methods where you would expect an event to be emitted to mention that it's on the client (natspec comments)

contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
lyaunzbe: after discussion: let's remove these for now, but make a note for the future to potentially re-introduce when we have more ubiquitous usage of data client library

contracts/interfaces/0.8.x/IFilteredSharedProjectMaxInvocationsLimit.sol
lyaunzbe: deleted this interface with the intention to consolidate in common minter interface

lyaunzbe:
@ryley-o @jakerockland request a re-review here 🙏 , emphasis on MinterSetPriceV5Holder which I believe was not fully ready during the initial review above. thank you!!

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,400 @@
...
+     */
+    function syncProjectMaxInvocationsToCore(
+        uint256 _projectId,
+        address _coreContract
+    ) public {
+        uint256 maxInvocations = MaxInvocationsLib
+            .syncProjectMaxInvocationsToCore(
+                _projectId,
+                _coreContract,
+                maxInvocationsProjectConfig
```
ryley-o, line 123: nit: I think it would be a bit cleaner to pass in the `storage MaxInvocationsLib.MaxInvocationsProjectConfig` directly instead of the entire mapping. It removes the library's concern with how the minter decides to map to the structs, which I think makes the library a little more readable (and sometimes more efficient).
WDYT?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,400 @@
...
+        uint256 _projectId,
+        address _coreContract,
+        uint256 _maxInvocations
+    ) external {
+        _onlyArtist(_projectId, _coreContract);
+        MaxInvocationsLib.manuallyLimitProjectMaxInvocations(
+            _projectId,
+            _coreContract,
+            _maxInvocations,
+            maxInvocationsProjectConfig
```
ryley-o, line 154: (same comment about passing the storage reference to struct instead of entire mapping)

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,400 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
ryley-o, line 1: nit: In general, we should order all functions in the minters according to: https://docs.soliditylang.org/en/latest/style-guide.html#order-of-functions

Additionally, similar to how you are already doing with the state variables, I think it would be good to denote sections of functions that are related to the different libraries, which might help when we plug in the different libraries to different minters.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,400 @@
...
+        address _coreContract
+    ) external payable returns (uint256 tokenId) {
+        tokenId = purchaseTo(msg.sender, _projectId, _coreContract);
+        return tokenId;
+    }
+
+    /**
+     * @notice gas-optimized version of purchase(uint256).
+     */
+    function purchase_H4M(
```
ryley-o, line 275: let's remove these gas-optimized purchase functions (for simplicity)

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,400 @@
...
+
+        // Note that `maxHasBeenInvoked` is only checked here to reduce gas
+        // consumption after a project has been fully minted.
+        // `_projectConfig.maxHasBeenInvoked` is locally cached to reduce
+        // gas consumption, but if not in sync with the core contract's value,
+        // the core contract also enforces its own max invocation check during
+        // minting.
+        require(
+            !_maxInvocationsProjectConfig.maxHasBeenInvoked,
+            "Maximum number of invocations reached"
```
ryley-o, line 308: nit: in general, let's keep error string messages to <=31 characters for gas + bytecode efficiency

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,400 @@
...
+            _to,
+            _projectId,
+            _coreContract,
+            msg.sender
+        );
+
+        // invocation is token number plus one, and will never overflow due to
+        // limit of 1e6 invocations per project. block scope for gas efficiency
+        // (i.e. avoid an unnecessary var initialization to 0).
+        unchecked {
```
ryley-o, line 333: can we use `MaxInvocationsLib.purchaseEffectsInvocations(...)` here instead of this unchecked block?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Holder.sol
```diff
@@ -0,0 +1,795 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
ryley-o, line 1: nit: I think these "extension" minters should probably be named `MinterSetPriceHolderV5` (or whatever version number is appropriate) to enable us to roll out a minter with new holder logic without rolling the version of MinterSetPrice. Thoughts?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Holder.sol
```diff
@@ -0,0 +1,795 @@
...
+    }
+
+    /**
+     *
+     * @notice Registers holders of NFTs at address `_NFTAddress` to be
+     * considered for minting. New core address is assumed to follow syntax of:
+     * `projectId = tokenId / 1_000_000`
+     * @param _NFTAddress NFT core address to be registered.
+     */
+    function registerNFTAddress(
```
ryley-o, line 193: Since this is a contract-level minter setting, we don't have a way to currently index this. I think we had discussed just removing this constraint altogether, and settled on removing this guardrail and relying on the artist to only allowlist appropriate projects. Does that sound good to you? (same goes for `unregisterNFTAddress`)

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,400 @@
...
+
+        return tokenId;
+    }
+
+    /**
+     * @notice Gets if price of token is configured, price of minting a
+     * token on project `_projectId`, and currency symbol and address to be
+     * used as payment. Supersedes any core contract price information.
+     * @param _projectId Project ID to get price information for.
+     * @return isConfigured true only if token price has been configured on
```
ryley-o, line 370:
```suggestion
     * @param _coreContract Contract address of the core contract.
     * @return isConfigured true only if token price has been configured on
```

ryley-o (commented):
ty @lyaunzbe! looking good, and in general, really like how this is organized and coming together. A few comments/discussion topics highlighted below 🙏

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,400 @@
...
+    mapping(address => mapping(uint256 => ProjectConfig)) public projectConfig;
+
+    // STATE VARIABLES FOR SplitFundsLib
+    // contractAddress => IsEngineCache
+    mapping(address => SplitFundsLib.IsEngineCache) private _isEngineCaches;
+
+    // STATE VARIABLES FOR MaxInvocationsLib
+    /// contractAddress => projectId => max invocations specific project config
+    mapping(address => mapping(uint256 => MaxInvocationsLib.MaxInvocationsProjectConfig))
+        public maxInvocationsProjectConfig;
```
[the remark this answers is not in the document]
lyaunzbe: great call, will definitely do this 👍

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,400 @@
...
+        ];
+        if (isEngineCache.isCached) {
+            return isEngineCache.isEngine;
+        } else {
+            bool isEngine = SplitFundsLib.getV3CoreIsEngine(
+                _coreContract,
+                isEngineCache
+            );
+            return isEngine;
+        }
```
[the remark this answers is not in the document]
lyaunzbe: ah yah that is a good point. i think the placing here was related to some initial thought here to continue to publicly expose isEngine as a getter on the minters in general, but that's not currently being achieved here anyway

packages/contracts/contracts/libs/0.8.x/TokenHolderLib.sol
```diff
@@ -0,0 +1,191 @@
...
+    ) internal {
+        require(
+            _ownedNFTAddresses.length == _ownedNFTProjectIds.length,
+            "TokenHolderLib: arrays must be same length"
+        );
+        for (uint256 i = 0; i < _ownedNFTAddresses.length; i++) {
+            require(
+                _registeredNFTAddresses.contains(_ownedNFTAddresses[i]),
+                "TokenHolderLib: address not registered"
+            );
```
jakerockland, line 113: Hrm, I think we may only need this check on adds, not on removes?

packages/contracts/contracts/libs/0.8.x/TokenHolderLib.sol
```diff
@@ -0,0 +1,191 @@
...
+     * must align with `_ownedNFTAddressesAdd`.
+     * @param _ownedNFTAddressesRemove NFT core addresses of projects to be
+     * removed from allowlist. Indexes must align with
+     * `_ownedNFTProjectIdsRemove`.
+     * @param _ownedNFTProjectIdsRemove Project IDs on
+     * `_ownedNFTAddressesRemove` whose holders will be removed from allowlist
+     * to mint project `_projectId`. Indexes must align with
+     * `_ownedNFTAddressesRemove`.
+     */
+    function allowRemoveHoldersOfProjects(
```
jakerockland, line 139: Nit: may suggest naming this `updateHoldersOfProjects` or `allowAndRemoveHoldersOfProjects`

packages/contracts/contracts/libs/0.8.x/SplitFundsLib.sol
```diff
@@ -0,0 +1,284 @@
...
+                additionalPayeePrimaryAddress_,
+                additionalPayeePrimaryRevenue_
+            );
+        }
+    }
+
+    /**
+     * @notice Returns whether a V3 core contract is an Art Blocks Engine
+     * contract or not. Return value of false indicates that the core is a
+     * flagship contract.
```
jakerockland, line 236: We should definitely make it clear here in the documentation that this method will modify the cache, not just read it.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Holder.sol
```diff
@@ -0,0 +1,795 @@
...
+    address public immutable delegationRegistryAddress;
+
+    /// Delegation registry address
+    IDelegationRegistry private immutable delegationRegistryContract;
+
+    /// minterType for this minter
+    string public constant minterType = "MinterHolderV4";
+
+    /// minter version for this minter
+    string public constant minterVersion = "v4.1.0";
```
jakerockland:
```suggestion
    string public constant minterVersion = "v5.0.0";
```

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Holder.sol
```diff
@@ -0,0 +1,795 @@
...
+    IMinterFilterV1 private immutable minterFilter;
+
+    /// Delegation registry address
+    address public immutable delegationRegistryAddress;
+
+    /// Delegation registry address
+    IDelegationRegistry private immutable delegationRegistryContract;
+
+    /// minterType for this minter
+    string public constant minterType = "MinterHolderV4";
```
jakerockland:
```suggestion
    string public constant minterType = "MinterSetPriceHolderV5";
```

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Holder.sol
```diff
@@ -0,0 +1,795 @@
...
+    // STATE VARIABLES FOR SplitFundsLib
+    // contractAddress => IsEngineCache
+    mapping(address => SplitFundsLib.IsEngineCache) private _isEngineCaches;
+
+    // STATE VARIABLES FOR MaxInvocationsLib
+    /// contractAddress => projectId => max invocations specific project config
+    mapping(address => mapping(uint256 => MaxInvocationsLib.MaxInvocationsProjectConfig))
+        public maxInvocationsProjectConfig;
+
+    /// Set of core contracts allowed to be queried for token holders
```
jakerockland, line 107:
```suggestion
    // STATE VARIABLES FOR TokenHolderLib
    /// Set of core contracts allowed to be queried for token holders
```

jakerockland: In addition to this suggestion, would probably suggest that we use a pattern of something like:

```
// STATE VARIABLES FOR {LibraryName}Lib begin here

...

// STATE VARIABLES FOR {LibraryName}Lib end here
```

Here and throughout such uses.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,650 @@
...
+            MerkleLib.DEFAULT_MAX_INVOCATIONS_PER_ADDRESS
+        );
+    }
+
+    /**
+     * @notice Returns whether or not the provided address `_coreContract`
+     * is an Art Blocks Engine core contract. Caches the result for future access.
+     * @param _coreContract Address of the core contract to check.
+     */
+    function _isEngine(address _coreContract) internal returns (bool) {
```
jakerockland, line 126: Should this be used at all vs. just using the library?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,650 @@
...
+            _proof,
+            address(0)
+        );
+        return tokenId;
+    }
+
+    /**
+     * @notice gas-optimized version of purchase(uint256).
+     */
+    function purchase_H4M(
```
jakerockland, line 337: I don't think we need to keep the gas-optimized variants (per the other comment to the same effect).

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,650 @@
...
+    IFilteredSharedMerkle
+{
+    /// Minter filter address this minter interacts with
+    address public immutable minterFilterAddress;
+
+    /// Minter filter this minter may interact with.
+    IMinterFilterV1 private immutable minterFilter;
+
+    /// minterType for this minter
+    string public constant minterType = "MinterSetPriceV5";
```
jakerockland, line 51:
```suggestion
    string public constant minterType = "MinterSetPriceMerkleV5";
```

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,650 @@
...
+
+    // STATE VARIABLES FOR MerkleLib
+    /// Delegation registry address
+    address public immutable delegationRegistryAddress;
+    /// Delegation registry address
+    IDelegationRegistry private immutable delegationRegistryContract;
+
+    /// contractAddress => projectId => merkle specific project config
+    mapping(address => mapping(uint256 => MerkleLib.MerkleProjectConfig))
+        public merkleProjectConfig;
```
jakerockland, line 77: Similar feedback to that in token holder re: usage of comment blocks.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Holder.sol
```diff
@@ -0,0 +1,795 @@
...
+
+        // EFFECTS
+        tokenId = minterFilter.mint_joo(_to, _projectId, _coreContract, vault);
+
+        // NOTE: delegate-vault handling **ends here**.
+
+        // invocation is token number plus one, and will never overflow due to
+        // limit of 1e6 invocations per project. block scope for gas efficiency
+        // (i.e. avoid an unnecessary var initialization to 0).
+        unchecked {
```
jakerockland, line 692: Same question/note as in https://github.com/ArtBlocks/artblocks-contracts/pull/637/files#r1210784290

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
```diff
@@ -0,0 +1,650 @@
...
+            _to,
+            _projectId,
+            _coreContract,
+            msg.sender
+        );
+
+        // invocation is token number plus one, and will never overflow due to
+        // limit of 1e6 invocations per project. block scope for gas efficiency
+        // (i.e. avoid an unnecessary var initialization to 0).
+        unchecked {
```
jakerockland, line 482: Same question/note as in https://github.com/ArtBlocks/artblocks-contracts/pull/637/files#r1210784290

jakerockland (approved):
Things here generally LGTM from an overall code structure / architecture perspective. Still some additional clean up / refactor needed, happy to do another review pass when that is done and tests are updated 🙏

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Holder.sol
lyaunzbe: Yes i agree 👍 👍  (kinda looks funky to not have the version be the last bit) will update

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
lyaunzbe: ah yep, I wrote the lib method but forgot to actually add it 😅 added now across all minter variants 👍

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
lyaunzbe: ah yes I think I had some initial confusion/misunderstanding about applying storage updates to a struct that's nested inside a mapping, but yes there should be no issue/this makes a lot more sense to do :)

packages/contracts/contracts/libs/0.8.x/TokenHolderLib.sol
lyaunzbe: resolving this b/c we decided to remove registration/unregistration entirely

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5Merkle.sol
lyaunzbe: moved this entirely to the lib, except for a new view func, `isEngineView`, on the minter to allow for easy viewing of this data (non-modifying vs the funcs that set cache)

lyaunzbe:
@jakerockland @ryley-o I believe this should be ready for another review  at your convenience 🙏 much appreciated. main focus was remaining tests/code review fixes discussed above

lyaunzbe:
looks like there are still some circle CI errors (some look just lint related on older/legacy files) will clean that up shortly as well

packages/contracts/contracts/libs/v0.8.x/minter-libs/TokenHolderLib.sol
```diff
@@ -0,0 +1,151 @@
...
+     * allowlisted. Indexes must align with `_ownedNFTProjectIds`.
+     * @param _ownedNFTProjectIds Project IDs on `_ownedNFTAddresses` whose
+     * holders shall be allowlisted to mint project `_projectId`. Indexes must
+     * align with `_ownedNFTAddresses`.
+     */
+    function allowHoldersOfProjects(
+        mapping(address => mapping(uint256 => mapping(address => mapping(uint256 => bool))))
+            storage allowedProjectHoldersMapping,
+        uint256 _projectId,
+        address _coreContract,
```
ryley-o, line 39: These large mappings can become a bit difficult to understand.
WDYT about doing the following:
```
// in this library, add the following struct, and names space per-project in the minter contract
// (same as we do for the Merkle project config):
struct  HolderProjectConfig {
    // projects whose holders are allowed to purchase a token on `projectId`
    // ownedNFTAddress => ownedNFTProjectIds => bool
    mapping(address => mapping(uint256 => bool)) allowedProjectHolders
}

// then in this function, we pass the struct directly, and can 
// remove the `_projectId` and `_coreContract` args 
function allowHoldersOfProjects(
    HolderProjectConfig storage holderProjectConfig,
    address[] memory _ownedNFTAddresses,
    uint256[] memory _ownedNFTProjectIds
) internal {
    ...
}
```

The above pattern also enables the `_projectId` and `_coreContract` input args to be removed from `removeHoldersOfProjects`, `allowAndRemoveHoldersOfProjects`, and `isAllowlistedNFT`.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
```diff
@@ -0,0 +1,790 @@
...
+     * Indexes must align with `_ownedNFTAddresses`.
+     */
+    function removeHoldersOfProjects(
+        uint256 _projectId,
+        address _coreContract,
+        address[] memory _ownedNFTAddresses,
+        uint256[] memory _ownedNFTProjectIds
+    ) external {
+        _onlyArtist(_projectId, _coreContract);
+        // require same length arrays
```
ryley-o, line 301:
```suggestion
```

packages/contracts/contracts/libs/v0.8.x/minter-libs/SplitFundsLib.sol
```diff
@@ -0,0 +1,354 @@
...
+     * flagship contract. This function does not update the cache state for the
+     * given V3 core contract.
+     * @dev this function reverts if a core contract does not return the
+     * expected number of return values from getPrimaryRevenueSplits() for
+     * either a flagship or engine core contract.
+     * @dev this function uses the length of the return data (in bytes) to
+     * determine whether the core is an engine or not.
+     * @param genArt721CoreV3 The address of the deployed core contract.
+     */
+    function getV3CoreIsEngine(
```
ryley-o, line 298: I would be okay not overloading this function name, and instead calling this one `getV3CoreIsEngineView`

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
```diff
@@ -0,0 +1,790 @@
...
+        if (
+            _maxInvocationsProjectConfig.maxInvocations == 0 &&
+            _maxInvocationsProjectConfig.maxHasBeenInvoked == false
+        ) {
+            syncProjectMaxInvocationsToCore(_projectId, _coreContract);
+        }
+    }
+
+    /**
+     * @notice Allows holders of NFTs at addresses `_ownedNFTAddresses`,
```
ryley-o, line 243: Now that artists have full control of which contracts to allowlist, I think we should have a large warning here that indicates that only Art Blocks and/or Art Blocks Engine contracts may be used, due to assumptions about how token ID and project ID relate to each other.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
```diff
@@ -0,0 +1,790 @@
...
+     * @param _contractAddress The address of the contract where the data is stored.
+     * @param _projectId The ID of the project whose data needs to be fetched.
+     * @return ProjectConfig instance with the project configuration data.
+     */
+    function projectConfig(
+        address _contractAddress,
+        uint256 _projectId
+    ) external view returns (ProjectConfig memory) {
+        return _projectConfigMapping[_contractAddress][_projectId];
+    }
```
ryley-o, line 472: if we add the `struct HolderProjectConfig`, we should add a view for it (analogous to what you did on Merkle minter)

ryley-o (commented):
This is really great!

All minor nits other than I think one comment with some substance 🎉 

Would like to take a final pass once tests are finished up (would like perform a detailed diff to ensure logic is consistent, observe coverage, etc.).

packages/contracts/contracts/libs/v0.8.x/minter-libs/SplitFundsLib.sol
jakerockland: I like that idea.

packages/contracts/contracts/libs/v0.8.x/minter-libs/TokenHolderLib.sol
jakerockland: This is a great suggestion – thank you @ryley-o

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
```diff
@@ -0,0 +1,790 @@
...
+            require(isValidVault, "Invalid delegate-vault pairing");
+            vault = _vault;
+        }
+
+        // EFFECTS
+        tokenId = minterFilter.mint_joo(_to, _projectId, _coreContract, vault);
+
+        // NOTE: delegate-vault handling **ends here**.
+
+        MaxInvocationsLib.purchaseEffectsInvocations(
```
jakerockland, line 757:
```suggestion
        MaxInvocationsLib.validatePurchaseEffectsInvocations(
```

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
```diff
@@ -0,0 +1,790 @@
...
+    // STATE VARIABLES FOR SplitFundsLib begin here
+    // contractAddress => IsEngineCache
+    mapping(address => SplitFundsLib.IsEngineCache) private _isEngineCaches;
+    // STATE VARIABLES FOR SplitFundsLib end here
+
+    // STATE VARIABLES FOR MaxInvocationsLib begin here
+    // contractAddress => projectId => max invocations specific project config
+    mapping(address => mapping(uint256 => MaxInvocationsLib.MaxInvocationsProjectConfig))
+        private _maxInvocationsProjectConfigMapping;
+
```
jakerockland, line 110: Nit: delete this blank line

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
```diff
@@ -0,0 +1,790 @@
...
+     */
+    mapping(address => mapping(uint256 => mapping(address => mapping(uint256 => bool))))
+        private _allowedProjectHoldersMapping;
+
+    // STATE VARIABLES FOR SplitFundsLib begin here
+    // contractAddress => IsEngineCache
+    mapping(address => SplitFundsLib.IsEngineCache) private _isEngineCaches;
+    // STATE VARIABLES FOR SplitFundsLib end here
+
+    // STATE VARIABLES FOR MaxInvocationsLib begin here
```
jakerockland, line 106:
```suggestion
    ////////////////////////////////////////////////////////////////////////////////////////////////////////////
    // STATE VARIABLES FOR MaxInvocationsLib begin here
    ////////////////////////////////////////////////////////////////////////////////////////////////////////////
```

jakerockland: Nit suggestion to do something to this effect here and throughout this PR to improve readability of these blocks.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
jakerockland: +++; great call out @ryley-o (we could even do a more formal rough heuristic check of this within the minter, and throw an error if we get things beyond expectations?).

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceMerkleV5.sol
```diff
@@ -0,0 +1,773 @@
...
+        }
+
+        tokenId = minterFilter.mint_joo(
+            _to,
+            _projectId,
+            _coreContract,
+            msg.sender
+        );
+
+        MaxInvocationsLib.purchaseEffectsInvocations(
```
jakerockland, line 754:
```suggestion
        MaxInvocationsLib.validatePurchaseEffectsInvocations(
```

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,417 @@
...
+
+        // EFFECTS
+        tokenId = minterFilter.mint_joo(
+            _to,
+            _projectId,
+            _coreContract,
+            msg.sender
+        );
+
+        MaxInvocationsLib.purchaseEffectsInvocations(
```
jakerockland, line 398:
```suggestion
        MaxInvocationsLib.validatePurchaseEffectsInvocations(
```

jakerockland (approved):
Smaller nits this time – feeling good about this as a template for the additional minters and would like to get this merged once tests are fully g2g so that we can do those iteratively.

package.json
```diff
@@ -19,8 +19,10 @@
     "build": "yarn workspace @artblocks/contracts run build && yarn workspace @artblocks/sdk run build",
     "codegen:project": "graphql-codegen dotenv_config_path=./.env --require dotenv/config --config graphql.config.ts --project",
     "prepare": "husky install",
-    "pre-commit": "yarn workspace @artblocks/contracts run pre-commit && yarn workspace @artblocks/sdk run pre-commit",
-    "pre-push": "yarn workspace @artblocks/sdk run pre-push",
+    "pre-commit": "echo",
+    "pre-push": "echo",
+    "pre-commitExempt": "yarn workspace @artblocks/contracts run pre-commit && yarn workspace @artblocks/sdk run pre-commit",
+    "pre-pushExempt": "yarn workspace @artblocks/sdk run pre-push",
```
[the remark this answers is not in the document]
lyaunzbe: ah yes 😅 purely for dev convenience at the moment..... will leave for now, but keeping this as unresolved to remind myself to revert before final merge/review

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
lyaunzbe: 109 is a wrapped line and it seems like our prettier config/formatter automatically adds that blank line on 110

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
lyaunzbe: @ryley-o i think the view for this data is actually already satisfied by the existing view `allowedProjectHolders`

lyaunzbe:
@ryley-o @jakerockland requested fixes have been applied and tests should be g2g 🙏 🙏 re-review much appreciated 🙏 🙏

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
ryley-o: ah, yes! resolving 👍

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
```diff
@@ -0,0 +1,794 @@
...
+
+    /**
+     * @notice Removes holders of NFTs at addresses `_ownedNFTAddresses`,
+     * project IDs `_ownedNFTProjectIds` to mint on project `_projectId`. If
+     * other projects owned by a holder are still allowed to mint, holder will
+     * maintain ability to purchase.
+     * `_ownedNFTAddresses` assumed to be aligned with `_ownedNFTProjectIds`.
+     * e.g. Removes holders of project `_ownedNFTProjectIds[0]` on token
+     * contract `_ownedNFTAddresses[0]` from mint allowlist of `_projectId`.
+     * @param _projectId Project ID to enable minting on.
```
ryley-o, line 297: nit: add param for core contract (comment applies to a few functions in this minter)

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
```diff
@@ -0,0 +1,434 @@
...
+     * @param _coreContract Core contract address for the given project.
+     * @param _projectId Project ID to set the maximum invocations for.
+     * @dev this enables gas reduction after maxInvocations have been reached -
+     * core contracts shall still enforce a maxInvocation check during mint.
+     */
+    function syncProjectMaxInvocationsToCore(
+        uint256 _projectId,
+        address _coreContract
+    ) public {
+        uint256 maxInvocations = MaxInvocationsLib
```
ryley-o, line 353: IMPORTANT: needs `_onlyArtist(_projectId, _coreContract);`
note: the other two minters are okay

Also, can we ensure all function permissions are tested? I think we should be catching this sort of thing in our test suite - for example, a test that checks that a non-artist is not able to call this function.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
```diff
@@ -0,0 +1,794 @@
...
+        require(
+            !_maxInvocationsProjectConfig.maxHasBeenInvoked,
+            "Max invocations reached"
+        );
+
+        // load price of token into memory
+        uint256 pricePerTokenInWei = _projectConfig.pricePerTokenInWei;
+
+        require(msg.value >= pricePerTokenInWei, "Min value to mint req.");
+
```
ryley-o, line 720: super minor nit: can this be moved after priceIsConfigured check? that keeps the diff between `MinterSetPriceV5` a little cleaner.

ryley-o (commented):
Awesome - I took a pretty detailed look at these, mostly smol nits, one important finding.

Overall I really, really like the test organization that you created! @lyaunzbe - I would like to quickly sync to go over the coverage details once you feel like all the tests have been added 🙏

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceMerkleV5.sol
```diff
@@ -0,0 +1,778 @@
...
+     * @param _coreContract Core contract address for the given project.
+     * @param _projectId Project ID to toggle the mint limit.
+     * @param _maxInvocationsPerAddress Maximum allowed invocations per
+     * allowlisted address.
+     * @dev default value stated above must be updated if the value of
+     * CONFIG_USE_MAX_INVOCATIONS_PER_ADDRESS_OVERRIDE is changed.
+     */
+    function setProjectInvocationsPerAddress(
+        address _coreContract,
+        uint256 _projectId,
```
ryley-o, line 277: nit: order of args - we usually have projectId, then coreContract

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceMerkleV5.sol
```diff
@@ -0,0 +1,778 @@
...
+     * @dev This function retrieves the number of times a purchaser has minted
+     * in a specific project from the projectUserMintInvocationsMapping.
+     * @param _coreContract The address of the core contract.
+     * @param _projectId The ID of the project.
+     * @param _purchaser The address of the purchaser.
+     * @return uint256 The number of times the purchaser has minted in the given project.
+     */
+    function projectUserMintInvocations(
+        address _coreContract,
+        uint256 _projectId,
```
ryley-o, line 412: nit: order of args - we usually have projectId, then coreContract

packages/contracts/contracts/interfaces/v0.8.x/IFilteredSharedHolder.sol
```diff
@@ -0,0 +1,79 @@
...
+     * considered for minting.
+     */
+    event UnregisteredNFTAddress(address indexed _NFTAddress);
+    /**
+     * @notice Allow holders of NFTs at addresses `_ownedNFTAddresses`, project
+     * IDs `_ownedNFTProjectIds` to mint on project `_projectId`.
+     * `_ownedNFTAddresses` assumed to be aligned with `_ownedNFTProjectIds`.
+     * e.g. Allows holders of project `_ownedNFTProjectIds[0]` on token
+     * contract `_ownedNFTAddresses[0]` to mint.
+     */
```
ryley-o, line 28: These should be removed now that we no longer are registering/unregistering NFT addresses on-chain

packages/contracts/contracts/interfaces/v0.8.x/IFilteredSharedHolder.sol
```diff
@@ -0,0 +1,79 @@
...
+    event AllowedHoldersOfProjects(
+        uint256 indexed _projectId,
+        address _coreContract,
```
ryley-o, line 31:
```suggestion
        address indexed _coreContract,
```

packages/contracts/contracts/interfaces/v0.8.x/IFilteredSharedHolder.sol
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * @notice Remove holders of NFTs at addresses `_ownedNFTAddresses`,
+     * project IDs `_ownedNFTProjectIds` to mint on project `_projectId`.
+     * `_ownedNFTAddresses` assumed to be aligned with `_ownedNFTProjectIds`.
+     * e.g. Removes holders of project `_ownedNFTProjectIds[0]` on token
+     * contract `_ownedNFTAddresses[0]` from mint allowlist.
+     */
+    event RemovedHoldersOfProjects(
+        uint256 indexed _projectId,
+        address _coreContract,
```
ryley-o, line 44:
```suggestion
        address indexed _coreContract,
```

packages/contracts/contracts/interfaces/v0.8.x/IFilteredSharedHolder.sol
```diff
@@ -0,0 +1,79 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
ryley-o, line 1: Thoughts on calling this `ISharedMinterHolderV0.sol`?

packages/contracts/contracts/interfaces/v0.8.x/IFilteredSharedMerkle.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
ryley-o, line 1: Thoughts on calling this `ISharedMinterHolderV0.sol`?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceV5.sol
lyaunzbe: have no validated that negative tests now exists for all funcs using modifiers acrossthese new variants

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceHolderV5.sol
```diff
@@ -0,0 +1,792 @@
...
+                IGenArt721CoreContractV3_Base(_coreContract)
+                    .projectIdToArtistAddress(_projectId),
+            "Only Artist"
+        );
+    }
+
+    /**
+     * @notice Initializes contract to be a Filtered Minter for
+     * `_minterFilter`, integrated with Art Blocks core contract
+     * at address `_genArt721Address`.
```
ryley-o, line 154:
```suggestion
     * @notice Initializes contract to be a Filtered Minter for
     * `_minterFilter`
```

packages/contracts/contracts/interfaces/v0.8.x/IFilteredSharedHolder.sol
lyaunzbe: sgtm will update to reflect

packages/contracts/contracts/interfaces/v0.8.x/IFilteredSharedMerkle.sol
lyaunzbe: sgtm will update to reflect

ryley-o (approved):
LGTM!

packages/contracts/contracts/interfaces/v0.8.x/ISharedMinterHolderV0.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title This interface adds support for including token holder gating when purchasing.
+ * @author Art Blocks Inc.
+ */
+interface IFilteredSharedHolder {
```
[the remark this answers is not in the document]
lyaunzbe: 😅

packages/contracts/test/dependency-registry/DependencyRegistryV0.test.ts
```diff
@@ -30,11 +30,11 @@ const coreContractsToTest = [
...
-interface DependencyRegistryV0TestContext extends Mocha.Context {
-  dependencyRegistry: DependencyRegistryV0;
-  genArt721Core: GenArt721CoreV3;
-  adminACL: AdminACLV0;
-}
+// interface DependencyRegistryV0TestContext extends Mocha.Context {
+//   dependencyRegistry: DependencyRegistryV0;
+//   genArt721Core: GenArt721CoreV3;
+//   adminACL: AdminACLV0;
+// }
```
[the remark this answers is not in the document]
lyaunzbe: oof thank you for catching this (def needs to be undone)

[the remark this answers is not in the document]
lyaunzbe: leaving this as it was prior

packages/contracts/test/minter-suite-minters/SetPrice/MinterSetPrice/configure.test.ts
```diff
@@ -0,0 +1,336 @@
...
+  },
+  // {
+  //   core: "GenArt721CoreV3_Explorations",
+  // },
+  // {
+  //   core: "GenArt721CoreV3_Engine",
+  // },
+  // {
+  //   core: "GenArt721CoreV3_Engine_Flex",
+  // },
```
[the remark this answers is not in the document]
lyaunzbe: finally uncommenting all of these :)

lyaunzbe:
posting screenshot of local coverage  run outputs, focus on new files added specifically:

Screenshot 2023-06-27 at 12 27 12 AM
Screenshot 2023-06-27 at 12 26 57 AM

NOTE that the lower coverage for SplitFundsLib is due to the fact that we have not yet merged in the MinterSetPriceERC20 minter variants that utilize some of the functions in that contract

package.json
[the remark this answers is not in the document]
ryley-o: @lyaunzbe - I've pushed an update to revert these, resolving 👍

PR #648, opened 2023-04-19
Library PoC

ryley-o:
## Description

This is a proof of concept of how a library could be used instead of a base class for the `MinterSetPriceV4` contract.

Will probably close this (archive) eventually, but leaving up for a bit to facilitate ongoing architecture discussions.

base class diff: https://www.diffchecker.com/6fM1a8Sc/

Minter set price diff: https://www.diffchecker.com/ETfnhy59/


ryley-o:
closing this to clean up open PRs.
In general, we partially use this concept in the new minter suite, with the exception that we (use solidity to automatically) define storage locations in the minter concrete contract instead using a manually configured storage location defined in the library (i.e. the `s()` function in this PoC's library), which we think is more readable for the non-solidity dev.

PR #661, opened 2023-04-25
SEA minter code review - force send ETH

ryley-o:
## Description of the change

Update SEA minter to force send ETH instead of relying on WETH as a fallback.

Note: This is future-proof with the current plan outlined in EIP-4758, which replaces SELFDESTRUCT with SENDALL

### DIFF
--- a/contracts/minter-suite/Minters/MinterSEAV0.sol
+++ b/contracts/minter-suite/Minters/MinterSEAV0.sol
@@ -3,8 +3,6 @@
 
 pragma solidity 0.8.17;
 
-import {IWETH} from "../../interfaces/0.8.x/IWETH.sol";
-
 import "../../interfaces/0.8.x/IGenArt721CoreContractV3_Base.sol";
 import "../../interfaces/0.8.x/IMinterFilterV0.sol";
 import "../../interfaces/0.8.x/IFilteredMinterSEAV0.sol";
@@ -115,11 +113,6 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
     /// minter version for this minter
     string public constant minterVersion = "v0.0.1";
 
-    /// The public WETH contract address
-    /// @dev WETH is used as fallback payment method when ETH transfers are
-    /// failing during bidding process (e.g. receive function is not payable)
-    IWETH public immutable weth;
-
     uint256 constant ONE_MILLION = 1_000_000;
 
     mapping(uint256 => ProjectConfig) public projectConfig;
@@ -193,13 +186,11 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
      * which this contract will be a minter.
      * @param _minterFilter Minter filter for which
      * this will a filtered minter.
-     * @param _wethAddress The WETH contract address to use for fallback
      * payment method when ETH transfers are failing during bidding process
      */
     constructor(
         address _genArt721Address,
-        address _minterFilter,
-        address _wethAddress
+        address _minterFilter
     ) ReentrancyGuard() MinterBase(_genArt721Address) {
         genArt721CoreAddress = _genArt721Address;
         genArtCoreContract_Base = IGenArt721CoreContractV3_Base(
@@ -211,7 +202,6 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
             minterFilter.genArt721CoreAddress() == _genArt721Address,
             "Illegal contract pairing"
         );
-        weth = IWETH(_wethAddress);
         // emit events indicating default minter configuration values
         emit AuctionDurationSecondsRangeUpdated({
             minAuctionDurationSeconds: minAuctionDurationSeconds,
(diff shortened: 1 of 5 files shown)


ryley-o:
please ignore the minor coverage blip (we have coverage of new functionality)

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -1110,15 +1104,46 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
...
+        // load state variable into memory for use in inline assembly
+        uint256 minterRefundGasLimit_ = minterRefundGasLimit;
+        // Manually inlined because the compiler doesn't inline functions with
+        // branches.
+        /// @solidity memory-safe-assembly
+        assembly {
+            // If insufficient balance, revert.
+            if lt(selfbalance(), _amount) {
+                // Store the function selector of `ETHTransferFailed()`.
+                mstore(0x00, 0xb12d13eb)
```
jakerockland, line 1128: Where is this selector defined?

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -1110,15 +1104,46 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
...
-            weth.transfer(to, amount);
+    function _forceSafeTransferETH(address _to, uint256 _amount) internal {
+        // load state variable into memory for use in inline assembly
+        uint256 minterRefundGasLimit_ = minterRefundGasLimit;
+        // Manually inlined because the compiler doesn't inline functions with
+        // branches.
+        /// @solidity memory-safe-assembly
+        assembly {
+            // If insufficient balance, revert.
+            if lt(selfbalance(), _amount) {
```
jakerockland, line 1126: Does this actually need to be done in assembly magic vs. outside of the assembly block?

jakerockland: E.g. check `address(this).balance`.

That said, I don't know that we even need this check at all. 🤔

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -1110,15 +1104,46 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
...
+                mstore8(0x20, 0xff) // Opcode `SELFDESTRUCT`.
+                // We can directly use `SELFDESTRUCT` in the contract creation.
+                // Compatible with `SENDALL`:
+                // https://eips.ethereum.org/EIPS/eip-4758
+                if iszero(create(_amount, 0x0b, 0x16)) {
+                    // For better gas estimation.
+                    if iszero(gt(gas(), 1000000)) {
+                        revert(0, 0)
+                    }
+                }
```
jakerockland, line 1145: Can we add a stack diagram here similar to the one we have in BytecodeStorage?

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: good call, this should never be encountered, so this check has been removed entirely 🙏

jakerockland: Fantastic – SGTM ser. 

May be worth leaving a comment here noting that we explicitly don't do this check as it should never be a valid reachable state.

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: the function selector is calculated by hashing the string, but I don't know why storing it at memory location 0x00 would do anything since the next step is to immediately revert.
This entire check has been removed, so I'm okay not diving deeper here, but this specific operation doesn't make sense to me. I'll include a question about this as part of our code review response 🙏

jakerockland: That SGTM ser – fantastic! TYVM!!

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: updated - let me know if this makes sense to you!

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: ohhh, actually this makes sense -
- `mstore` stores 32 byes, so memory looks like: `0x00000000000000000000000000000000000000000000000000000000b12d13eb`
- revert starts at bytes 28 in memory, and has width of 4 bytes, so return data is simply the function selector of `ETHTransferFailed()`, `0xb12d13eb`

I'm guessing services like etherscan are smart enough to have common functions reverse-mapped, and this would probably show up on their site with a revert reason of `ETHTransferFailed()`

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: @jakerockland if the update with the stack diagram looks good to you, I will resolve and merge this 🙏

contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -1110,15 +1104,62 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
...
+            // Transfer the ETH and check if it succeeded or not.
+            if iszero(call(minterRefundGasLimit_, _to, _amount, 0, 0, 0, 0)) {
+                // if the transfer failed, we create a temporary contract with
+                // initialization code that uses `SELFDESTRUCT` to force send
+                // the ETH.
+                // note: Compatible with `SENDALL`:
+                // https://eips.ethereum.org/EIPS/eip-4758
+
+                //---------------------------------------------------------------------------------------------------------------//
+                // Opcode  | Opcode + Arguments  | Description  | Stack View                                                     //
```
jakerockland, line 1139: Nit: align "Stack View"

jakerockland (approved):

contracts/minter-suite/Minters/MinterSEAV0.sol
jakerockland: LGTM ser – thank you for doing this.

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: updated 🙏

ryley-o:
noted this PR in the ongoing EIP-4758 discussion: https://ethereum-magicians.org/t/eip-4758-deactivate-selfdestruct/8710/39?u=ryley-o

Cc @jakerockland

PR #665, opened 2023-04-26
Provide tokenIdToHashSeed consistency in V3

jakerockland:
## Description of the change

Update CoreV3 for flagship and explorations to both support tokenIdToHashSeed and version bump to 3.1.0 / 3.1.1.

Additionally, refactor out a shared `IGenArt721CoreContractExposesHashSeed` interface to be used across all contracts that conform to this, removing it from the `_Engine`-specific core contract interface (this is safe, as no reliance on said interface outside of compile-time reliance is currently used – e.g. no usage in our subgraph, etc.).

### DIFF
--- a/contracts/explorations/GenArt721CoreV3_Explorations.sol
+++ b/contracts/explorations/GenArt721CoreV3_Explorations.sol
@@ -237,7 +237,7 @@ contract GenArt721CoreV3_Explorations is
     /// version & type of this core contract
     /// coreVersion is updated from Flagship V3 core due to minor changes
     /// implemented in the Explorations version of the contract.
-    string public constant coreVersion = "v3.1.0";
+    string public constant coreVersion = "v3.1.1";
     /// coreType remains consistent with flagship V3 core because external &
     /// public functions used for indexing are unchanged.
     string public constant coreType = "GenArt721CoreV3";
--- a/contracts/explorations/GenArt721CoreV3_Explorations.sol
+++ b/contracts/explorations/GenArt721CoreV3_Explorations.sol
@@ -237,7 +237,7 @@ contract GenArt721CoreV3_Explorations is
     /// version & type of this core contract
     /// coreVersion is updated from Flagship V3 core due to minor changes
     /// implemented in the Explorations version of the contract.
-    string public constant coreVersion = "v3.0.3";
+    string public constant coreVersion = "v3.1.0";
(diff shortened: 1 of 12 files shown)


contracts/explorations/GenArt721CoreV3_Explorations.sol
```diff
@@ -237,7 +237,7 @@ contract GenArt721CoreV3_Explorations is
     /// version & type of this core contract
     /// coreVersion is updated from Flagship V3 core due to minor changes
     /// implemented in the Explorations version of the contract.
-    string public constant coreVersion = "v3.0.3";
+    string public constant coreVersion = "v3.1.0";
```
ryley-o, line 240: nit: I think would be good to make this v3.1.1 do differentiate from flagship

ryley-o, line 240: nit: I think would be good to make this v3.1.1 do differentiate from flagship

ryley-o (approved):
In the interest of maintaining a reference to the solid understanding of our as-deployed flagship and explorations contracts, while also supporting any new deployments with these changes in our subgraph, I think the following might make sense:
- make a new `IGenArt721CoreContractV3p1` that `is IGenArt721CoreContractV3`, with the added new view function `tokenIdToHashSeed` defined in the interface

contracts/explorations/GenArt721CoreV3_Explorations.sol
jakerockland: Great call out – done!

jakerockland:
> In the interest of maintaining a reference to the solid understanding of our as-deployed flagship and explorations contracts, while also supporting any new deployments with these changes in our subgraph, I think the following might make sense:
> 
> * make a new `IGenArt721CoreContractV3p1` that `is IGenArt721CoreContractV3`, with the added new view function `tokenIdToHashSeed` defined in the interface
> 
> edit - discussed alternate solution offline that seems better

Updated to account for our discussion ser, re-requesting review 🙏

jakerockland:
@yoshiwarab moving you to CC, I don't think this PR actually _needs_ multi-review.

cc @shantanubala as well.

ryley-o (approved):
LGTM - Cc @lyaunzbe as this is a change that needs to be in the new polyptych minter 👍

PR #670, opened 2023-04-27
Add basic form of `BytecodeStorage` library versioning

jakerockland:
## Description of the change

Adds basic `BytecodeStorage` library versioning, with backwards compatible interoperability for reads.

Test coverage validating this interoperability has been added for the known cases where this interoperability should be possible (between the original library version and this new versioned-library version.

This pattern is intended to be as simple as possible (using a single 32 bytes slot for now) but should ideally be extensible over time, both for any deployments in a pre-EOF world, but also readily adaptable/flexible to adjusting for the necessary upgrades to this library down the road for EOF v1.


contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -187,20 +218,110 @@ library BytecodeStorage {
...
+    /**
+     * @notice Get version for given contract bytecode
+     * @param _address address of deployed contract with bytecode containing concat(invalid opcode, version, deployer-address, data)
+     * @return version version read from contract bytecode
+     */
+    function getLibraryVersionForBytecode(
+        address _address
+    ) internal view returns (bytes32) {
+        return _bytecodeVersionAt(_address);
+    }
```
ryley-o, line 230: any reason to not just make this public?

ryley-o (commented):
a couple thoughts to discuss before approving

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -187,20 +218,110 @@ library BytecodeStorage {
...
+            )
+            switch mload(versionString)
+            case "BytecodeStorage_V1.0.0_________ " {
+                // note: must check against literal strings, as Yul does not allow for
+                //       dynamic strings in switch statements.
+                version := V1_VERSION_STRING
+            }
+            default {
+                version := V0_VERSION_STRING
+            }
```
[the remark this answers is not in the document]
jakerockland: This is excellent point – updating to incorporate this type of handling.

For now, in the `UNKNOWN_VERSION` case, will just return empty string when reads are attempted, but in future PR will add functionality in that `UNKNOWN_VERSION` case of attempting to fall-back to `sstore2`-compatible reads.

Will resolve this comment when updates are made and this is ready for re-review.

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: This was just internal for now, as all of the readers are internal for now until we make the split to separate public external library for reads and internal library for writes – so done in accordance with that convention until we make the bigger iterative split. That seem right? @ryley-o Happy to adjust if not.

contracts/libs/0.8.x/BytecodeStorageV1.sol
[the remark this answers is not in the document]
jakerockland: Addressed in https://github.com/ArtBlocks/artblocks-contracts/pull/670/commits/ea36c6c4ce28a6ee6e4a3065e3599e881ad0420e, will handle the `sstore2` compatible reads in a separate PR, per the note above. Resolving and re-requesting review! 🙏

contracts/libs/0.8.x/BytecodeStorageV1.sol
ryley-o: that sounds good!

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -187,20 +228,117 @@ library BytecodeStorage {
...
+            extcodecopy(
+                _address,
+                versionString,
+                VERSION_OFFSET,
+                0x20 // 32-byte version string
+            )
+            // note: must check against literal strings, as Yul does not allow for
+            //       dynamic strings in switch statements.
+            switch mload(versionString)
+            case 0x2060486000396000513314601057fe5b60013614601957fe5b6000357fff0000 {
```
ryley-o, line 333: nit: might want to add a bit more context in a dev comment explaining that V0 always has these static bytes from bytecode position 1:33

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -187,20 +228,117 @@ library BytecodeStorage {
...
+                version := V0_VERSION_STRING // pre-dates actual versioning w/ version strings
+            }
+            case "BytecodeStorage_V1.0.0_________ " {
+                version := V1_VERSION_STRING
+            }
```
ryley-o, line 338: super minor nit: might want to handle latest version as first case here?

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -31,13 +38,37 @@ library BytecodeStorage {
...
+    uint256 internal constant V0_ADDRESS_OFFSET = 72;
+    uint256 internal constant V0_DATA_OFFSET = 104;
+    uint256 internal constant V1_ADDRESS_OFFSET = ADDRESS_OFFSET;
+    uint256 internal constant V1_DATA_OFFSET = DATA_OFFSET;
+
+    // Define the set of known valid version strings that may be stored in the deployed storage contract bytecode
+    // note: These are all intentionally exactly 32-bytes and are null-terminated
+    // Used for storage contracts that were deployed by an unknown source
+    bytes32 internal constant UNKNOWN_VERSION_STRING =
+        "UNKNOWN_VERSION_STRING_________ ";
```
ryley-o, line 61: nit: can you add a dev comment to give a bit more context as to why these need to be null-terminated?

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -141,26 +180,28 @@ library BytecodeStorage {
...
             // copy code to memory, excluding the deployer-address
-            extcodecopy(_address, add(data, 0x20), DATA_OFFSET, size)
+            extcodecopy(_address, add(data, 0x20), dataOffset, size)
         }
     }
 
     /**
      * @notice Get address for deployer for given contract bytecode
-     * @param _address address of deployed contract with bytecode containing concat(invalid opcode,  deployer-address, data)
+     * @param _address address of deployed contract with bytecode containing concat(invalid opcode, version, deployer-address, data)
```
ryley-o, line 189: nit: perhaps a comment here noting that V0 formatting is also accepted.

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -141,26 +180,28 @@ library BytecodeStorage {
...
+        // handle case where address contains code < addressOffset
         // note: the first check here also captures the case where
         //       (bytecodeSize == 0) implicitly, but we add the second check of
         //       (bytecodeSize == 0) as a fall-through that will never execute
-        //       unless `DATA_OFFSET` is set to 0 at some point.
-        if ((bytecodeSize < DATA_OFFSET) || (bytecodeSize == 0)) {
+        //       unless `addressOffset` is set to 0 at some point.
+        if ((bytecodeSize < addressOffset) || (bytecodeSize == 0)) {
             revert("ContractAsStorage: Read Error");
         }
```
ryley-o, line 206: I think this case is never possible?
we revert if bytecode size is zero in `_bytecodeVersionAt` (called within `_bytecodeAddressOffsetAt`), and I don't think it is possible for bytecodeSize to be less than the addressOffset unless we messed up the values of the constants in this lib.
I do think we might want to revert if `bytecodeSize < addressOffset + 32` though?

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -187,20 +228,117 @@ library BytecodeStorage {
...
+        // get the size of the data
+        uint256 bytecodeSize = _bytecodeSizeAt(_address);
+        // handle case where address contains code < VERSION_OFFSET
+        // note: the first check here also captures the case where
+        //       (bytecodeSize == 0) implicitly, but we add the second check of
+        //       (bytecodeSize == 0) as a fall-through that will never execute
+        //       unless `VERSION_OFFSET` is set to 0 at some point.
+        if ((bytecodeSize < VERSION_OFFSET) || (bytecodeSize == 0)) {
+            revert("ContractAsStorage: Read Error");
+        }
```
ryley-o, line 312: I think subcall in `_bytecodeSizeAt` reverts if `bytecodeSize == 0`, so can probably drop that here?
Also, do we want to check instead for the case `bytecodeSize < VERSION_OFFSET + 32`?

ryley-o (approved):
overall LGTM! just minor nits 🙏 Approving in the context that I'll review #681 next!

Also, I think this bumps up the engine contracts > contract size limit, but should be okay when we split these up into read/write libs

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: ++; SGTM, will update to just check `bytecodeSize < addressOffset + 32`

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Good call – re-ordering to this effect.

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Good call – expanded on this!

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Sure thing! Done.

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Good call out – updated the documentation of this to be both clearer and more maintainable.

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Updated!

yoshiwarab:
> Hrm, in that case I wonder if I should just do the splitting within this PR (or at least this PR chain), so that we don't have the repo in a broken state.

> WDYT @ryley-o @yoshiwarab ?

This sounds good to me, do you want to just continue on this then?

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -46,7 +88,7 @@ library BytecodeStorage {
     /**
      * @notice Write a string to contract bytecode
      * @param _data string to be written to contract. No input validation is performed on this parameter.
-     * @return address_ address of deployed contract with bytecode containing concat(invalid opcode, deployer-address, data)
+     * @return address_ address of deployed contract with bytecode containing
```
yoshiwarab, line 91: Should we remove the `containing` from this comment?

test/libs/BytecodeStorageV1_BackwardsCompatibleReadsMock.test.ts
```diff
@@ -0,0 +1,380 @@
...
+  assignDefaultConstants,
+} from "../util/common";
+import {
+  SQUIGGLE_SCRIPT,
+  SKULPTUUR_SCRIPT_APPROX,
+  CONTRACT_SIZE_LIMIT_SCRIPT,
+  GREATER_THAN_CONTRACT_SIZE_LIMIT_SCRIPT,
+  MULTI_BYTE_UTF_EIGHT_SCRIPT,
+} from "../util/example-scripts";
+
```
yoshiwarab, line 28: Nit: Not totally necessary but nice for DX, you can create an interface that extends T_Config and includes the expected fields that you add to the config so that you can call contract methods with typesafety/get auto complete while writing tests. See the dependency registry tests for example

yoshiwarab (approved):

ryley-o:
>Hrm, in that case I wonder if I should just do the splitting within this PR (or at least this PR chain), so that we don't have the repo in a broken state.
WDYT @ryley-o @yoshiwarab ?

>This sounds good to me, do you want to just continue on this then?

I think folding everything into this PR chain before merging to main makes sense 👍 we don't want to ever not be able to deploy engine from main 🙏

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Ah good catch – will fix and resolve this comment once I've done so.

test/libs/BytecodeStorageV1_BackwardsCompatibleReadsMock.test.ts
jakerockland: Not _necessary_, but is goated feedback. Thank you for calling this out 🙏 🔥 

I will update to do so, and resolve this comment once I've pushed an update.

jakerockland:
> > Hrm, in that case I wonder if I should just do the splitting within this PR (or at least this PR chain), so that we don't have the repo in a broken state.
> 
> > WDYT @ryley-o @yoshiwarab ?
> 
> This sounds good to me, do you want to just continue on this then?

Happy to ser if that sounds good to you 💜

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -3,13 +3,18 @@
 
 pragma solidity ^0.8.0;
 
-//---------------------------------------------------------------------------------------------------------------//
-// NOTE: This library version is still an active work-in-progress.
-//---------------------------------------------------------------------------------------------------------------//
-
 /**
  * @title Art Blocks Script Storage Library
  * @notice Utilize contract bytecode as persistant storage for large chunks of script string data.
```
lyaunzbe, line 8: nit spelling: s/persistant/persistent

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -3,13 +3,18 @@
...
+ *         This library is intended to have an external deployed copy that is released in the future,
+ *         and, as such, has been designed to support both updated V1 (versioned, with purging removed)
+ *         reads as well as backwards-compatible reads for both a) the unversioned "V0" storage contracts
+ *         which were deployed by the original version of this libary and b) contracts that were deployed
+ *         using one of the SSTORE2 implementations referenced below.
+ *         For these pre-V1 storage contracts (which themselves did not have any explict versioning semantics)
```
lyaunzbe, line 14: nit spelling: s/explict/explicit

lyaunzbe (approved):
awesome! (and thank you for the well written comments/docs 🙏 )

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Updated!

test/libs/BytecodeStorageV1_BackwardsCompatibleReadsMock.test.ts
jakerockland: Updated!

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Great catch – updated, thank you! 🙏

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Great catch – updated, thank you! 🙏

shantanubala (approved):
LGTM

PR #679, opened 2023-05-01
SEA Minter - code review final nits

ryley-o:
## Description of the change

Update SEA minter with a few minor SEA minter code review nits.

(This does not modify the minter behavior or ABIs)

### DIFF
--- a/contracts/minter-suite/Minters/MinterSEAV0.sol
+++ b/contracts/minter-suite/Minters/MinterSEAV0.sol
@@ -423,9 +423,9 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
         uint256 _auctionDurationSeconds,
         uint256 _basePrice
     ) external {
+        _onlyNonZero(_basePrice);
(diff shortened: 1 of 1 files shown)


contracts/minter-suite/Minters/MinterSEAV0.sol
```diff
@@ -888,7 +887,6 @@ contract MinterSEAV0 is ReentrancyGuard, MinterBase, IFilteredMinterSEAV0 {
         );
         // load entire auction into memory
         auction = _projectConfig.activeAuction;
-        return auction;
```
jakerockland, line 891: Nit: in this case would it be more consistent with our typical use of style to remove the named param in the return above instead and still return explicitly here?

jakerockland (approved):

contracts/minter-suite/Minters/MinterSEAV0.sol
ryley-o: yeah, good point. I've reverted this to be unchanged, thanks!

PR #680, opened 2023-05-01
Dev Test Arbitrum-Goerli Deployment

lindsgil:
## Description of the change
Adding Arbitrum specific env vars and configuration details. 
Adding deployment logs and configuration file from dev test flex contract deployment on Arbitrum Goerli


jakerockland:
@lindsgil should we be updating our shared deployer scripts / commands here with this PR as well?

lindsgil:
> @lindsgil should we be updating our shared deployer scripts / commands here with this PR as well?

@jakerockland Good call! I added a `deploy:dev-arbitrum:v3-engine` script to the package.json

package.json
```diff
@@ -22,6 +22,7 @@
     "codegen:staging": "NODE_ENV=staging yarn graphql-codegen --require dotenv/config --config ./graphql.config.js dotenv_config_path=./.env",
     "codegen:mainnet": "NODE_ENV=mainnet yarn graphql-codegen --require dotenv/config --config ./graphql.config.js dotenv_config_path=./.env",
     "deploy:dev:v3-engine": "yarn codegen:dev && NODE_ENV=dev yarn hardhat run --network goerli scripts/engine/V3/generic-engine-deployer.ts",
+    "deploy:dev-arbitrum:v3-engine": "yarn codegen:dev && NODE_ENV=dev yarn hardhat run --network arbitrum-goerli scripts/engine/V3/generic-engine-deployer.ts",
```
ryley-o, line 25: Are we going to create a new Hasura instance for arbitrum-dev? If so, I think we need:
- new env vars (e.g. `GRAPHQL_API_ENDPOINT_ARBITRUM_DEV`, `HASURA_ADMIN_SECRET_ARBITRUM_DEV`)
- update the NODE_ENV here to be `arbitrum-dev` instead of `dev`
- update `graphql.config.js` to handle the new NODE_ENV option
- update logic in `scripts/util/graphql-client-utils.ts`, `getEndpointAndAdminSecret` to handle the new NODE_ENV option

Could then retroactively update the `deployment-config.dev.ts` file here to be named `deployment-config.arbitrum-dev.ts`, and update the `environment: "arbitrum-dev",`

Currently, the deployment script that was run in this PR upserted a contract and project into our dev hasura, so we can delete those rows if the eventual plan is to stand up a new hasura instance for each new environment.

Note that if we do not plan to create a new Hasura instance for arbitrum-dev, this is WAI and no changes needed 🙏

ryley-o (commented):
this is great - want to ask one question below to clarify our indexing plan of arbitrum-goerli 🙏 🚀

package.json
lindsgil: TYVM for these call outs @ryley-o ! 🙏 There will be new Hasura instances for Arbitrum so I will update the logic accordingly and then remove those rows from our dev db 🙏

ryley-o (approved):

PR #668, opened 2023-04-27
Remove the usage of `SELFDESTRUCT` in `BytecodeStorage`

jakerockland:
## Description of the change

Purge the use of purging! 

Closes #451 

~Only merge after #665 – also, do not merge until review feels coherent between this and the TBD PR that addresses https://github.com/ArtBlocks/artblocks-contracts/issues/660~

Do not merge until this changeset looks good _in addition to_ https://github.com/ArtBlocks/artblocks-contracts/pull/670, as both should be merged at the same time.

---

Note that the approach of maintaining multiple versioned files of our library is taken in order to allow for testing backwards compatible reads, though backwards compatible reads cannot be reliably added until addressing https://github.com/ArtBlocks/artblocks-contracts/issues/660 which is being taken on in a companion PR to this one. These PRs should be reviewed in tandem.

Issue #660, opened 2023-04-25
Add basic "file versioning" into bytecode storage library


jakerockland:
Semi related to #422

jakerockland:
https://github.com/ArtBlocks/artblocks-contracts/issues/451 probably ought to be completed before this issue is.

jakerockland:
This is done.

Issue #422, opened 2023-01-06
Update BytecodeStorage Library to be EOF-compliant, while also considering Shard Blob Storage


ryley-o:
Although EOF is not currently scheduled to be in the Shanghai scope of changes, we will eventually need to focus in earnest on an updated BytecodeStorage library that is EOF-compliant being integrated into our core contracts, and potentially new versions of these core contracts being deployed across the board for any V3-contract clients.

We can begin this work from a Solidity development perspective before [EIP 3541](https://eips.ethereum.org/EIPS/eip-3540) is final, but it won't be ship-able until said EIP is final :smile:

Also related, and also not scheduled as part of Shanghai, we will have to decide if we want to use (or optionally use) [EIP 4844 ](https://eips.ethereum.org/EIPS/eip-4844)Shard Blob as contract storage space for scripts. This decision may affect if we choose to continue using BytecodeStorage at all, and it may occur at the same point in time as EOF-compliance becomes required.

Cc @jakerockland

jakerockland:
Important to note that this will not only impact our V3 contracts that are deployed that use our `BytecodeStorage` library, but will also impact our dependency registry which is expected to be shipped before EIP-3540 is active, which will then require an upgrade of said dependency registry to use the new updated EIp-3540-friendly-library.

cc @yoshiwarab

jakerockland:
> Important to note that this will not only impact our V3 contracts that are deployed that use our `BytecodeStorage` library, but will also impact our dependency registry which is expected to be shipped before EIP-3540 is active, which will then require an upgrade of said dependency registry to use the new updated EIp-3540-friendly-library.
> 
> cc @yoshiwarab

To clarify, it is this wording that suggests we should have such a transition plan in place.

> Finally, create transaction must be allowed to contain legacy initcode and deploy legacy code because otherwise there is no transition period allowing upgrading transaction signing tools. Deprecating such transactions may be considered in the future.

jakerockland:
This comment is also notable: https://ethereum-magicians.org/t/evm-object-format-eof/5727/40

ryley-o:
>This comment is also notable: https://ethereum-magicians.org/t/evm-object-format-eof/5727/40

This is a great comment that I think aligns with everyone's incentives and will likely become true in the long run. A great reason for us to be proactive on this 🙏

jakerockland:
[...]

PR #681, opened 2023-05-01
`STORE2`-compatible reads in `BytecodeStorage`

jakerockland:
## Description of the change

Staging this up for separate review, but if approved before #670 will merge directly into that branch, otherwise should be merged in very close conjunction with #670 in support of #660.

This PR updates the `BytecodeStorage` library to provide backwards-compatible reads that are compatible with `SSTORE2` as an additional read option as well as explicit-offset bytes reads, in advance of plans to split off reads into a shared external public utility library (which will live in-companion to the embedded internal library for writes).

For additional context, please see: 
- #670 
- #668 
- #660 
- #422

### DIFF
--- a/contracts/libs/0.8.x/BytecodeStorageV1.sol
+++ b/contracts/libs/0.8.x/BytecodeStorageV1.sol
@@ -8,8 +8,9 @@ pragma solidity ^0.8.0;
  * @notice Utilize contract bytecode as persistant storage for large chunks of script string data.
  *         This library is intended to have an external deployed copy that is released in the future,
  *         and, as such, has been designed to support both updated V1 (versioned, with purging removed)
- *         reads as well as backwards-compatible reads for the unversioned "V0" storage contracts
- *         which were deployed by the original version of this libary.
+ *         reads as well as backwards-compatible reads for both a) the unversioned "V0" storage contracts
+ *         which were deployed by the original version of this libary and b) contracts that were deployed
+ *         using one of the SSTORE2 implementations referenced below.
  *         For these pre-V1 storage contracts (which themselves did not have any explict versioning semantics)
  *         backwards-compatible reads are optimistic, and only expected to work for contracts actually
  *         deployed by the original version of this library – and may fail ungracefully if attempted to be
@@ -49,6 +50,7 @@ library BytecodeStorage {
     uint256 internal constant DATA_OFFSET = 65;
 
     // Define the set of known *historic* offset values for where the "meta bytes" end, and the "data bytes" begin.
+    uint256 internal constant SSTORE2_FALLBACK_DATA_OFFSET = 1;
     uint256 internal constant V0_ADDRESS_OFFSET = 72;
     uint256 internal constant V0_DATA_OFFSET = 104;
     uint256 internal constant V1_ADDRESS_OFFSET = ADDRESS_OFFSET;
@@ -153,6 +155,12 @@ library BytecodeStorage {
         uint256 bytecodeSize = _bytecodeSizeAt(_address);
         // the dataOffset for the bytecode
         uint256 dataOffset = _bytecodeDataOffsetAt(_address);
+        // validate that version is readable
+        // note: never expected to throw for reads, as we
+        //       fall back to SSTORE-2 reads optimistically
+        if (dataOffset == 0) {
+            revert("ContractAsStorage: Unsupported Version");
+        }
         // handle case where address contains code < dataOffset
         // note: the first check here also captures the case where
         //       (bytecodeSize == 0) implicitly, but we add the second check of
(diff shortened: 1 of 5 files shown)


contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -180,8 +184,27 @@ library BytecodeStorage {
...
+     * @return data string read from contract bytecode
+     */
+    function readFromBytecode(
+        address _address
+    ) internal view returns (string memory data) {
+        // the dataOffset for the bytecode
+        uint256 dataOffset = _bytecodeDataOffsetAt(_address);
+        // validate that version is readable
+        // note: never expected to throw for reads, as we
+        //       fall back to SSTORE-2 reads optimistically
```
jakerockland, line 203: Something I would love feedback in during the PR review phase for this is whether or not we should throw an `"Unsupported Version"` error for attempted reads on contracts that we don't think are SSTORE2-written (don't use 0x00 as the offset byte); this feels like more headache than it is worth, and could potentially be annoying for anyone who may have hypothetically minor-ly forked SSTORE2 to have a different "dead byte", but I can see how it would be more "semantically correct" to throw in this case, and force folks to use the manual offset reads for such scenarios... would love thoughts on this and happy to update accordingly.

ryley-o (commented):
Some initial thoughts on this - keeping it high level since this can affect a lot of this PR 🙏

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Great call out – updating to use this pattern.

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -196,6 +219,10 @@ library BytecodeStorage {
         uint256 bytecodeSize = _bytecodeSizeAt(_address);
         // the dataOffset for the bytecode
         uint256 addressOffset = _bytecodeAddressOffsetAt(_address);
+        // validate that version is readable
+        if (addressOffset == 0) {
+            revert("ContractAsStorage: Unsupported Version");
+        }
```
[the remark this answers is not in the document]
jakerockland: Good call – updating to throw `"ContractAsStorage: Unsupported Version"` in our offset calculator helpers.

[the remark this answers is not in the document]
jakerockland: Updated to use this pattern.

jakerockland:
@ryley-o I did an update pass here per our conversation, ready for your re-review ser 🙏 💜

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -303,9 +338,10 @@ library BytecodeStorage {
     ) private view returns (bytes32 version) {
         // get the size of the data
         uint256 bytecodeSize = _bytecodeSizeAt(_address);
-        // handle case where address contains code < minimum expected version string size
+        // handle case where address contains code < minimum expected version string size,
+        // by returning early with the unknown version string
         if (bytecodeSize < (VERSION_OFFSET + 32)) {
-            revert("ContractAsStorage: Read Error");
+            return UNKNOWN_VERSION_STRING;
```
ryley-o, line 344: nit: since this is a private function, and we no longer perform any actions other than reverting when we return `UNKNOWN_VERSION_STRING`, I think we could just revert here (and in the assembly block below) instead of returning the unknown version string.

Additionally, we could return an [enum](https://docs.soliditylang.org/en/latest/types.html#enums) (e.g. `enum SupportedBytecodeStorageVersions { V0, V1 }`) instead of the version strings, which would help bound the possible return values and clean up some of the checks in `_bytecodeAddressOffsetAt` and `_bytecodeDataOffsetAt`.

If you think the current implementation is easier to grok, I would be fine leaving as-is (no super strong opinions from me here)

ryley-o (approved):
This is awesome 😄 !!!

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: This function is private, but `getLibraryVersionForBytecode` is internal (and will be public in the new split library) so I think we do want to keep a return value for the "unknown" state.

Given that I anticipate we may have minor versions over time, I am generally inclined to keep the versioning to be a `bytes32` string for now (note: us bounding this to a `bytes32` slot rather than allowing it to be an unbounded `string` is an important consideration in my rationale for keeping it this way, I'd feel differently about using `string) – and consider adding an "enum wrapper" down the road if we think that would be clearer to manage. 

I don't feel strongly about this though, so happy to adjust in an iteration if you feel more strongly than initially surfaced here or if we get feedback from prospective library clients who would like us to expose this functionality.

For now, going to resolve this and merge this PR down into #670, but can do additional iteration in future PRs as needed.

ryley-o: Got it and everything you notes makes sense to me! Thanks for double clicking here 🙏

jakerockland: > Got it and everything you notes makes sense to me! Thanks for double clicking here 🙏

Vibey – TY ser 💜

PR #684, opened 2023-05-03
Split `BytecodeStorage` into public/internal libraries

### LINKED ISSUE #660: Add basic "file versioning" into bytecode storage library

jakerockland:
## Description of the change

This PR builds on top of #668 and #670 to actualize the architecture discussed in https://github.com/ArtBlocks/artblocks-contracts/issues/422#issuecomment-1523789750 of a split public shared external reader library and an internal writer library.

This PR is intended to be atomically merged into `main` at the same time as #668 and #670 and is thus marked as "do not merge" until all three have been through review and are ready to merge. Because of this, we are *not* bumping the versions of the client CoreV3 contracts using this library with this changeset, as they are already version-bumped in #668.

~Note: that this PR is still marked as a WIP as we have not yet made the necessary changes to the deployment script of this PR, as that will be a "phase 2" update to this changeset after I have gotten an initial LGTM from reviewers on the overall architecture and code structure here.~

Our shared deployer script has been added, and deployment using this library linking approach has been validated end-to-end on goerli, with shared deployments of the library being deployed on both goerli and mainnet. 

Pending the final approval of this PR, I will merge this as well as #668 and #670 into main, and we can consider the V1 version of this new `BytecodeStorage` library as released!! (closes #660).

### DIFF
--- a/contracts/DependencyRegistryV0.sol
+++ b/contracts/DependencyRegistryV0.sol
@@ -831,7 +831,7 @@ contract DependencyRegistryV0 is
 
     /**
      * Helper for calling `BytecodeStorageReaderV1` external library reader method,
-     * added for gas-optimization purposes.
+     * added for bytecode size reduction purposes.
      */
     function _readFromBytecode(
         address _address
--- a/contracts/DependencyRegistryV0.sol
+++ b/contracts/DependencyRegistryV0.sol
@@ -34,7 +34,6 @@ contract DependencyRegistryV0 is
     IDependencyRegistryV0
 {
     using BytecodeStorageWriterV1 for string;
-    using BytecodeStorageWriterV1 for address;
     using Bytes32Strings for bytes32;
     using Strings for uint256;
     using EnumerableSet for EnumerableSet.Bytes32Set;
--- a/contracts/DependencyRegistryV0.sol
+++ b/contracts/DependencyRegistryV0.sol
@@ -34,9 +34,7 @@ contract DependencyRegistryV0 is
     IDependencyRegistryV0
 {
     using BytecodeStorageWriterV1 for string;
-    using BytecodeStorageReaderV1 for string;
     using BytecodeStorageWriterV1 for address;
-    using BytecodeStorageReaderV1 for address;
     using Bytes32Strings for bytes32;
     using Strings for uint256;
     using EnumerableSet for EnumerableSet.Bytes32Set;
@@ -737,7 +735,7 @@ contract DependencyRegistryV0 is
             return "";
         }
 
-        return dependency.scriptBytecodeAddresses[_index].readFromBytecode();
+        return _readFromBytecode(dependency.scriptBytecodeAddresses[_index]);
     }
 
     /**
@@ -830,5 +828,15 @@ contract DependencyRegistryV0 is
     function _transferOwnership(address newOwner) internal override {
         OwnableUpgradeable._transferOwnership(newOwner);
         adminACLContract = IAdminACLV0(newOwner);
+    }
+
+    /**
+     * Helper for calling `BytecodeStorageReaderV1` external library reader method,
+     * added for gas-optimization purposes.
+     */
+    function _readFromBytecode(
+        address _address
+    ) internal view returns (string memory) {
+        return BytecodeStorageReaderV1.readFromBytecode(_address);
     }
 }
--- a/contracts/DependencyRegistryV0.sol
+++ b/contracts/DependencyRegistryV0.sol
@@ -33,7 +33,7 @@ contract DependencyRegistryV0 is
     OwnableUpgradeable,
     IDependencyRegistryV0
 {
-    using BytecodeStorageWriterV1 for string;
+    using BytecodeStorageWriter for string;
     using Bytes32Strings for bytes32;
     using Strings for uint256;
     using EnumerableSet for EnumerableSet.Bytes32Set;
@@ -830,12 +830,12 @@ contract DependencyRegistryV0 is
     }
 
     /**
-     * Helper for calling `BytecodeStorageReaderV1` external library reader method,
+     * Helper for calling `BytecodeStorageReader` external library reader method,
      * added for bytecode size reduction purposes.
      */
     function _readFromBytecode(
         address _address
     ) internal view returns (string memory) {
-        return BytecodeStorageReaderV1.readFromBytecode(_address);
+        return BytecodeStorageReader.readFromBytecode(_address);
     }
 }
--- a/contracts/DependencyRegistryV0.sol
+++ b/contracts/DependencyRegistryV0.sol
@@ -33,8 +33,10 @@ contract DependencyRegistryV0 is
     OwnableUpgradeable,
     IDependencyRegistryV0
 {
-    using BytecodeStorage for string;
-    using BytecodeStorage for address;
+    using BytecodeStorageWriterV1 for string;
+    using BytecodeStorageReaderV1 for string;
+    using BytecodeStorageWriterV1 for address;
+    using BytecodeStorageReaderV1 for address;
     using Bytes32Strings for bytes32;
     using Strings for uint256;
     using EnumerableSet for EnumerableSet.Bytes32Set;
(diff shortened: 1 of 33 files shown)


contracts/DependencyRegistryV0.sol
```diff
@@ -829,4 +829,14 @@ contract DependencyRegistryV0 is
         OwnableUpgradeable._transferOwnership(newOwner);
         adminACLContract = IAdminACLV0(newOwner);
     }
+
+    /**
+     * Helper for calling `BytecodeStorageReaderV1` external library reader method,
+     * added for gas-optimization purposes.
```
ryley-o, line 835: nit: bytecode size reduction, not gas (this comment applies to all contracts in this PR)

contracts/DependencyRegistryV0.sol
```diff
@@ -33,8 +33,8 @@ contract DependencyRegistryV0 is
     OwnableUpgradeable,
     IDependencyRegistryV0
 {
-    using BytecodeStorage for string;
-    using BytecodeStorage for address;
+    using BytecodeStorageWriterV1 for string;
+    using BytecodeStorageWriterV1 for address;
```
ryley-o, line 37: we don't actually need to attach this to address anymore 🙂 (this comment applies to all contracts in this PR)

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -33,7 +37,14 @@ pragma solidity ^0.8.0;
...
  */
-library BytecodeStorage {
+
+/**
+ * @title Art Blocks Script Storage Library (Public, Reads)
+ * @author Art Blocks Inc.
+ * @notice The public library for reading from storage contracts. This library is intended to be deployed as a
+ *         standalone contract, and provides all _read_ functionality.
+ */
+library BytecodeStorageReaderV1 {
```
ryley-o, line 47: nit: I think generally better to keep the "V1" in the file name (specified by importing files), but drop the "V1" off of the library name itself. This simplifies the updates required when we move to a new version to only be import updates.
Same goes for the Writer library in this file

contracts/libs/0.8.x/BytecodeStorageV1.sol
```diff
@@ -45,112 +56,71 @@ library BytecodeStorage {
...
+    uint256 private constant V1_ADDRESS_OFFSET = ADDRESS_OFFSET;
+    uint256 private constant V1_DATA_OFFSET = DATA_OFFSET;
 
     // Define the set of known valid version strings that may be stored in the deployed storage contract bytecode
     // note: These are all intentionally exactly 32-bytes and are null-terminated. Null-termination is used due
     //       to this being the standard expected formatting in common web3 tooling such as ethers.js. Please see
     //       the following for additional context: https://docs.ethers.org/v5/api/utils/strings/#Bytes32String
     // Used for storage contracts that were deployed by an unknown source
-    bytes32 internal constant UNKNOWN_VERSION_STRING =
+    bytes32 private constant UNKNOWN_VERSION_STRING =
```
ryley-o, line 82: nit: I think we could break out the version strings to be defined outside of the libraries, which makes the shared use of CURRENT_VERSION a bit easier to read?
Check out the diff here (although please ignore the private/internal in the diff 😓 ) https://www.diffchecker.com/oNDpqBSG/

ryley-o (approved):
LGTM - really like this architecture 🙏 

a few comments with some minor nits is all 🥳

contracts/DependencyRegistryV0.sol
jakerockland: Great catch – updated, here and throughout 🙏

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: @ryley-o I'm torn on doing this, vs just making these constants public. I think my inclination is the later, because then it namespaces the constants. I've updated this PR to this effect. WDYT?

contracts/libs/0.8.x/BytecodeStorageV1.sol
jakerockland: Good call – updated here and throughout.

contracts/libs/0.8.x/BytecodeStorageV1.sol
ryley-o: Oh, I like how you organized this much better! LGTM!

scripts/util/constants.ts
```diff
@@ -9,6 +9,13 @@ export const DELEGATION_REGISTRY_ADDRESSES = {
   mainnet: "0x00000000000076A84feF008CDAbe6409d2FE638B",
 };
 
+// BytecodeStorageReader library addresses on supported networks
+export const BYTECODE_STORAGE_READER_LIBRARY_ADDRESSES = {
+  // note: same address for goerli and mainnet
```
ryley-o, line 14: different addresses?

scripts/engine/V3/generic-engine-deployer.ts
```diff
@@ -207,13 +208,21 @@ async function main() {
     await randomizer.deployed();
     const randomizerAddress = randomizer.address;
     console.log(
-      `[INFO] Randomizer ${deployDetails.randomizerContractName} deployed at ${randomizerAddress}`
+      `[INFO] Randomizer ${deployDetails.randomizerContractName} deployed at g${randomizerAddress}`
```
ryley-o, line 211: super minor nit - is that g intentional?

scripts/util/constants.ts
ryley-o, line 14: should this say different addresses? this is a case where a create3 pattern might be ideal to get the same address across many different chains, but I don't want to spend time over-optimizing, so totally up to you.

scripts/engine/V3/generic-engine-deployer.ts
```diff
@@ -207,13 +208,21 @@ async function main() {
...
     );
     await delay(EXTRA_DELAY_BETWEEN_TX);
 
     // Deploy Core contract
+    // Ensure that BytecodeStorageReader library is linked in the process
+    const bytecodeStorageLibraryAddress =
+      BYTECODE_STORAGE_READER_LIBRARY_ADDRESSES[networkName];
```
ryley-o, line 218: can you add a check in the validation section to catch if this is undefined before we perform any deployment actions? (e.g. immediately fail if we were to run this today on arbitrum-goerli)

deployments/engine/V3/internal-testing/staging-v3-react-demo-2/DEPLOYMENTS.md
```diff
@@ -0,0 +1,42 @@
...
+
+**Metadata**
+
+- **Starting Project Id:** 0
+- **Token Name:** V3 Engine Demo External Reader Lib
+- **Token Ticker:** DEMO
+- **Auto Approve Artist Split Proposals:** true
+- **Render Provider Address, Primary Sales:** deployer
+- **Platform Provider Address, Primary Sales:** deployer
+
```
ryley-o, line 34: I think it would be good to update the script to add a line like:
```
- **BytecodeStorageReader:** <bytecodeStorageLibraryAddress>
```

ryley-o (approved):
🚀 very awesome!

scripts/engine/V3/generic-engine-deployer.ts
jakerockland: Ugh, nah thank you for catching! Fixed.

scripts/util/constants.ts
jakerockland: AH, good catch. We should indeed. I don't think its worth the headache of doing the create3 pattern here given that this is a stateless read-only library, but if we got that feedback from clients could use create3 pattern for a future version deployment. Updated the documentation! 🙏

scripts/engine/V3/generic-engine-deployer.ts
jakerockland: Great call – done!

deployments/engine/V3/internal-testing/staging-v3-react-demo-2/DEPLOYMENTS.md
jakerockland: Great call – added to the engine deployment script.

PR #697, opened 2023-05-08
Restructure as monorepo, adding shell for sdk package

shantanubala:
## Description of the change

This PR is a stub containing an initial implementation of a frontend data library that assists with configuring and using minters using the latest minter suite rearchitecture.


yoshiwarab:
@shantanubala @jakerockland I think if we're adding this package into this repo we should probably update the repo to be structured as a proper monorepo (i.e. with a top-level package.json that declares workspaces that can each have their own independent dependencies). If we don't want to do this work now, my suggestion would be to leave this in its own repo.

shantanubala:
> @shantanubala @jakerockland I think if we're adding this package into this repo we should probably update the repo to be structured as a proper monorepo (i.e. with a top-level package.json that only declares workspaces and shared dev dependencies with packages that can each have their own independent dependencies). If we don't want to do this work now, my suggestion would be to leave this in its own repo.


I've used workspaces in the past with Yarn and they work great, but I think the repository structure can be decoupled from the development work.

I'm a bit hesitant to shuffle around the directory structure of the repo while there's a lot of ongoing work with PR's in progress on the actual smart contract code. I want to avoid merge conflicts right now, but towards the end of Q2 when we feel like we can freeze the contract code for a couple days, I can pretty quickly move the contracts into a subfolder + create a workspace package.json that unifies the frontend library and contract code.

This shouldn't block us from publishing the frontend library on NPM - so we should be able to ship this library before we have a fully polished repository structure.

yoshiwarab:
@shantanubala, after speaking with @jakerockland we decided it's probably best to just get the mono repo structure in place early. Went ahead and update this PR to do that. I'm going to remove the do not merge label and change it from draft state so we can get it in as soon as possible and avoid merge conflicts.

jakerockland (approved):

.gitignore
```diff
@@ -1,56 +1,3 @@
...
-# Coverage directory used by tools like istanbul
-coverage
-coverage.json
-coverageEnv
-/contracts-flat/
-
-/cache/
-/artifacts/
-
-.env
```
ryley-o, line 39: WDYT about keeping `.env` in here for safety of us and anyone who has previously forked this repo?

package.json
```diff
@@ -1,105 +1,21 @@
 {
```
ryley-o, line 1: We should add a new root-level readme that summarizes our docs.

ryley-o (approved):
Overall looks great 🙏 

Seems like everything should ~just work here for the most part.

In addition to the couple very minor comments:
- please scan the `artblocks-docs` repo for anything linking directly to our contracts repo, and adjust those links accordingly
- a cursory scan of some of the markdown files in the contracts package reveals a few broken links. do you mind updating those as part of this?

ryley-o:
Also want to note that I verified no changes to hardhat's created `artifacts` directory via diff of the directory created on `main` and directory created on `sb/create-frontend-library` 👍

.gitignore
shantanubala: Looks like @yoshiwarab has implemented this, resolving 🙏

package.json
shantanubala: Looks like @yoshiwarab has implemented this, resolving 🙏

PR #704, opened 2023-05-10
Demonstrate pattern for `CALL`ing linked library rather than linked calls via `DELEGATECALL`

jakerockland:
## Description of the change

Add mocks with additional tests to validate harnessing of external library as a `CALL`-able (vs. `DELEGATECALL`-able) public contract.

This is the pattern that we plan to use in the future when we eventually release a `BytecodeStorageV2` variant of this read library, that proxy-reads into the `BytecodeStorageV1` variant of the library for backwards compatible reads into legacy (pre-EOF) storage contracts.

image


ryley-o (approved):
LGTM 🙏 Thanks for verifying this behavior!

jakerockland:
> LGTM 🙏 Thanks for verifying this behavior!

TYVM for the quick review ser! 🙏

PR #706, opened 2023-05-10
Bump Solidity compiler version

jakerockland:
## Description of the change

Update Solidity compiler version from `^0.8.17` to `^0.8.19`.

Note, not updating to very latest `0.8.20` currently as this is causing test breakages for an unclear reason, going to let that version cook a bit more.

### DIFF
--- a/.openzeppelin/goerli.json
+++ b/.openzeppelin/goerli.json
@@ -16,7 +16,7 @@
       "address": "0x7F6767bB1aB8841DDc04317ac60291f799f268A6",
       "txHash": "0x0ba7e432066118ce2fdc639f3ab824ce84696838278ba29164e778c1a76ad951",
       "layout": {
-        "solcVersion": "0.8.17",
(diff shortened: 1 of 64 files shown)


ryley-o (approved):
LGTM - thanks for doing this.

I do want to note that Hardhat does not fully support Solidity 0.8.19 yet, and some features such as stack traces might not work correctly. I think this is okay though.

Solidity changelogs:
- https://blog.soliditylang.org/2023/02/01/solidity-0.8.18-release-announcement/
- https://blog.soliditylang.org/2023/02/22/solidity-0.8.19-release-announcement/

Just to note, I don't think these updates affect our current codebase in any meaningful way, and I don't see any immediate areas where we should be immediately using the new [user defined value types](https://docs.soliditylang.org/en/latest/types.html#user-defined-value-types) or [operators](https://blog.soliditylang.org/2023/02/22/user-defined-operators/), as they would probably make our code more difficult to audit and read in the near term.

PR #726, opened 2023-05-22
Add utility functions and tests for handling delegation/allowlists, include typechain

shantanubala:
## Description of the change

I'm starting to implement utility functions for handling purchase/purchaseTo calls to our minters - this PR implements utilities for delegation and merkle proofs - as a followup in a separate PR, I will implement the actual purchasing logic.

This PR includes the following changes:

1. Implements utility functions for determining if the currently-connected user wallet is a delegate for a single vault and/or generating merkle proofs
2. Adds typechain type generation (`yarn run build:typechain`) as a runnable script to for type-safe contract calls
3. Adds the necessary dependencies and tests for handling delegates / merkle proofs

The code is heavily based on our existing implementation, and should provide a drop-in replacement for the existing functionality on the Art Blocks web site.

## Testing

This PR includes automated tests, but these tests mock out the actual calls to the delegation registry - if there is a more elegant way to end-to-end test this portion of the library, it would be good to explore as a followup, but that seemed like it wouldn't provide too much benefit.

---
  - https://app.asana.com/0/0/1204650215229016


packages/sdk/package.json
```diff
@@ -27,7 +30,8 @@
     "test": "jest --collect-coverage",
     "build:docs": "jsdoc -c jsdoc.json",
     "pre-commit": "yarn lint",
-    "pre-push": "yarn test"
+    "pre-push": "yarn test",
+    "build:typechain": "typechain --target ethers-v5 --out-dir ./src/contracts './artifacts/*.json' './node_modules/@artblocks/contracts/artifacts/**/*.json'"
```
shantanubala, line 34: This is roughly the same approach taken by the Art Blocks web site for generating types - @yoshiwarab does this look good to you?

packages/sdk/package.json
```diff
@@ -49,5 +53,9 @@
   },
   "peerDependencies": {
     "ethers": "^6.3.0"
+  },
+  "dependencies": {
+    "@artblocks/contracts": "^1.0.2",
```
shantanubala, line 58: It seems like we want our SDK to depend on the latest _published_ NPM package, and it should be a hard dependency instead of a peer dependency. Does that sound right?

packages/sdk/artifacts/IDelegationRegistry.json
```diff
@@ -0,0 +1,501 @@
+[
```
shantanubala, line 1: Since we need to call `getDelegationsByDelegate` from the SDK, we also need to have typechain generate types for this interface.

packages/sdk/src/contracts/IDelegationRegistry.ts
```diff
@@ -0,0 +1,867 @@
+/* Autogenerated file. Do not edit manually. */
```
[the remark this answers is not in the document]
ryley-o: I have a bit of a preference to keep all autogenerated files out of our repo, but happy to defer to your intuition here since you are most familiar with the code in this package.

packages/sdk/src/purchase/delegation.ts
```diff
@@ -0,0 +1,69 @@
...
+  signerOrProvider: Signer | providers.Provider
+): IDelegationRegistry => {
+  return IDelegationRegistry__factory.connect(
+    contractAddress,
+    signerOrProvider
+  );
+};
+
+/**
+ * @summary Get the vault that the user has delegated to.
```
ryley-o, line 48: nit: I think some of the wording/naming is a little odd here - I would phrase this as: "Get the vault that the user is a delegate for". This is because the vault was the one that had to perform the delegation - the user/hot wallet is simply the delegate. 

for example, I would start description below with something like: "This function is used to determine if the user is the delegate of a vault."

ryley-o (approved):
LGTM!
Appreciate the smol PR size here 🙏 
Not sure if you are aware, but there is a delegate.cash SDK available, but not sure if it would be useful in this case: https://docs.delegate.cash/delegatecash/technical-documentation/javascript-sdk

packages/sdk/src/contracts/IDelegationRegistry.ts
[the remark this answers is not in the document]
yoshiwarab: ++ Would prefer to generate at build time so we always know it's in sync with the relevant abis

packages/sdk/src/purchase/delegation.ts
```diff
@@ -0,0 +1,69 @@
...
+export const DELEGATION_REGISTRY = "0x00000000000076a84fef008cdabe6409d2fe638b";
+
+/**
+ * @summary Error thrown when a user is a delegate on behalf of multiple vaults.
+ * @description Currently, only a single vault can be delegated to at a time.
+ * This error is thrown when a user has delegated to multiple vaults, and provides
+ * an easy way for SDK consumers to catch this error and display a helpful message
+ * to the user.
+ */
+export class MultipleDelegationsError extends Error {
```
yoshiwarab, line 19: My inclination would be to update `getDelegateVault` to be `getDelegateVaults` and allow the SDK consumer to decide how they want to handle the case where there's multiple vaults. We could set up purchase so that it takes an optional vault that's provided by the caller and then we don't need to care about multiple vaults in the SDK itself.

For context, I think we just didn't want to figure out the UX for users with multiple vaults but since we're not dealing with the UX here directly I think it's fine to leave this more flexible

yoshiwarab (approved):

packages/sdk/src/contracts/IDelegationRegistry.ts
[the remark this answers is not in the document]
shantanubala: Sounds good, removing 👍

packages/sdk/package.json
shantanubala: Done!

packages/sdk/src/purchase/allowlist.test.ts
```diff
@@ -0,0 +1,30 @@
...
+describe("generateUserMerkleProof", () => {
+  it("should generate a correct merkle proof for a user on the allowlist", () => {
+    expect(generateUserMerkleProof(ALLOWLIST, ALLOWLIST[0])).toEqual([
+      "0x63d47d072f0f348e428f18ebb7cdbba5df86105bab84576f0eea618c16eca7d6",
+      "0xaf489f4dffceae68780af004591d929c5ac342c4213558632bfb56615986f8e1",
+      "0xf662b2098c85aeb86eeda04d033d18ffef266ea445807791f8ca2c656369d95d",
+    ]);
+  });
+
+  it("should generate a empty proof for a user not on the allowlist", () => {
```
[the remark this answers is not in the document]
shantanubala: Done 👍

PR #730, opened 2023-05-23
fix deployment file locations

ryley-o:
## Description of the change

I think these got caught up during the recent transition to monorepo 🙏


jakerockland (approved):
🤦‍♂️ 🙏 🙏

PR #736, opened 2023-05-23
Remove husky pre-push test

ryley-o:
## Description of the change

Remove husky pre-push test run in contracts package to increase development speed, and instead rely on our CI to enforce the passing of tests.


packages/contracts/package.json
```diff
@@ -32,7 +32,7 @@
     "deploy:goerli:minter": "yarn hardhat run --network goerli scripts/minter-deployments/V3/generic-minter-deployer-v3core.ts",
     "deploy:mainnet:minter": "yarn hardhat run --network mainnet scripts/minter-deployments/V3/generic-minter-deployer-v3core.ts",
     "pre-commit": "yarn lint",
-    "pre-push": "yarn test:parallel"
+    "pre-push": "echo no pre-push action in contracts package"
```
ryley-o, line 35: I think this is the right way to do this? because we have a pre-push in one of the other packages as well, so keeping the script, but doing nothing

jakerockland (approved):

PR #760, opened 2023-06-01
update config template and generic deployer script

ryley-o:
## Description of the change

Update deployer script to allow configuring of non-standard render provider split percentages.

### DIFF
--- a/packages/contracts/scripts/engine/V3/generic-engine-deployer.ts
+++ b/packages/contracts/scripts/engine/V3/generic-engine-deployer.ts
@@ -407,6 +407,70 @@ async function main() {
       console.log(`[INFO] Skipping adding placeholder initial token.`);
     }
 
+    // update split percentages if something other than the default is provided
+    // primary sales
+    if (deployDetails.renderProviderSplitPercentagePrimary) {
+      // a render provider split percentage override was provided
+      if (
+        deployDetails.renderProviderSplitPercentagePrimary < 0 ||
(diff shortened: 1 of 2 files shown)


packages/contracts/scripts/engine/V3/generic-engine-deployer.ts
```diff
@@ -407,6 +407,76 @@ async function main() {
       console.log(`[INFO] Skipping adding placeholder initial token.`);
     }
 
+    // update split percentages if something other than the default is provided
+    // primary sales
+    if (deployDetails.renderProviderSplitPercentagePrimary) {
+      // a render provider split percentage override was provided
+      if (
+        deployDetails.renderProviderSplitPercentagePrimary < 0 ||
+        deployDetails.renderProviderSplitPercentagePrimary > 90
```
jakerockland, line 416: I wonder if we should read this from the contract rather than hard-coding?

jakerockland (approved):

packages/contracts/scripts/engine/V3/generic-engine-deployer.ts
ryley-o: good call, I've updated in [ed9ff12](https://github.com/ArtBlocks/artblocks-contracts/pull/760/commits/ed9ff124a3e399d8664709e85917f92f30a1ca7f).
Overall, this is a bit safer since we have to specify the current platform provider split when updating the render provider split, so extra happy with this idea.

PR #761, opened 2023-06-01
Upgrade typechain, typechain/ethers v5, make compatible

ryley-o:
## Description of the change

Perform upgrades recommended by dependabot in #741 & #757, and update directory names to be compatible with the new dependency versions.

Specifically, we had four directories that started with numbers that are incompatible with the new typechain dependency version. This is a documented issue in typechain here: https://github.com/dethcrypto/TypeChain/issues/716

For a more immediate workaround, the four directories are renamed to begin with a letter in this PR.

Note that this may or may not have downstream impacts to consumers of the @artblocks/contracts npm package, although it seems unlikely that consumers would be hard-coding the directory paths vs. abstracting those types of operations via hardhat libraries.


jakerockland (approved):
@ryley-o thank you very much for the update here – the `v{version}` prefixed format is probably more idiomatic anyway, so this is a net positive for multiple reasons! 🙏 🙏

PR #769, opened 2023-06-03
GenArt721CoreV4 with hook to artist contract

campionfellin:
## Description of the change

New core contract. Main changes are that instead of "minting" to this contract, the mint is delegated to the artist's contract. Token URI is still managed in AB, and no other major functionality changes. Will describe detailed changes below.

## Smaller Changes

The main things to look out for are that `addProject` now takes in `address _artistContractAddress`, which is the contract that the token ultimately mints to.

```solidity
function addProject(
        string memory _projectName,
        address payable _artistAddress,
        address _artistContractAddress
) external {
```

In `mint_Ecf` we now mint to the creator's contract -

```
  // Mint to the artist's contract.
  uint256 thisTokenId = IManifoldCore(project.artistContractAddress)
      .mintExtension(_to);
```

instead of this -

```
_mint(_to, thisTokenId);
```

The `tokenURI` function interface is now - 

```
function tokenURI(
        address _creatorCore,
        uint256 _tokenId
    ) public view override returns (string memory) {
```

Which goes with the `ICreatorTokenExtensionURI` interface from Manifold.


jakerockland:
Howdy ser! I would love to get more context on this desired usage of this changeset before reviewing in depth. Also, I would probably suggest that we name this something like `GenArt721CoreV3_Manifold` (if designed to be used in companion with Manifold artist contracts) or `GenArt721CoreV3_Auxillary` (if something more generically usable) rather than doing a major version bump to V4.

jakerockland:
Upon a quick look here, I think this approach that breaks the relationship of tokenID to projectID mapping is something that would not be immediately compatible with infra assumptions on our side, and so is not a readily supportable change.

@campionfellin @tlip it would be helpful I think to get a better understanding from a higher level of abstraction around what you're trying to accomplish here for Prohibition so that we can talk through what options may make the most sense. 

Also, it probably makes sense to fork this off of the Prohibition fork contract if this is planning to be used there @campionfellin @tlip

campionfellin:
Will write some more description here now, sorry for leaving it in this state.

packages/contracts/contracts/GenArt721CoreV4.sol
```diff
@@ -0,0 +1,2010 @@
...
+        emit ProjectUpdated(_projectId, FIELD_PROJECT_PAUSED);
+    }
+
+    /**
+     * @notice Adds new project `_projectName` by `_artistAddress`.
+     * @param _projectName Project name.
+     * @param _artistAddress Artist's address.
+     * @dev token price now stored on minter
+     */
+    function addProject(
```
campionfellin, line 861: Updated here, with new `_artistContractAddress` parameter.

packages/contracts/contracts/GenArt721CoreV4.sol
```diff
@@ -0,0 +1,2010 @@
...
+        projects[_projectId].name = _projectName;
+        emit ProjectUpdated(_projectId, FIELD_PROJECT_NAME);
+    }
+
+    /**
+     * @notice Updates artist contract address of project `_projectId` to be `__artistContractAddress`.
+     * @param _projectId Project ID.
+     * @param _artistContractAddress New project artist contract address.
+     */
+    function updateProjectArtistContractAddress(
```
campionfellin, line 912: New function to allow updating the artist's contract.

packages/contracts/contracts/GenArt721CoreV4.sol
```diff
@@ -0,0 +1,2010 @@
...
+
+    /**
+     * @dev See {IERC165-supportsInterface}.
+     */
+    function supportsInterface(
+        bytes4 interfaceId
+    ) public view virtual override returns (bool) {
+        return
+            interfaceId == type(IManifold).interfaceId ||
+            interfaceId == type(IManifoldCreatorExtensionURI).interfaceId ||
```
campionfellin, line 1883: It now implements the `IManifoldCreatorExtensionURI` which tells the artist's contract that it's OK to fetch `tokenURI` info from here.

packages/contracts/contracts/GenArt721CoreV4.sol
```diff
@@ -0,0 +1,2010 @@
...
+        string artist;
+        address descriptionAddress;
+        string website;
+        string license;
+        string projectBaseURI;
+        bytes32 scriptTypeAndVersion;
+        string aspectRatio;
+        // mapping from script index to address storing script in bytecode
+        mapping(uint256 => address) scriptBytecodeAddresses;
+        address artistContractAddress;
```
campionfellin, line 199: New field as part of project. This is the contract that the token is ultimately minted to.

packages/contracts/contracts/GenArt721CoreV4.sol
```diff
@@ -0,0 +1,2010 @@
...
+        project.invocations = invocationsAfter;
+
+        // mark project as completed if hit max invocations
+        if (invocationsAfter == maxInvocations) {
+            _completeProject(_projectId);
+        }
+
+        // Mint to the artist's contract.
+        uint256 thisTokenId = IManifoldCore(project.artistContractAddress)
+            .mintExtension(_to);
```
campionfellin, line 424: This is where the token actually gets minted now (instead of to the GenArt contract)

packages/contracts/contracts/GenArt721CoreV4.sol
```diff
@@ -0,0 +1,2010 @@
...
+
+    /**
+     * @notice Gets token URI for token ID `_tokenId`.
+     * @param _creatorCore Address of creatorCore contract.
+     * @param _tokenId Token ID to be queried.
+     * @return string URI of token ID `_tokenId`.
+     * @dev token URIs are the concatenation of the project base URI and the
+     * token ID.
+     */
+    function tokenURI(
```
campionfellin, line 1860: New functionality, takes in `_creatorCore` contract to fulfill the `ICreatorTokenExtensionURI` interface.

packages/contracts/contracts/GenArt721CoreV4.sol
```diff
@@ -0,0 +1,2010 @@
...
+
+    /// version & type of this core contract
+    string public constant coreVersion = "v4.0.0";
+    string public constant coreType = "GenArt721CoreV4";
+
+    /// default base URI to initialize all new project projectBaseURI values to
+    string public defaultBaseURI;
+
+    // Maps tokenIds to projectIds because cannot just divide by 1M anymore
+    mapping(uint => uint) public tokenIdToProjectIds;
```
campionfellin, line 254: This mapping is to ensure we keep track of tokens in the case of non-serial minting. Here is the scenario - 

1. Create drop on AB, say for 100 tokens.
2. People mint 3 tokens on AB.
3. Creator mints a token normally on the same contract
4. People mint 2 more tokens on AB.

Then this AB contract only manages tokens 1-3 and 4-5. This is why we must keep track.

packages/contracts/contracts/GenArt721CoreV4.sol
```diff
@@ -0,0 +1,2010 @@
...
+
+    /**
+     * @notice Gets the project ID for a given `_tokenId`.
+     * @param _tokenId Token ID to be queried.
+     * @return _projectId Project ID for given `_tokenId`.
+     */
+    function tokenIdToProjectId(
+        uint256 _tokenId
+    ) public view returns (uint256 _projectId) {
+        return tokenIdToProjectIds[_tokenId];
```
campionfellin, line 1804: No longer divided by ONE_MILLION because on Manifold tokens are minted incrementally. See above comment for why we save in this mapping

packages/contracts/contracts/GenArt721CoreV4.sol
tlip: This wouldn't work because if you have multiple `artistContracts` giving a `tokenId = 1`, then they'll keep overwriting each other.

tlip:
> I can certainly fork it off of there, will likely need to update the engine contract instead of the core minting contract.

The Engine fork for Prohibition _is_ the core minting contract for Prohibition.

> Does your rendering engine assume everything is based on the million-based format? i.e. tokenID `418000000` for token 0 on project 418?
>
> What would the work be required to read tokenIds from the contract?

You're right about all `tokenId`s for a given project are prefixed with `projectId * 1000000`. I can't make any assumptions about how AB handles these things in the generator but I'm going to assume that nonlinear tokenIds (and possibly even conflicting tokenIds across different projects) isn't something they're going to put work into supporting. On that note, Prohibition also assumes `tokenId`s match that exact convention for each project.

I also don't know where the generator is reading mint events from so this could potentially break that as well.

There are a few more reasons that I haven't listed here but I don't think Prohibition would be able to support this.

Is the goal here just to have artists own the contract afterwards? I can think of methods of doing so from a "within the system" way and less of a BYOC way

packages/contracts/contracts/GenArt721CoreV4.sol
campionfellin: Great catch, because I am saving it below -

```
tokenIdToProjectIds[thisTokenId] = _projectId;
```

Will really need to have that be like this instead - 

```
tokenIdToProjectIds[project.artistContractAddress][thisTokenId] = _projectId;
```

Will update!

campionfellin:
Closing, can circle back one day.

PR #817, opened 2023-06-22
Revert "Bump hardhat from 2.15.0 to 2.16.0"

ryley-o:
Reverts ArtBlocks/artblocks-contracts#813.

It appears that the latest solidity-coverage is not yet compatible with the latest version of hardhat. I think it is preferable to roll back this change and wait on a fix, and revisit this in a polish week, as upgrading to the latest release of `solidity-coverage` does not fix the incompatibility (and the latest hardhat version came out 16 hours ago).

If this is merged, we should follow with an Asana polish task to ensure it does not get dropped.

### DIFF
--- a/packages/contracts/package.json
+++ b/packages/contracts/package.json
@@ -63,7 +63,7 @@
     "eslint-config-custom": "*",
     "ethereum-waffle": "^4.0.10",
     "ethers": "^5.7.2",
-    "hardhat": "^2.16.0",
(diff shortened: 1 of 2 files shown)


jakerockland:
Thank you for catching this and reverting ser 🙏

jakerockland (approved):

ryley-o:
added polish task for this activity 🙏

ryley-o:
fixed in #820

PR #838, opened 2023-06-27
Shared polyptych minter

ryley-o:
## Description of the change

Add shared Polyptych minter - one variant that supports ERC20 minting, one that supports ETH minting.

Uses our lib approach to mix the functionality of a couple minters such that parity with existing (non-shared) polyptych minter is reached:
- one variant uses logic in `SplitFundsLib` to support ERC20 payments
- uses logic in `MaxInvocationsLib` to support minter-local max invocations
- uses the logic in `TokenHolderLib` to perform parent token logic
- adds new functionality in specific to polyptych operations in `PolyptychLib`

This minter is designed to only support fixed price minting, which is believed to be the most common use case for polyptych minting (since it is gated to existing token holders)

## Related Updates
The shared randomizer will be updated to specify a "hash seed setter" contract on a per-project basis (vs. a per-contract basis). This will better enable different projects on a single contract to simultaneously use ERC20 minting and ETH minting. Those updates decoupled and in PR #848.

## Tests

- Tests are added to reach sufficient coverage on the new minters. Additionally, all critical functionality is intended to be tested.

## Review

- encourage looking at diffs between the new minters and fixed price + token holder minters during review

PR #848, opened 2023-06-30
Project-level hash seed setter contracts

ryley-o:
## Description of the change

Update shared randomizer from core-contract-scoped hash seed setter contracts to project-scoped hash seed setter contracts.

## Change Drivers

This change is driven by a need for more flexibility when defining which contract may set a project's hash seed. To date, only the polyptych minter has been used as a hash seed setter. However, in the future, we will have at least two polyptych minters in our shared minter suite (ETH and ERC20 variants in #838), and will likely have other future minters that may want to modify hash generation behaviors. Defining a hash setter for each project eliminates any concerns regarding simultaneous projects using different polyptych minters on a given contract.

## Security

This change does give slightly more power to artists, however all artist actions are highly audible and detectible. Additionally, artist's incentives generally align with not performing public actions that would severly damage their reputation (e.g. using their power to assign specific token hashes in an unfair manner).

Behavior before:
- Only contract admins could change the way a project's hash could be generated (by setting a contract-level hash seed setter contract)

Behavior after this change:
- Only artists can change the way a project's hash can be generated (by setting a project-level hash seed setter contract)

## Tests

- [x] update tests for these new behaviors
- [x] reach 100% coverage

## Other
We don't currently index shared randomizer data in our subgraphs or provide interactions from our frontend, so there is no required follow-on work from this change.


ryley-o:
@lyaunzbe - tagging you for review - this change was based off of some convo I had w/ @jakerockland, for the reasons described in the PR description.
I'll update the tests to be passing + 100% coverage, but I don't think that affects review much 👍

packages/contracts/contracts/randomizer/shared/SharedRandomizerV0.sol
```diff
@@ -180,7 +172,7 @@ contract SharedRandomizerV0 is ISharedRandomizerV0 {
         // does not require any gating.
```
jakerockland, line 172: Unrelated to this PR specifically, and can be separate follow up, but I think we probably want to update the documentation to more aggressively clarify that there are specific assumptions around this method being used atomically at the time of mint by the core contract that shares a relationship to the randomizer implementation.

jakerockland (approved):

packages/contracts/contracts/randomizer/shared/SharedRandomizerV0.sol
ryley-o: sounds good - very much agree this is communicated very weirdly 😅 will toss in a follow-up PR 🙏

PR #871, opened 2023-07-12
Shared SEA Minter

ryley-o:
## Description of the change

Add shared serial English auction minter that supports ETH minting.

Uses our lib approach to refactor the logic in the (never deployed) SEAV0 minter.

## Design Choices

Please see the internal design document for details on a few design decisions when making the SEA minter a shared minter. 
Link to comment in relevant DD section: https://docs.google.com/document/d/1Ix0LhPHqeDgCouKPSYZufQY9zQ5cNqf8yXHbus8y3w8/edit?disco=AAAA09wWgQ8

## Follow-on

This minter splits out two new libraries, that follow-on PRs should update other shared minters to use:
- AuthLib: internal library that contains logic to gate functions to artist, core contract adminACL, and/or minter filter adminACL
- ABHelpers: internal library that contains helper functions for common operations in the Art Blocks ecosystem of smart contracts. 

## Tests

- [x] TODO: add tests for coverage + effects of the new shared minter, and library functions
>note: writing of tests should not hold up review of this minter


ryley-o:
@jakerockland @lyaunzbe minor update in [516d621](https://github.com/ArtBlocks/artblocks-contracts/pull/871/commits/516d621aad6d3cead0ffaf5897993be0dc7fc835) that fixes the handling of an edge case where a project is reset during an active auction. packed storage, so incremental gas costs should be negligible.

.vscode/settings.json
```diff
@@ -21,5 +21,8 @@
   },
   "graphql-config.dotEnvPath": ".env",
   "solidity.packageDefaultDependenciesDirectory": "node_modules",
-  "graphql-codegen.configFilePath": "./graphql.config.ts"
+  "graphql-codegen.configFilePath": "./graphql.config.ts",
+  "[solidity]": {
+    "editor.defaultFormatter": "NomicFoundation.hardhat-solidity"
+  }
```
ryley-o, line 27: @lyaunzbe

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
```diff
@@ -0,0 +1,1169 @@
...
+
+    /// minterType for this minter
+    string public constant minterType = "MinterSEAV1";
+
+    /// minter version for this minter
+    string public constant minterVersion = "v1.0.0";
+
+    uint256 constant ONE_MILLION = 1_000_000;
+
+    uint256 public constant MIN_AUCTION_DURATION_SECONDS = 60; // seconds
```
ryley-o, line 120: note: hard-coding here results in a subgraph e2e test taking >60s. Alternatives could be to remove guardrail entirely, or make it defined in constructor.

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
```diff
@@ -0,0 +1,1169 @@
...
+    function _initializeAuctionWithBid(
+        uint256 _projectId,
+        address _coreContract,
+        uint256 _targetTokenId
+    ) internal {
+        SEALib.SEAProjectConfig storage _SEAProjectConfig = _SEAProjectConfigs[
+            _coreContract
+        ][_projectId];
+        SEALib.Auction storage _auction = _SEAProjectConfig.activeAuction;
+        // CHECKS
```
ryley-o, line 1018: @jakerockland - This was not part of our previous SEA minter design, but thinking we might want to add an additional constraint before initializing a new token auction.

Basically, if `msg.sender` is not the project artist, we should require that the project is not paused and is active on the core contract. This is more aligned with our typical interaction pattern.

The check adds gas to every auction-initialization transaction, but seems appropriate from a real-world security design standpoint.

ryley-o: (also, there are a couple ways to implement this, and I think it is already "implemented" for every auction **except the final token,** since `_tryMintTokenToNextSlot` does mint a next token through the core contract with the permissions of `msg.sender`)

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
```diff
@@ -0,0 +1,1169 @@
...
+            timestampStart: _timestampStart.toUint64(),
+            auctionDurationSeconds: _auctionDurationSeconds.toUint32(),
+            basePrice: _basePrice,
+            minBidIncrementPercentage: _minBidIncrementPercentage
+        });
+
+        // sync local max invocations if not initially populated
+        if (
+            !MaxInvocationsLib.maxInvocationsIsInitialized(
+                _maxInvocationsProjectConfig
```
ryley-o, line 365: @jakerockland - do you like the use of this library function here? If so, I can copy it to the other minters as well. If not, we can remove the function from the library, and make this minter consistent with the other minters.

packages/contracts/test/minter-suite-minters/SEA/MinterSEA/views.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  },
+  // {
+  //   core: "GenArt721CoreV3_Explorations",
+  // },
+  // {
+  //   core: "GenArt721CoreV3_Engine",
+  // },
+  // {
+  //   core: "GenArt721CoreV3_Engine_Flex",
+  // },
```
jakerockland, line 28: Is the commenting out of these intentional?

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
jakerockland: I do dig this yes – I think that is a great suggestion.

One nit, would suggest that we make the conditional branch here start with the positive though, in the way of:

```suggestion
        // sync local max invocations if not initially populated
        if (
            MaxInvocationsLib.maxInvocationsIsInitialized(
                _maxInvocationsProjectConfig
            )
        ) {
            // for convenience, try to mint to next token slot
            _tryMintTokenToNextSlot(_projectId, _coreContract);
        } else {
            syncProjectMaxInvocationsToCore(_projectId, _coreContract);
            // @dev syncProjectMaxInvocationsToCore function calls
            // _tryMintTokenToNextSlot, so we do not call it here.
        }
```

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
jakerockland: Could you clarify why we need to do this explicitly outside of the requirement already at the core contract level (e.g. if already handled for all tokens by way of the core contract requirements on msg.sender, I'm not sure I understand why we need to special case this)?

jakerockland:
@ryley-o just wanted to double-check that this is broadly ready for review before diving in?

packages/contracts/test/minter-suite-minters/SEA/MinterSEA/views.test.ts
ryley-o: it is for now - tests are still being written and doing this makes running them/iteration a bit quicker during development. They will be uncommented before tests are finalized.

ryley-o:
>@ryley-o just wanted to double-check that this is broadly ready for review before diving in?

@jakerockland - yes this is ready for review 👍 (I'm writing tests for coverage, but I don't expect those to affect review)

packages/contracts/contracts/libs/v0.8.x/minter-libs/MaxInvocationsLib.sol
```diff
@@ -122,4 +127,89 @@ library MaxInvocationsLib {
...
+     * both max invocations and maxHasBeenInvoked are still initial values.
+     * This is because if maxInvocations were ever set to zero,
+     * maxHasBeenInvoked would be set to true.
+     */
+    function maxInvocationsIsInitialized(
+        MaxInvocationsProjectConfig storage _maxInvocationsProjectConfig
+    ) internal view returns (bool) {
+        return
+            !(_maxInvocationsProjectConfig.maxInvocations == 0 &&
+                _maxInvocationsProjectConfig.maxHasBeenInvoked == false);
```
jakerockland, line 147: Nit: but I think this is easier to read, and I think _should_ be more gas efficient as it can short circuit this way on the first condition

```suggestion
            _maxInvocationsProjectConfig.maxInvocations != 0 || _maxInvocationsProjectConfig.maxHasBeenInvoked;
```

packages/contracts/contracts/libs/v0.8.x/minter-libs/SEALib.sol
```diff
@@ -0,0 +1,323 @@
...
+     * the last `_timeBufferSeconds` of the auction.
+     * NOTE: This function does not check if the bid is valid. It is assumed
+     * that the bid has already been checked for validity by the caller.
+     * @param _auction Auction to update
+     * @param _timeBufferSeconds Time buffer to extend the auction to if the
+     * bid is placed within the last `_timeBufferSeconds` of the auction.
+     * @param _bidAmount bid amount
+     * @param _bidder bidder's payable address
+     */
+    function AuctionUpdateBid(
```
jakerockland, line 131: Is there a reason we're uppercasing this method name?
```suggestion
    function auctionUpdateBid(
```

packages/contracts/contracts/libs/v0.8.x/minter-libs/SEALib.sol
```diff
@@ -0,0 +1,323 @@
...
+                _auction
+            );
+        isConfigured =
+            projectIsConfigured_ ||
+            auctionIsAcceptingIncreasingBids_;
+        // only return non-zero price if auction is configured
+        if (isConfigured) {
+            if (auctionIsAcceptingIncreasingBids_) {
+                // return minimum next bid, given current bid
+                tokenPriceInWei = SEALib.getMinimumNextBid(_auction);
```
jakerockland, line 297: Nit: if this is a library-internal function, can we underscore-case this and not refer to the library in the call (because we're calling from within the library)? 

```suggestion
                tokenPriceInWei = _getMinimumNextBid(_auction);
```

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
jakerockland: Hard-coding this value seems totally fine based on intended usage of this minter, and ease of upgradeability in the future in our new minter suite architecture that we're moving toward.

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
```diff
@@ -0,0 +1,1215 @@
...
+            _selector: this.updateRefundGasLimit.selector
+        });
+        // @dev max gas limit implicitly checked by using uint16 input arg
+        // @dev min gas limit is based on rounding up current cost to send ETH
+        // to a Gnosis Safe wallet, which accesses cold address and emits event
+        require(_minterRefundGasLimit >= 7_000, "Only gte 7_000");
+        // EFFECTS
+        minterRefundGasLimit = _minterRefundGasLimit;
+        emit MinterRefundGasLimitUpdated(_minterRefundGasLimit);
+    }
```
jakerockland, line 289: Curious if we want to make this a constant to simplify and reduce need for spurious minterfilter admin gates?

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
```diff
@@ -0,0 +1,1215 @@
...
+    function projectMaxInvocations(
+        uint256 _projectId,
+        address _coreContract
+    ) external view returns (uint256) {
+        return
+            uint256(
+                _maxInvocationsProjectConfigMapping[_coreContract][_projectId]
+                    .maxInvocations
+            );
+    }
```
jakerockland, line 630: Inclined a bit to make this something that just gets exposed by `MaxInvocationsLib` (and takes in a `_maxInvocationsProjectConfigMapping`), but maybe not worth – just know that this definitely exists across lots (all) of minters.

packages/contracts/contracts/libs/v0.8.x/minter-libs/AuthLib.sol
```diff
@@ -0,0 +1,158 @@
...
+import "../../../interfaces/v0.8.x/IMinterFilterV1.sol";
+
+/**
+ * @title Art Blocks Authorization Minter Library
+ * @notice This library contains helper functions that may be used by minters
+ * to check authorization for performing operations.
+ * @author Art Blocks Inc.
+ */
+
+library AuthLib {
```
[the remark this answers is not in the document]
jakerockland: On a related note, I'm inclined that this library lives in the higher level libs directory rather than minter specific, because I can see us using these auth lib methods for other things down the line.

jakerockland (commented):
Generally LGTM, but would love to do one more pass once tests are final and g2g just given how the size of this PR it feels worth doing a second pass when ready.

packages/contracts/contracts/libs/v0.8.x/minter-libs/AuthLib.sol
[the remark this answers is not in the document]
ryley-o: yeah - sorry this could be its own PR, but would prefer to just add to other minters in follow-on PR at this point (ABHelpers as well). good point that this could have been done as a separate PR to achieve a smoler PR 😅 
re: directory - good call, and going to bump this up a level (to be consistent with ABHelpers.sol)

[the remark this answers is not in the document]
ryley-o: updated in [ac464f5](https://github.com/ArtBlocks/artblocks-contracts/pull/871/commits/ac464f55e43c194a9c026ed099b88bf418c0ed23)

packages/contracts/test/minter-suite-minters/SEA/MinterSEA/views.test.ts
ryley-o: the tests are now ready and these are uncommented 👍

packages/contracts/contracts/libs/v0.8.x/minter-libs/MaxInvocationsLib.sol
ryley-o: good call - thanks for the suggestion!

packages/contracts/contracts/libs/v0.8.x/minter-libs/SEALib.sol
ryley-o: re: underscore - we actually call this directly from `MinterSEAV1` as well, so it is a typical library function, so we wouldn't make it private and use a leading underscore
re: not referring to library in the call, good call, and we did that a few places, all of which are updated in [90401a2](https://github.com/ArtBlocks/artblocks-contracts/pull/871/commits/90401a2dec350595832a59a8b2cd3daf92f5d779)

packages/contracts/contracts/libs/v0.8.x/minter-libs/SEALib.sol
ryley-o: updated in [90401a2](https://github.com/ArtBlocks/artblocks-contracts/pull/871/commits/90401a2dec350595832a59a8b2cd3daf92f5d779)

packages/contracts/contracts/libs/v0.8.x/minter-libs/SEALib.sol
```diff
@@ -0,0 +1,323 @@
...
+     * Overwrite the active auction for a project with a new auction.
+     * @dev This function is used to initialize a new auction. Care must be
+     * taken to ensure that the existing auction is fully complete and settled.
+     * @param _SEAProjectConfig SEAProjectConfig to update
+     * @param _targetTokenId token ID to create the auction for
+     * @param _bidAmount initial bid amount
+     * @param _bidder initial bidder's payable address
+     * @return endTime end time of the newly created auction
+     */
+    function OverwriteProjectActiveAuction(
```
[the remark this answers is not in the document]
ryley-o: updated in [90401a2](https://github.com/ArtBlocks/artblocks-contracts/pull/871/commits/90401a2dec350595832a59a8b2cd3daf92f5d779)

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
ryley-o: I *think* it is still worth making this admin-configurable on the first iteration of this minter. Until we release this into the wild, I'm hesitant to hard-code this value. Currently, making it configurable only adds this function - events still need to be emitted in constructor, handled in subgraph, etc.
If you have stronger feelings on this, happy to discuss more!

jakerockland: I don't feel super strongly here – if we think its worth having this flexibility, SGTM to keep it included.

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
ryley-o: The same thought applies to `projectMaxHasBeenInvoked` (which is actually a bit of a misnomer and maybe `projectMinterMaxHasBeenInvoked` would be better).
Personally, I like this suggestion (as ideally the minter does not need to be aware of struct variable names), and have added it as a follow-on PR task in Asana here: https://app.asana.com/0/0/1205103275547354/f
(since this has an asana task, going to resolve in this PR 👍 )

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
ryley-o: good call - updated in [c75484d](https://github.com/ArtBlocks/artblocks-contracts/pull/871/commits/c75484d7248638a68120c967015d5211419802b2)

packages/contracts/test/minter-suite-minters/SEA/MinterSEA/integration.test.ts
```diff
@@ -0,0 +1,1147 @@
...
+          // THIS DOES NOT CURRENTLY REVERT
+          // await expectRevert(
+          //   config.minter
+
+          //     .connect(config.accounts.user)
+          //     .createBid(targetTokenOne, config.genArt721Core.address, {
+          //       value: config.basePrice,
+          //     }),
+          //   "No new auctions when project is paused"
+          // );
```
ryley-o, line 622: @jakerockland - edge case I was trying to describe

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
ryley-o: yeah - sorry this is a weird edge case to describe - I wrote a test to describe the exact situation (linked below). It all has to do with our use of pre-minting the next token, so we don't actually rely on the core contract stopping a new mint for the last invocation.
https://github.com/ArtBlocks/artblocks-contracts/pull/871/files#r1269957682

ryley-o:
>Generally LGTM, but would love to do one more pass once tests are final and g2g just given how the size of this PR it feels worth doing a second pass when ready.

@jakerockland - tests are now final (with caveat of ongoing edge case discussion), so this is ready for a second pass! 🙏

jakerockland:
> > Generally LGTM, but would love to do one more pass once tests are final and g2g just given how the size of this PR it feels worth doing a second pass when ready.
> 
> @jakerockland - tests are now final (with caveat of ongoing edge case discussion), so this is ready for a second pass! 🙏

SG SG ser, doing another pass now.

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
jakerockland: Per our chat offline – I think this type of behavior is WAI and we shouldn't leak into the minter abstraction more than we need to re: paused state.

packages/contracts/test/minter-suite-minters/SEA/MinterSEA/integration.test.ts
ryley-o: as discussed, the current minter behavior actually does seem desirable, so will update this test to desire the current behavior 🙏 thanks!

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
ryley-o: thank you for chatting!

packages/contracts/contracts/libs/v0.8.x/minter-libs/AuthLib.sol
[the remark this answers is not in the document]
jakerockland: SGTM ser – thank you! 🙏 💜

packages/contracts/contracts/libs/v0.8.x/minter-libs/SEALib.sol
jakerockland: AH, missed that. makes sense to me ser – TYVM for the update here.

packages/contracts/contracts/minter-suite/Minters/MinterSEAV1.sol
jakerockland: Fantastic – sounds great to me ser 🙏

jakerockland (approved):
🙏 🚢 💜

PR #892, opened 2023-07-21
Dependency registry cleanup v0 updates

jakerockland:
## Description of the change

Minor cleanup/consistency updates, pulled from https://github.com/ArtBlocks/artblocks-contracts/pull/520 for the V0 dependency registry working draft.


packages/contracts/test/dependency-registry/DependencyRegistryV0.test.ts
```diff
@@ -1447,6 +1447,12 @@ describe(`DependencyRegistryV0`, async function () {
         expect(supportedCoreContracts).to.deep.eq([
           config.genArt721Core.address,
         ]);
+
+        const isSupportedCoreContract =
+          await config.dependencyRegistry.isSupportedCoreContract(
+            config.genArt721Core.address
+          );
+        expect(isSupportedCoreContract).to.eq(true);
```
ryley-o, line 1455: nit: should add a case to test the false case

ryley-o (approved):

packages/contracts/test/dependency-registry/DependencyRegistryV0.test.ts
jakerockland: Good call out – adding this in the "allows admin to remove supported core contract" case below as an additional check, seems best fit there.

PR #893, opened 2023-07-21
Use auth lib

ryley-o:
## Description of the change

Update all new shared minters to use the AuthLib introduced in #871.

This also cleans up some of the comments in `AuthLib.sol`.

Pure refactor - no functionality is changed.


ryley-o:
Cc @lyaunzbe for some updated patterns on how the new DA minters can use AuthLib.sol

jakerockland (approved):

PR #895, opened 2023-07-21
SSTORE2-compatible dependency registry (for scripts)

jakerockland:
## Description of the change

This PR makes a few interrelated changes:

- Updates the dependency registry implementation to allow for providing direct pointers (address locations of existing wrriten bytecode storage content) when specifying dependency scripts. 
- When providing direct pointers, allow support for both contracts written using the Art Blocks `BytecodeStorage` library implementation, but also provide fallback support for pointing to `SSTORE2`-written data.
- Provide a utility view method that allows introspection of whether or not a given piece of on-chain storage was written with our `BytecodeStorage` library or an unknown library (which we will fallback to reading as an `SSTORE2` written blob).

Appropriate test coverage is also added for these changes.

As discussed internally, the rationale for this change is to allow the reuse of referencing existing uploads of dependencies that may already exist on chain (e.g. an upload of p5js 1.0.0 that has already been uploaded on-chain via EthFS) as script versions that the dependency registry is aware of, without requiring the superfluous re-upload of these script references on-chain in their entirety.


packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -727,10 +776,31 @@ contract DependencyRegistryV0 is
             dependencyDetails[_dependencyType].scriptBytecodeAddresses[_index];
     }
 
+    /**
+     * @notice Returns the storage library version for
+     *         dependency `_dependencyTypes` at script index `_index`.
+     */
```
ryley-o, line 782: nit: add natspec for input params and return here. might be helpful to also describe fallback behavior to UNKNOWN_VERSION_STRING non-bytecode storage data (including SSTORE).

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -727,10 +776,31 @@ contract DependencyRegistryV0 is
...
+    function getDependencyScriptBytecodeStorageVersionAtIndex(
+        bytes32 _dependencyType,
+        uint256 _index
+    ) external view returns (bytes32) {
+        return
+            BytecodeStorageReader.getLibraryVersionForBytecode(
+                dependencyDetails[_dependencyType].scriptBytecodeAddresses[
+                    _index
```
ryley-o, line 790: nit: might be worth checking if this is the zero address and return with intuitive error message?

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -727,10 +776,31 @@ contract DependencyRegistryV0 is
...
+
     /**
      * @notice Returns script for dependency `_dependencyType` at script index `_index`.
      * @param _dependencyType dependency to be queried.
      * @param _index Index of script to be queried.
+     * @dev This method attempts to introspectively determine which library version of
+     *      `BytecodeStorage` was used to write the stored script string that is being
+     *      read back, in order to use the proper read approach. If the version is
+     *      non-determinate, a fall-back to reading using the assumption that the bytes
+     *      were written with `SSTORE2` is used.
```
ryley-o, line 803: nit: natspec for return value here

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -727,10 +776,31 @@ contract DependencyRegistryV0 is
             dependencyDetails[_dependencyType].scriptBytecodeAddresses[_index];
     }
 
+    /**
+     * @notice Returns the storage library version for
+     *         dependency `_dependencyTypes` at script index `_index`.
+     */
+    function getDependencyScriptBytecodeStorageVersionAtIndex(
+        bytes32 _dependencyType,
+        uint256 _index
```
ryley-o, line 785: nit: some places we call this `scriptId`, other places we call it `index`. probably should choose one or the other (I don't have strong opinion either way!)

packages/contracts/contracts/DependencyRegistryV0.sol
jakerockland: Good call out – updated (with notes on the version string expectations in a `@dev` block).

(Also, went through and made the documentation nomenclature for `_dependencyType` consistent throughout this file, as this seems like an area for possible confusion that is made clearer by just using the same documentation everywhere `@param _dependencyType Name of dependency type (i.e. "type@version") used to identify dependency.`)

packages/contracts/contracts/DependencyRegistryV0.sol
jakerockland: Good call-out; done!

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -742,7 +821,20 @@ contract DependencyRegistryV0 is
             return "";
         }
 
-        return _readFromBytecode(dependency.scriptBytecodeAddresses[_index]);
+        address scriptAddress = dependency.scriptBytecodeAddresses[_index];
+        bytes32 storageVersion = BytecodeStorageReader
+            .getLibraryVersionForBytecode(scriptAddress);
+
+        if (storageVersion == BytecodeStorageReader.UNKNOWN_VERSION_STRING) {
+            return
```
ryley-o, line 829: wondering if it would make sense to add this fallback logic into the bytecodestorage library itself, or just keep here?

ryley-o (approved):
LGTM - interested in your thoughts about specifying encoding (see comment)

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -742,7 +821,20 @@ contract DependencyRegistryV0 is
...
+                string(
+                    BytecodeStorageReader.readBytesFromSSTORE2Bytecode(
+                        scriptAddress
+                    )
```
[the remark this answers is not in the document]
ryley-o: for gzip specifically, we can just assume it is base64 encoded since we know it is EthFS, but if other libraries were uploaded to EthFS that made sense to use, we would need a way to specify their encoding. We can defer recording encoding for now as well, since this is an upgradeable contact, and we currently aren't planning on pointing to EthFS for any dependency data at the moment, IIUC.

packages/contracts/contracts/DependencyRegistryV0.sol
jakerockland: Good call out – going to go with `index` throughout (updated in recent commit), as this is a simpler language to use consistently across a) dependencies, b) repository URLs, and c) CDN URLs, rather than using varied nomenclature throughout. 

This departs a bit from how we refer to the script indexes on CoreV3 (where we use `scriptId`), but I think this is the preferable trade-off. 

That seem right to you @ryley-o ?

Note: I've done a broader pass over the entirety of our implementation here around this to try to improve consistency and readability, so an additional overall review pass would be much appreciated.

packages/contracts/contracts/DependencyRegistryV0.sol
jakerockland: Rather than checking if this is the zero address, I think its more appropriate to apply the check on index out of bounds querying here, above, and more holistically throughout this PR which I've applied in the latest commit. 

For example, I've also updated this behavior below, to fail in this manner rather than optimistically returning the empty string, which I'm not sure is behavior that makes sense (I think we should fail more clearly when attempting to read out of bounds content).

WDYT about this approach @ryley-o @yoshiwarab ?

packages/contracts/contracts/DependencyRegistryV0.sol
jakerockland: I don't love introducing this logic into the library itself, given that there isn't a great way to do on-chain introspection into a contract and _know_ that it was written with `SSTORE2`, so I like the design decision of pushing the specific fallback handling for `SSTORE2` reads onto the library client contract (such as this one), but could be definitely convinced otherwise (if you feel strongly, or if we got strong feedback from other prospective library clients in the future).

jakerockland:
> LGTM - interested in your thoughts about specifying encoding (see comment)

Thank you ser! Re-requesting another review pass if that is ok though, given meaningful changes made across the entirety of the file for consistency purposes based on feedback.

packages/contracts/contracts/DependencyRegistryV0.sol
ryley-o: sounds good to me! will cruise through another review 👍

packages/contracts/contracts/DependencyRegistryV0.sol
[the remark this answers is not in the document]
jakerockland: Hrm – this one is interesting to consider, will try to throw some time on for you + @yoshiwarab and I to talk about this a bit more synchronously.

That said, my inclination is that if we do want to go this route, I'd approach this in a follow up PR rather than updating this one with additional handling logic (though I will add the dev comment now).

Leaving this comment unresolved for now and will throw some time for us 3 onto the calendar.

packages/contracts/contracts/DependencyRegistryV0.sol
ryley-o: thanks for the solid points, and I agree with what you are saying here 🙏 SGTM!

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -97,6 +99,13 @@ contract DependencyRegistryV0 is
         );
     }
 
+    function _onlyInRangeIndex(
+        uint256 _index,
+        uint256 _maxIndex
```
ryley-o, line 104: nit: should this be called `_length` instead of `_maxIndex` (since length is maxIndex+1)?

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -191,33 +201,80 @@ contract DependencyRegistryV0 is
...
-        uint256 _scriptId,
+        uint256 _index,
         string memory _script
     ) external {
         _onlyAdminACL(this.updateDependencyScript.selector);
         _onlyNonEmptyString(_script);
         _onlyExistingDependencyType(_dependencyType);
         Dependency storage dependency = dependencyDetails[_dependencyType];
-        require(_scriptId < dependency.scriptCount, "scriptId out of range");
+        _onlyInRangeIndex(_index, dependency.scriptCount);
```
ryley-o, line 220: nit: using named args might be appropriate here (for all uses in this contract) since we have two adjacent values with the same type

packages/contracts/contracts/DependencyRegistryV0.sol
ryley-o: ah, yes this makes sense to me 👍

ryley-o (approved):

yoshiwarab (approved):

packages/contracts/contracts/DependencyRegistryV0.sol
jakerockland: Vibey – TY ser!

packages/contracts/contracts/DependencyRegistryV0.sol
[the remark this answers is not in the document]
jakerockland: Per our conversation today synchronously – I think we treat this as an area of follow up work, but decouple that from this initial PR. 

I've filed this issue (https://github.com/ArtBlocks/artblocks-contracts/issues/911) as one for myself to follow up with on an additional PR for you two to review @yoshiwarab @ryley-o.

packages/contracts/contracts/DependencyRegistryV0.sol
jakerockland: Good call out – updated per your suggestion in latest commit.

packages/contracts/contracts/DependencyRegistryV0.sol
jakerockland: Good call out – updated per your suggestion in latest commit.

PR #896, opened 2023-07-21
Helpers in MaxInvocationsLib.sol

ryley-o:
## Description of the change

Use Library getters and helpers in all new minters.

This is follow-on work based off the feedback received in the following comments:
- add new getter functions: https://github.com/ArtBlocks/artblocks-contracts/pull/871#discussion_r1269762757
- use `maxInvocationsIsInitialized` across all new minters: https://github.com/ArtBlocks/artblocks-contracts/pull/871#discussion_r1268756860

---
Asana task:  https://app.asana.com/0/0/1205103275547354/f


jakerockland (changes requested):
This generally LGTM.

But one bigger suggestion here – it seems like the intended use case of `maxInvocationsIsInitialized` is actually intended to check the opposite (its always used w/ a boolean not) so would be a bit more readable to re-write as:

```
    function maxInvocationsIsUnconfigured(
        MaxInvocationsProjectConfig storage _maxInvocationsProjectConfig
    ) internal view returns (bool) {
        return
            _maxInvocationsProjectConfig.maxInvocations == 0 &&
            !_maxInvocationsProjectConfig.maxHasBeenInvoked;
    }
 ```

WDYT @ryley-o ?

ryley-o:
>But one bigger suggestion here – it seems like the intended use case of maxInvocationsIsInitialized is actually intended to check the opposite (its always used w/ a boolean not) so would be a bit more readable to re-write as:

ah, good call. I generally would stay away from too many "not x" terms when returning booleans, but in this case since we won't be ever using it as `if (!notInitialized)`, I think makes sense.

Will update and re-ping you when ready 👍

jakerockland:
> > But one bigger suggestion here – it seems like the intended use case of maxInvocationsIsInitialized is actually intended to check the opposite (its always used w/ a boolean not) so would be a bit more readable to re-write as:
> 
> ah, good call. I generally would stay away from too many "not x" terms when returning booleans, but in this case since we won't be ever using it as `if (!notInitialized)`, I think makes sense.
> 
> Will update and re-ping you when ready 👍

+++; generally agree that boolean methods should return a positive in the general case, but in this situation I think opposite makes sense given that we're really always using it to look for the uninitialized state. TY ser 🙏

jakerockland (approved):
TYVM for the updates here 🙏

PR #901, opened 2023-07-24
Deploy shared minters

ryley-o:
## Description of the change

Add deployment scripts for new shared minters.

This adds deployment scripts that enable config-based deployments of the new shared minter contracts, as well as globally allowlisting the new contracts on their shared minter filters.

Note that a dev deployment was performed to ensure the script is WAI. If this is desired to be changed in the future, we can always re-deploy on dev.

---

Asana task: https://app.asana.com/0/0/1205113430357542/f


packages/contracts/scripts/minter-deployments/shared-minters-deployer.ts
```diff
@@ -0,0 +1,195 @@
...
+        ASSUMED_MINTER_FILTER_NAME,
+        deployDetails.minterFilterAddress
+      );
+      const tx = await minterFilter.approveMinterGlobally(minterAddress);
+      await tx.wait();
+      console.log(
+        `[INFO] ${deployDetails.minterName} approved globally on minter filter at ${deployDetails.minterFilterAddress}`
+      );
+      await delay(EXTRA_DELAY_BETWEEN_TX);
+    }
```
jakerockland, line 122: I think we should try catch here – to gracefully fail when the connected wallet doesn't actually have the ability to perform this step.

jakerockland (approved):
LGTM, with one note re: graceful failure mode.

packages/contracts/scripts/minter-deployments/shared-minters-deployer.ts
ryley-o: good call - updated in [d077ad5](https://github.com/ArtBlocks/artblocks-contracts/pull/901/commits/d077ad50496abcbac2326631ccd3f60b969bec75)

PR #910, opened 2023-07-26
Shared minter suite - mint gas tests

ryley-o:
## Description of the change

Quantify mint gas cost changes for the new shared minter suite.

Gas costs are expected to go up a bit for the new shared minter suite, because core contract of a project must be passed as a parameter instead of assumed (essentially hard-coded into bytecode during minter deployment in non-shared minters). We also might lose some struct packing efficiency due to our use of many structs across different libraries, instead of a custom struct fully optimized for each minter.

This PR adds gas tests to quantify the changes, and we may choose to gas golf with respect to these tests in follow-on PRs if desired.

## Results

| Comparison | Legacy gas used | New gas used | Increase (%) |
| --- | --- | --- | --- |
| Minter Set Price  | 116652 | 120869 | +3.6% |
| Minter Merkle | 128284 | 134617 | +4.9% |
| Minter DA Exp | 121308 | 128330 | +5.8% |
| Minter DA Exp w/Settlement | 103393 | 108354 | +4.8% |

> specific tests not added for holder & ERC20 minters, because they have no reason to be significantly different from the trends shown above

## Conclusion

In general, increases on the order of ~5% are reasonable for the new architecture, and indicate that we have done nothing dramatically inefficient while adding the new capabilities to the minters.

---
Asana task: https://app.asana.com/0/0/1205139431852175/f


ryley-o:
Cc @lyaunzbe for visibility on these.

lyaunzbe (approved):

ryley-o:
Cc @jakerockland - bypassed branch protection rules

PR #919, opened 2023-07-28
Engine deploys (07/28)

jakerockland:
## Description of the change

Engine deployments for:

(goerli / staging) : 
- https://github.com/ArtBlocks/artblocks/issues/1183
- https://github.com/ArtBlocks/artblocks/issues/1314

(mainnet / prod) : 
- https://github.com/ArtBlocks/artblocks/issues/1283
- https://github.com/ArtBlocks/artblocks/issues/1291


ryley-o (approved):
I see a few extra mainnet deployments in the grailers dao flex due to what looks like a few issues encountered.

This has the side-effect of those contracts being included in our subgraph (e.g. our mainnet subgraph includes the abandoned `0x324bc45c9f257f177166be4f6e7d2f551b5a1f03` deployment).

I think at a minimum we should unregister those deployments from the EngineRegistry, and potentially follow-on update our subgraph handler to remove the entity (see logic on why we didn't remove the entity [here](https://github.com/ArtBlocks/artblocks-subgraph/blob/d09cf54e7341396445b0b1f73ab22105855b2e22/src/core-registry.ts#L73), but I think in these cases, it might make sense to remove the contract from the store). Thoughts?

jakerockland:
> I see a few extra mainnet deployments in the grailers dao flex due to what looks like a few issues encountered.
> 
> This has the side-effect of those contracts being included in our subgraph (e.g. our mainnet subgraph includes the abandoned `0x324bc45c9f257f177166be4f6e7d2f551b5a1f03` deployment).
> 
> I think at a minimum we should unregister those deployments from the EngineRegistry, and potentially follow-on update our subgraph handler to remove the entity (see logic on why we didn't remove the entity [here](https://github.com/ArtBlocks/artblocks-subgraph/blob/d09cf54e7341396445b0b1f73ab22105855b2e22/src/core-registry.ts#L73), but I think in these cases, it might make sense to remove the contract from the store). Thoughts?

Yesser! Much agreed, I have some post-mortem time set up for us to go over what happened here, what next steps we need to take as immediate follow up, and how to better handle this in the future 🙏 💜

PR #920, opened 2023-07-28
Engine deployment scripts shared minter suite

ryley-o:
## Description of the change

Update Engine deployment scripts to use new shared minter suite.

This PR implements the following changes:

- Updates Engine deploy config template to only relevant parameters when using shared minter suite
- Deprecates existing generic Engine deploy scripts
- Implements an updated generic Engine deploy script that deploys new Engine contracts that use the new shared minter suite (and shared randomizer)
- Defines constants that recored the active set of shared minter filters and shared randomizers, as well as provide utility functions to get the active core registry being used by the shared minter filter.

Due to shared Engine contracts requiring shared contracts be already deployed, existing engine deployment scripts remain functional, but are labeled as deprecated.

---

Asana task: https://app.asana.com/0/0/1205139431852192/f

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205234702001143

PR #943, opened 2023-08-08
Limit max network gas price during deployments

ryley-o:
## Description of the change

Limit max network gas price during deployments.

This implementation does the following:
- configures a max gas price per network in the hardhat config file
- prior to deployments, the scripts are updated to perform a single check that ensures nominal network gas price is lte the configured max gas price for the network being deployed to

Note that if gas were to significantly spike during deployments, this implementation does not pause an ongoing deployment. Only an initial check is performed. This seemed like a reasonable design choice, but gladly accept any feedback in this area.

## Follow-on
#920 MUST be updated with these changes after this is merged!


ryley-o:
@yoshiwarab I decided to stop wasting time on augmenting the hardhat NetworkConfig type to add the custom `maxNominalGasPriceGwei` parameter. If it is obvious to you how to do that, feel free to push to this PR. If not, probably not worth our time (since we aren't operating in strict mode).

jakerockland (approved):
General changeset here LGTM – thank you @ryley-o !

That said, I am wondering if we may want to have the calling of this new `requireLowGasPrice` util be something that happens via our `package.json` file (calling this as a separate command that happens `&&`'d before the deployment script calls) – thinking here is that this may be a bit less error prone (easier to catch if forgotten) than requiring that it be called inline within each relevant deployment script?

I don't feel super strongly on this, but did want to raise for discussion.

ryley-o:
>I am wondering if we may want to have the calling of this new requireLowGasPrice util be something that happens via our package.json file

@jakerockland I really like this suggestion to reduce the chance of errors, and implemented in [2d508b2](https://github.com/ArtBlocks/artblocks-contracts/pull/943/commits/2d508b23760336bedab44979a43481dc9793ff9f)

packages/contracts/scripts/util/utils.ts
```diff
@@ -1,12 +1,50 @@
...
+  );
+  if (gasPriceGwei > maxNominalGasPriceGwei) {
+    throw new Error(
+      `[ERROR] Gas price is ${gasPriceGwei.toString()} gwei, which is above the configured max nominal gas price of ${maxNominalGasPriceGwei} gwei`
+    );
+  }
+  console.log(
+    `[INFO] Gas price is ${gasPriceGwei.toString()} gwei, which is lt the configured max nominal gas price of ${maxNominalGasPriceGwei} gwei`
+  );
+}
```
jakerockland, line 46: Nit: I would think the only place we anticipate this to be used is in `packages/contracts/scripts/require-low-gas.ts` itself, would suggest that we move it there for now and can split off into utils later if actually needed?

jakerockland (approved):
TY for the update ser – I dig it! 🙏

packages/contracts/scripts/util/utils.ts
ryley-o: good call, updated!

PR #971, opened 2023-08-22
Refactor set price and purchase

ryley-o:
## Description of the change

Refactor set price minters to use SetPriceLib for consistency, and break out simple purchase functions to specific interface.

This fixes two issues with the current codebase:
- removes unnecessary, inactive `purchase` and `purchaseTo` functions on minters that require additional args when minting
- moves the `projectConfig` struct to its own library, because it is not used by _every_ new minter (e.g. DA minters ended up not using the struct)

 tests are updated to handle the new name of the view function providing visibility of the fixed price config struct.


ryley-o:
Cc @lyaunzbe here in case any additional thoughts on this refactor

packages/contracts/contracts/libs/v0.8.x/minter-libs/SetPriceLib.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+// import "../../../interfaces/v0.8.x/IMinterBaseV0.sol";
+// import "../../../interfaces/v0.8.x/IGenArt721CoreContractV3_Base.sol";
+// import "../../../interfaces/v0.8.x/IGenArt721CoreContractV3.sol";
+// import "../../../interfaces/v0.8.x/IGenArt721CoreContractV3_Engine.sol";
```
lyaunzbe, line 7: nit: can these be removed since unused/commented out?

jakerockland (approved):
🙏 🙏

lyaunzbe (approved):

packages/contracts/contracts/libs/v0.8.x/minter-libs/SetPriceLib.sol
ryley-o: oops! nice catch - commented these out to check that they were unused imports, then forgot to remove! updated!

PR #972, opened 2023-08-22
Use authLib in polyptych erc20 minter

ryley-o:
## Description of the change

Use authLib in polyptych erc20 minter.

This is a bug fix that should have been implemented originally.


jakerockland (approved):
🙏 💜

PR #984, opened 2023-08-23
View for all contract approved minters

ryley-o:
## Description of the change

Add view that returns all minters approved for a contract on MinterFilterV2.

(follow-on from #983)

This adds a gas-unbounded convenience function that returns a list of all minters approved on a given contract on MinterFilterV2.

This is intended to be useful for artists configuring their minters. The functions also help with providing an easy way to determine the set of minters that are globally approved on a given contract, which is a net-benefit to our overall security model.

Tests are added for the new view functionality.

## NOTE
A minor bug was fixed as part of this PR that calls `minterType` from `ISharedMinterV0` instead of the legacy `IFilteredMinterV0`. This doesn't affect any actual call behavior, but is more appropriate, since minters for `MinterFilterV2` are expected to be shared minters, not legacy minters.


lyaunzbe (approved):

ryley-o:
bypassed branch protection, Cc @jakerockland

PR #988, opened 2023-08-24
Reduce DA mint gas costs

ryley-o:
## Description of the change

Reduce DA mint gas costs via struct packing auction data.

This imposes stricter limits on
- max auction prices (max price of ~300 million ETH)
- only allows our minter to function for the next 34 thousand years
...but has the benefit of reducing DA mint gas costs by 2%. All DA minters benefit from this change, since the packing is implemented in the DA libraries used by the minters.

| Comparison | Before gas used | New gas used | Increase (%) |
| --- | --- | --- | --- |
| Minter Shared DA Exp w/Settlement | 108377 | 106254 | -2% |

## Recommendation
The new maximum values are acceptable for our ETH minters, so this seems like an easy 2% efficiency gain.

Note: both previous and new max limits on price are likely NOT acceptable for ERC20 minters, which likely need entire uint256 domain to define price.


lyaunzbe (approved):
👍 🚀

ryley-o:
Cc @jakerockland bypassed branch protection rules

PR #989, opened 2023-08-24
Use uint256 in interfaces of price

ryley-o:
## Description of the change

Use `uint256` in interfaces dealing with price values.

In general, we want to pack prices in structs for gas efficiency in our minters. When dealing with ETH minters, maximum values are known well (300 million ETH is expected to be a safe upper price bound for an NFT sale).

However, if we were to develop a DA ERC20 minter, both decimals and upper bounds of pricing are no longer known, and uint256 must be used for price storage and interfaces.

## Design Choices

This update implements the following:
- updates interface functions and events to always communicate price values as `uint256`
  - this adds negligible gas cost, while making the current events compatible with ERC20 variants if we implement in the future (eliminating downstream impacts on subgraph and frontend)
- keep prices packed in storage
  - this avoids adding non-negligible gas costs to mint transactions
  - the downside is we will need to fork some libraries (e.g. make a `DAExpNotPackedLib`) to make them ERC20 compatible in the future, which seems like a reasonable tradeoff considering the gas efficiency tradeoffs associated with not packing prices
    - estimate ~4-5% additional mint gas costs if we were to fully unpack the ETH variants of these libs
- function inputs to the minter accept `uint256`, but function inputs to the underlying libraries accept the type that is used in storage. This separates the roles of the library and the minter: the library lets the minter decide if it wants to have function inputs equal to the storage type within the library or uint256.

### DIFF
--- a/packages/contracts/contracts/interfaces/v0.8.x/ISharedMinterDAExpSettlementV0.sol
+++ b/packages/contracts/contracts/interfaces/v0.8.x/ISharedMinterDAExpSettlementV0.sol
@@ -11,7 +11,7 @@ interface ISharedMinterDAExpSettlementV0 is ISharedMinterDAExpV0 {
         uint256 indexed _projectId,
         address indexed _coreContract,
         uint24 _numPurchased,
-        uint232 _netPosted
+        uint256 _netPosted
     );
 
     /// returns latest purchase price for project `_projectId`, or 0 if no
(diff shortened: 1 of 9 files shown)


lyaunzbe (approved):

ryley-o:
Cc @jakerockland - bypassed branch protection rules

Issue #929, opened 2023-08-02
Add dependency license concept to DependencyRegistry V0


jakerockland:
This should be reflected in the dependency info struct – likely want to only support a fixed set of known licenses to provide basic data validation.

ryley-o:
This was implemented and is in use 😄 marking closed

PR #999, opened 2023-08-28
Dependency reg dev

sterlingcrispin:
## Description of the change

This addresses #929 "Add dependency license concept to DependencyRegistry V0"

With this change there's a  `Bytes32Set _licenseTypes` which holds potential license types, and each `Dependency` struct now stores a `bytes32 licenseType ` of which license that dependency uses

When new dependencies are added, they must pass an argument that says what license the dependency is. This license must already exist in the set of  `_licenseTypes`.

New license types can be registered via `addLicenseType`

There's also some minor optimization of the existing contract, for example in `addDependency` rather than checking for `contains` before adding the dependency, 

```
        require(
            !_dependencyTypes.contains(_dependencyType),
            "Dependency type already exists"
        );
        _dependencyTypes.add(_dependencyType);
```

Instead, the function just tries to add the dependency, which returns false if the set already contained the dependency 

```
        require(
            // @dev the add function returns false if set already contains value
            _dependencyTypes.add(_dependencyType),
            "Dependency type already exists"
        );
```

The same pattern is now also used for adding and removing `_supportedCoreContracts`

@ryley-o @jakerockland


packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -633,11 +669,49 @@ contract DependencyRegistryV0 is
...
+     * @notice Returns number of registered license types
+     * @return Number of registered license types.
+     * @dev This is only intended to be called outside of block
+     * execution where there is no gas limit.
+     */
+    function getLicenseTypes() external view returns (string[] memory) {
+        string[] memory licenseTypes = new string[](
+            _licenseTypes.length()
+        );
+        uint256 numLicenseTypes = _licenseTypes.length();
```
ryley-o, line 682: nit - would rather switch the order of these two statements, and only call `_licenseTypes.length()` once. We should follow that pattern in some of the other view functions in this contract as well (might make sense as follow-on work if you would like!)

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -114,39 +129,60 @@ contract DependencyRegistryV0 is
         _transferOwnership(_adminACLContract);
     }
 
+    /**
+     * @notice Adds a new license type that can be used by dependencies.
+     */
+    function addLicenseType(bytes32 _licenseType) external {
```
ryley-o, line 135: I think it makes sense to only provide the ability to add new license types. Just wanted to double click and ensure you agree!

packages/contracts/contracts/DependencyRegistryV0.sol
ryley-o, line 682: nit - would rather switch the order of these two statements, and only call `_licenseTypes.length()` once. We should follow that pattern in some of the other view functions in this contract as well (might make sense as follow-on work if you would like!)

packages/contracts/contracts/interfaces/v0.8.x/IDependencyRegistryV0.sol
```diff
@@ -18,8 +18,13 @@ interface IDependencyRegistryV0 {
         uint256 indexed _projectId
     );
 
+    event LicenseTypeAdded(
+        bytes32 _licenseType
```
ryley-o, line 22: nit: let's make this indexed to enable filtering to a specific license type

ryley-o (approved):
LGTM 🙏 couple of very minor nits/double-checks, and would like @jakerockland to give this another set of eyes once he returns, prior to merging into main.

packages/contracts/contracts/DependencyRegistryV0.sol
jakerockland: ++

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -114,39 +129,60 @@ contract DependencyRegistryV0 is
...
+        _onlyAdminACL(this.addLicenseType.selector);
+        require(
+            _licenseTypes.add(_licenseType),
```
jakerockland, line 138:
```suggestion
            // @dev the add function returns false if set already contains value
            _licenseTypes.add(_licenseType),
```

jakerockland (approved):

packages/contracts/contracts/DependencyRegistryV0.sol
sterlingcrispin: sounds good will close this comment

packages/contracts/contracts/DependencyRegistryV0.sol
sterlingcrispin: Resolved here as an additional commit to this PR -> https://github.com/ArtBlocks/artblocks-contracts/pull/999/commits/c671723691f577f11b74f54f6f4890c325283a7a

PR #1017, opened 2023-09-07
Dependency reg dev on chain license

sterlingcrispin:
## Adds storage of full license text to dependency registry

This PR builds off of #999 and adds the ability to store the full text of a software license using the BytecodeStorageWriter in a similar pattern to the dependency script storage that currently exists. 

New functions 

- addLicenseText
- updateLicenseText
- getLicenseText
- removeLicenseLastText

Right now it doesn't include pointer functions like

-adding a license pointer
-updating that license pointer

@ryley-o @lyaunzbe questions for you:

Should I add those pointer functions in or is that overkill?

At this point I think it makes sense to split the on chain script storage, and on chain license storage into two new contracts at this point and have the dependency registry refer to them for the data.  Do you agree?

The script and license storage code follow really similar patterns right now. That whole concept could be refactored into its own contract that both the new script storage contract and license storage contract inherit from. Should I go ahead and do that while addressing the previous point?


ryley-o:
Thank you @sterlingcrispin 🙏 A few initial thoughts from me, curious on different perspectives from @lyaunzbe & @jakerockland 
 
>Should I add those pointer functions in or is that overkill?

I think this is overkill for now, as I'm not aware of any license text that has already been uploaded to ethereum via SSTORE2 (meaning we would probably not be using this functionality for now). This is also easy to add on in a contract upgrade if needed later on. 

>At this point I think it makes sense to split the on chain script storage, and on chain license storage into two new contracts at this point and have the dependency registry refer to them for the data. Do you agree?

I certainly see the benefits of splitting, however, I think it still makes sense to keep this all in one contract. If we still have all functions to read and/or write on the dependency registry, adding in a layer of contracts calling each other adds gas costs, especially when there are large strings involved. We could pass pointers around to eliminate that, but then we might be getting less readable? If we fully delegate all read/write functionality to the two new contracts, I think integrating with this dependency registry gets a little more complex, and performing contract upgrades would now need to happen to potentially three contracts instead of one.

In general, we tend to be okay with repeating code patterns in a contract if it improves readability and overall integration simplicity. We could potentially use an internal library (or internal helper functions) if we think there are chunks of code that are common and make sense to define in a single place.

Curious on thoughts from others here!

>The script and license storage code follow really similar patterns right now. That whole concept could be refactored into its own contract that both the new script storage contract and license storage contract inherit from. Should I go ahead and do that while addressing the previous point?

I think mostly same response as above, with added note that we have generally opted for libraries/helper functions over inheritance to help keep our contract source code a bit more straightforward when possible.

jakerockland:
> Thank you @sterlingcrispin 🙏 A few initial thoughts from me, curious on different perspectives from @lyaunzbe & @jakerockland
> 
> > Should I add those pointer functions in or is that overkill?
> 
> I think this is overkill for now, as I'm not aware of any license text that has already been uploaded to ethereum via SSTORE2 (meaning we would probably not be using this functionality for now). This is also easy to add on in a contract upgrade if needed later on.

++; would agree with your rationale here @ryley-o 🙏 

> > At this point I think it makes sense to split the on chain script storage, and on chain license storage into two new contracts at this point and have the dependency registry refer to them for the data. Do you agree?
> 
> I certainly see the benefits of splitting, however, I think it still makes sense to keep this all in one contract. If we still have all functions to read and/or write on the dependency registry, adding in a layer of contracts calling each other adds gas costs, especially when there are large strings involved. We could pass pointers around to eliminate that, but then we might be getting less readable? If we fully delegate all read/write functionality to the two new contracts, I think integrating with this dependency registry gets a little more complex, and performing contract upgrades would now need to happen to potentially three contracts instead of one.
> 
> In general, we tend to be okay with repeating code patterns in a contract if it improves readability and overall integration simplicity. We could potentially use an internal library (or internal helper functions) if we think there are chunks of code that are common and make sense to define in a single place.
> 
> Curious on thoughts from others here!

Not sure I understand the suggestion here – is it to manage the storing of dependencies vs. licenses in separate registry contracts? If so, would agree with your assessment here @ryley-o that that is not worthwhile. That said, if suggestion is just storing these in separate storage contract blobs in on-contracts storage, then yes would agree they should be split.

> > The script and license storage code follow really similar patterns right now. That whole concept could be refactored into its own contract that both the new script storage contract and license storage contract inherit from. Should I go ahead and do that while addressing the previous point?
> 
> I think mostly same response as above, with added note that we have generally opted for libraries/helper functions over inheritance to help keep our contract source code a bit more straightforward when possible.

+++ re: libraries and protocol-oriented composability >>> inheritance patterns where possible, generally I think this feeds better "SOLID principle" conforming code. 

If we think that refactoring to library-ize here is worthwhile, I'm supportive though don't feel strongly on the matter.

jakerockland (approved):

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -114,39 +135,58 @@ contract DependencyRegistryV0 is
         _transferOwnership(_adminACLContract);
     }
 
+    /**
+     * @notice Adds a new license type that can be used by dependencies.
+     */
```
ryley-o, line 140: nit: add natspec for `@param _licenseType`

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -114,39 +135,58 @@ contract DependencyRegistryV0 is
...
+    function addLicenseType(bytes32 _licenseType) external {
+        _onlyAdminACL(this.addLicenseType.selector);
+        // @dev the add function returns false if set already contains value
+        require(_licenseTypes.add(_licenseType), "License type already exists");
+        require(
+            _licenseType != bytes32(""),
+            "License type cannot be empty string"
+        );
```
ryley-o, line 148: nit: swap these two statements, as that order more closely follows the [checks-effects-interactions pattern](https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html)

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -114,39 +135,58 @@ contract DependencyRegistryV0 is
...
+    /**
+     * @notice Adds a new license type that can be used by dependencies.
+     */
+    function addLicenseType(bytes32 _licenseType) external {
+        _onlyAdminACL(this.addLicenseType.selector);
+        // @dev the add function returns false if set already contains value
+        require(_licenseTypes.add(_licenseType), "License type already exists");
+        require(
+            _licenseType != bytes32(""),
+            "License type cannot be empty string"
```
ryley-o, line 147: nit: typically prefer error messages to be <=31 characters, such that they fit into a single, 32-byte chunk. This helps reduce contract bytecode size (and deployment costs).

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -633,11 +671,47 @@ contract DependencyRegistryV0 is
         return _dependencyTypes.at(_index).toString();
     }
 
+    /**
+     * @notice Returns number of registered license types
+     * @return Number of registered license types.
```
ryley-o, line 676:
```suggestion
     * @notice Returns array of registered license types
     * @return Array of registered license types.
```

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -856,6 +932,129 @@ contract DependencyRegistryV0 is
...
+     * @notice Adds a full license text to license `_licenseType`, by way of
+     *         providing a string to write to bytecode storage.
+     * @param _licenseType Name of license type (i.e. "MIT") used to identify license.
+     * @param _text Text to be added. Required to be a non-empty string,
+     *                but no further validation is performed.
+     */
+    function addLicenseText(
+        bytes32 _licenseType,
+        string memory _text
+    ) external {
```
ryley-o, line 945: lets bring these external functions (external view are fine) up to the block of external functions, such that we follow the [solidity style guide's order of functions](https://docs.soliditylang.org/en/latest/style-guide.html#order-of-functions)

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -856,6 +932,129 @@ contract DependencyRegistryV0 is
...
+        _onlyAdminACL(this.updateLicenseText.selector);
+        _onlyNonEmptyString(_text);
+        _onlyExistingLicenseType(_licenseType);
+        License storage licenseEntry = allLicenses[_licenseType];
+        _onlyInRangeIndex({
+            _index: _index,
+            _length: licenseEntry.licenseChunkCount
+        });
+        // store license text chunk in contract bytecode, replacing reference address from
+        // the contract that no longer exists with the newly created one
```
ryley-o, line 981:
```suggestion
        // the contract that is no longer referenced with the newly created one
```

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -856,6 +932,129 @@ contract DependencyRegistryV0 is
...
+     * @notice Removes the last license text chunk from license `_licenseType`.
+     * @param _licenseType Name of license type (i.e. "MIT") used to identify license.
+     */
+    function removeLicenseLastText(bytes32 _licenseType) external {
+        _onlyAdminACL(this.removeLicenseLastText.selector);
+        _onlyExistingLicenseType(_licenseType);
+        License storage licenseEntry = allLicenses[_licenseType];
+        require(
+            licenseEntry.licenseChunkCount > 0,
+            "there is no license text to remove"
```
ryley-o, line 997: nit: prefer error messages to be <= 31 characters

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -856,6 +932,129 @@ contract DependencyRegistryV0 is
...
+    /**
+     * @notice Removes the last license text chunk from license `_licenseType`.
+     * @param _licenseType Name of license type (i.e. "MIT") used to identify license.
+     */
+    function removeLicenseLastText(bytes32 _licenseType) external {
+        _onlyAdminACL(this.removeLicenseLastText.selector);
+        _onlyExistingLicenseType(_licenseType);
+        License storage licenseEntry = allLicenses[_licenseType];
+        require(
+            licenseEntry.licenseChunkCount > 0,
```
ryley-o, line 996: nit: could load `licenseEntry.licenseChunkCount` into memory to avoid a few more SLOADs in this function. okay to leave as-is if you feel that degrades readability too much.

ryley-o (approved):
This LGTM! just minor nits from me 🥳 

Overall, this seems like the right approach here, as the dependency registry is upgradeable, and we can add functions for edge-cases as needed. If we do find licenses already uploaded via SSTORE2, I think it would make sense to add a function to add/update license pointers, but I'm not aware of any right now (haven't looked in detail though!). Another thing we might encounter would be if a dependency needed to update its license, but that also seems very unlikely, and could probably only happen if we messed up when uploading the data (in which case we could add that function+event in a contract upgrade to fix).

lyaunzbe:
> > Thank you @sterlingcrispin 🙏 A few initial thoughts from me, curious on different perspectives from @lyaunzbe & @jakerockland
> > > Should I add those pointer functions in or is that overkill?
> > 
> > 
> > I think this is overkill for now, as I'm not aware of any license text that has already been uploaded to ethereum via SSTORE2 (meaning we would probably not be using this functionality for now). This is also easy to add on in a contract upgrade if needed later on.
> 
> ++; would agree with your rationale here @ryley-o 🙏
> 
> > > At this point I think it makes sense to split the on chain script storage, and on chain license storage into two new contracts at this point and have the dependency registry refer to them for the data. Do you agree?
> > 
> > 
> > I certainly see the benefits of splitting, however, I think it still makes sense to keep this all in one contract. If we still have all functions to read and/or write on the dependency registry, adding in a layer of contracts calling each other adds gas costs, especially when there are large strings involved. We could pass pointers around to eliminate that, but then we might be getting less readable? If we fully delegate all read/write functionality to the two new contracts, I think integrating with this dependency registry gets a little more complex, and performing contract upgrades would now need to happen to potentially three contracts instead of one.
> > In general, we tend to be okay with repeating code patterns in a contract if it improves readability and overall integration simplicity. We could potentially use an internal library (or internal helper functions) if we think there are chunks of code that are common and make sense to define in a single place.
> > Curious on thoughts from others here!
> 
> Not sure I understand the suggestion here – is it to manage the storing of dependencies vs. licenses in separate registry contracts? If so, would agree with your assessment here @ryley-o that that is not worthwhile. That said, if suggestion is just storing these in separate storage contract blobs in on-contracts storage, then yes would agree they should be split.
> 
> > > The script and license storage code follow really similar patterns right now. That whole concept could be refactored into its own contract that both the new script storage contract and license storage contract inherit from. Should I go ahead and do that while addressing the previous point?
> > 
> > 
> > I think mostly same response as above, with added note that we have generally opted for libraries/helper functions over inheritance to help keep our contract source code a bit more straightforward when possible.
> 
> +++ re: libraries and protocol-oriented composability >>> inheritance patterns where possible, generally I think this feeds better "SOLID principle" conforming code.
> 
> If we think that refactoring to library-ize here is worthwhile, I'm supportive though don't feel strongly on the matter.

agree with y'alls points above and also personally like the idea of using an internal lib for common code across script/license onchain storage

lyaunzbe:
@ryley-o @sterlingcrispin I do think licenses changing for a dep pkg by an author/owner can definitely happen, is there a good reason to leave this out now for a future upgrade? Definitely won't die on this hill though, as ultimately creating a whole new dependency in that edge case is totally viable escape hatch :)

ryley-o:
> I do think licenses changing for a dep pkg by an author/owner can definitely happen, is there a good reason to leave this out now for a future upgrade? Definitely won't die on this hill though, as ultimately creating a whole new dependency in that edge case is totally viable escape hatch 😄

If this seems like it can happen, would very much rather add the (very cheap) ability to swap to a different license vs. the (very expensive) re-uploading of an entire dependency. Was thinking license would be static for a given dependency version, but I could be wrong @lyaunzbe !

lyaunzbe:
> > I do think licenses changing for a dep pkg by an author/owner can definitely happen, is there a good reason to leave this out now for a future upgrade? Definitely won't die on this hill though, as ultimately creating a whole new dependency in that edge case is totally viable escape hatch 😄
> 
> If this seems like it can happen, would very much rather add the (very cheap) ability to swap to a different license vs. the (very expensive) re-uploading of an entire dependency. Was thinking license would be static for a given dependency version, but I could be wrong @lyaunzbe !

Ahhh, sorry this was a lapse on my end, I forgot we were treating different versions of the same package as different dependencies in the mapping. I totally agree that most likely a change in a package's license would be coupled with a version bump 👍 👍

yoshiwarab:
Closing in favor of https://github.com/ArtBlocks/artblocks-contracts/pull/1215

PR #1022, opened 2023-09-10
PoC: Deploy from a gnosis safe

ryley-o:
## Description of the change

\<add description here\>


ryley-o:
retiring this PR based on new deployment strategies centered around the keyless create2 factory and proxy/Engine factory contracts, which accomplish a similar goal as this PoC aimed to achieve (with other security and scaling benefits)!

PR #1069, opened 2023-09-28
Review 1 [I-08] Emit events from libraries (SplitFundsLib)

ryley-o:
Required prior to Tranche 2 contract review.

## Description of the change

Update SplitFundsLib to define and emit events from within the library.

As discussed, due to the need to pass `projectId` and `coreContract` args to the library to populate events, this tips the library into a place where the benefits of using diamond storage outweigh the drawbacks. Thus, diamond storage is also implemented for the library. Additionally, some logic is migrated from the minter into the library, because the library is taking on a bit more responsibility (i.e. managing storage and events) and it was more clear to consolidate some logic (e.g. `preMintChecks`, etc.).


packages/contracts/contracts/libs/v0.8.x/minter-libs/SplitFundsLib.sol
```diff
@@ -159,25 +190,27 @@ library SplitFundsLib {
...
+                );
+                require(success, "Artist payment failed");
+            }
+            // additional payee payment
+            if (additionalPayeePrimaryRevenue_ > 0) {
+                (bool success, ) = additionalPayeePrimaryAddress_.call{
+                    value: additionalPayeePrimaryRevenue_
+                }("");
+                require(success, "Additional Payee payment failed");
+            }
```
lyaunzbe, line 213: super small nit for this function in general, other than sourcing the payments, much of this code is duplicate. would it make sense to refactor the shared parts into another func ( + conditional provider payment if engine === true)

lyaunzbe (approved):

packages/contracts/contracts/libs/v0.8.x/minter-libs/SplitFundsLib.sol
ryley-o: good call! I was battling stack too deep issues, but this was a smol disaster 😆 updated to use some helper functions and eliminate the redundant code in [792d0e6](https://github.com/ArtBlocks/artblocks-contracts/pull/1069/commits/792d0e684be26eebff2a394bdc7bf6750f91fc55)

PR #1091, opened 2023-10-05
Review 1 [M-01] Max Invocation & Revenue Withdrawal Restrictions

ryley-o:
## Description of the change

Add safeguards to help protect against malicious artist artificially inflating sellout price by adding restrictions on when max invocations can be edited, and when revenues can be withdrawn without admin consent.

>IMPORTANT: While these updates intend to provide additional safeguards to help prevent unilateral suspicious behavior by an artist, collectors should never purchase tokens on the Dutch Auction w/Settlement minter with the expectation that they will pay a lower net price than when they send their purchase transaction. Sellouts do happen, and sometimes within the span of only a few blocks.

This PR adds two additional safeguards to protect against accidental (or potentially malicious) artist behavior on the `MinterDAExpSettlementV3` minter.

### 1. Restrict changing minter-local max invocations after initial purchase on minter

Minter-local max invocations are now only changeable prior to the first purchase on the minter. This prevents an artist from accidentally reducing max invocations during a live auction, which could artificially inflate sellout prices (in turn reducing settlement funds available participants of the auction). This change is only implemented as a guard-rail for artists to help them not take any actions that could potentially harm their reputation.

### 2. Detect any "nonstandard activity" during auction, and require admin concurrence to withdraw revenues if detected

A simple mechanism that detects if any tokens were minted on a different minter, or if core contract max invocations were reduced, after the initial token purchase in an auction, is added. Put simply, the minter estimates the number of tokens to be auctioned during the first purchase during an auction, and then compares the actual number of purchases to the expected value during revenue withdrawal in the case of a sellout above auction base price. If the auction sold out, but did not sell the expected number of tokens, "nonstandard activity" has been detected, and `admin` must call the revenue withdrawal function instead of `artist`.

This separation of powers allows admin to intelligently determine the best course of action if "nonstandard activity" is detected. Potential results could range from admin simply determining that something accidental happened that didn't impact auction sellout price, to admin determining that the artist maliciously attempted to manipulate their auction's sellout price. Admin always has the ability to replace an artist's wallet with a different wallet under admin's control, and can collect revenues and distribute refunds as they see fit.

### Alternate Considerations

The issues highlighted in [M-01] were known prior to review, and we had originally opted to keep things simple and efficient by simply acknowledging that there are some trust assumptions in an artist. This is a fair assumption, as an artist is a privileged party with a reputation to defend. That being said, the [M-01] issue highlighted a pretty poor UX for both collectors and artists, and the solution developed in this PR seemed favorable and still relatively straightforward.

### Drawbacks

The following sequence of artist operations results in an unfortunate increase in admin overhead. Steps, in order:
- Artist configures auction on this minter (which automatically syncs minter max invocations to core contract)
- Artist reduces max invocations on core contract
- Artist does re-update max invocations on minter contract
- Auction begins and initial purchase is made

Because the minter is not aware of operations performed on core contract in the second bullet point above, the auction will be marked as "nonstandard activity" and require admin to call the withdraw revenues function. Because we can't know if the artist reducing core max invocations is malicious or intentionally intended to limit the project size, we need to defer to admin for concurrence. An alternate design option here would be to revert the initial purchase, but that seemed like an overall worse UX for this minter. It is recommended that the artist dashboard monitor for this type of situation, and artists are warned if their minter max invocations are > core contract max invocations.

### Gas Impacts

Struct packing was still available in the settlement library struct, so average mint prices are only increased 0.6% on the minter for the changes in this PR.

### Tests

Tests are updated to be passing, and new tests are added to cover the new lines of code.


ryley-o:
added a very minor update in [c360720](https://github.com/ArtBlocks/artblocks-contracts/pull/1091/commits/c3607202e55f67593b52f05dd564749864c18b1f) to improve handling of an edge case that I think might happen more than initially expected.

packages/contracts/contracts/libs/v0.8.x/minter-libs/DAExpLib.sol
```diff
@@ -36,7 +36,7 @@ library DAExpLib {
         // @dev max uint40 ~= 1.1e12 sec ~= 34 thousand years
         uint40 timestampStart;
         uint40 priceDecayHalfLifeSeconds;
-        // @dev max uint88 ~= 3e26 Wei = ~300 million ETH, which is well above
+        // @dev max uint88 ~= 3e26 Wei = ~30 million ETH, which is well above
```
lyaunzbe, line 39: wasn't 300 correct here? (3e26/10^18)

lyaunzbe (approved):
had one q about comments, otherwise lgtm

packages/contracts/contracts/libs/v0.8.x/minter-libs/SettlementExpLib.sol
```diff
@@ -339,8 +381,64 @@ library SettlementExpLib {
...
+            (
+                uint256 coreInvocations,
+                uint256 coreMaxInvocations
+            ) = MaxInvocationsLib.coreContractInvocationData(
+                    _projectId,
+                    _coreContract
+                );
+            // snap chalkline on the number of tokens to be auctioned on this
+            // minter
+            // @dev ackgnowledge that this value may be stale if the core
```
lindsgil, line 399: sp: acknowledge

lindsgil (approved):

packages/contracts/contracts/libs/v0.8.x/minter-libs/DAExpLib.sol
ryley-o: oops, fixed in [4a94e1d](https://github.com/ArtBlocks/artblocks-contracts/pull/1091/commits/4a94e1d7da92db22f47d666e2e71340e6a270d6d)

packages/contracts/contracts/libs/v0.8.x/minter-libs/SettlementExpLib.sol
ryley-o: good call, acknowledged and fixed in [3f96e49](https://github.com/ArtBlocks/artblocks-contracts/pull/1091/commits/3f96e4907fc321d93c96a3368c1e253d08970a01)

PR #1160, opened 2023-10-31
MS Review Final - Libs Review

ryley-o:
## Description of the change

Update contracts for final nits from libs review.

Incorporated all suggested changes except the following:

>// @audit-info Code improvement: I'd recommend to pack state with similar lifetime together (i.e adding an inner struct `AuctionFactorySettings` that regroup those 4 variables).

  * _I want to stay efficiently packed here and don’t think that it is possible to pack all 5 sub-32-byte variables if I use a struct. That being said, re-ordered things and used comments to more clearly designate the nice concept of `AuctionFactorySettings`_

>// @audit-info Code improvement: consider renaming to `getPrice` as the "unsafe" part is missleading as it depends on where the `maxHasBeenInvoked` comes from.

  * _Currently still prefer to put ‘unsafe’ in the name to make devs think twice before using it. acknowledge that the name is a bit misleading, but prefer devs make a Type I error vs. a Type II error in this area._

---
  - https://app.asana.com/0/0/1205846875012044


ryley-o:
merging all the final review items into a single PR, so closing this intermediate PR.

PR #1206, opened 2023-11-21
Use encrypted wallet keystores

ryley-o:
## Description of the change

Update hardhat config to load wallets from encrypted keystores.

This updates our hardhat config and scripting process to do the following:
- run hardhat deployment scripts from wallets loaded from encrypted keystores
- update env vars to define a single wallet encrypted keystore file instead of private keys
- remove mainnet/testnet designation from wallet env files, which has not been heavily utilized in our processes

A utility script to save a private key as an encrypted wallet is added for convenience.

---
  - https://app.asana.com/0/0/1205995396915359


packages/contracts/hardhat.config.ts
```diff
@@ -9,6 +9,27 @@ import "@nomicfoundation/hardhat-verify";
...
+import { getDeployerWallet } from "./scripts/util/get-deployer-wallet";
+
+// ----- WALLET CONFIGURATION -----
+// initialize with dummy fallback private key
+let PRIVATE_KEY: string =
+  "DEAD000000000000000000000000000000000000000000000000000000000000";
+// if process argument "run" is present, use the deployer wallet
+// @dev use arg "run" check to avoid wallet nuisance when hardhat
+// runs pre-processes such as "compile" prior to running scripts
+if (process.argv.includes("run")) {
```
yoshiwarab, line 21: 🚀

packages/contracts/scripts/util/get-deployer-wallet.ts
```diff
@@ -0,0 +1,60 @@
...
+var readlineSync = require("readline-sync");
+import { readFileSync } from "fs";
+import { Wallet, ethers } from "ethers";
+
+// Exports a function that returns the private key of the wallet to use for
+// deploying contracts. The private key is loaded from an encrypted keystore
+// file, defined by env var `WALLET_ENCRYPTED_KEYSTORE_FILE`. If this env var
+// is not set, null is returned, and a warning will be logged.
+// @dev use closure to only ever ask for password once
+export var getDeployerWallet = ((): (() => Wallet | null) => {
```
yoshiwarab, line 11: Nit: Any reason we're using var in this file?

yoshiwarab (approved):

packages/contracts/scripts/util/get-deployer-wallet.ts
ryley-o: good call - fixed in [8300810](https://github.com/ArtBlocks/artblocks-contracts/pull/1206/commits/830081088301ffd0698357dead7c8c98ea9bb6ec)

packages/contracts/hardhat.config.ts
ryley-o: I realized this had a smol bug, but fixed using similar trick in [83325e4](https://github.com/ArtBlocks/artblocks-contracts/pull/1206/commits/83325e40a8b492c8602516217c1e4c4e15eb1db5)

PR #1193, opened 2023-11-19
Update DependencyRegistryV0.sol

Aryan9592:
This PR solely aims to enhance the `for` loops by using `unchecked block` and changing `i++` to `++i`. 
For reference - https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#for-loops-improvement

Also got some typos fixed..especially the word `dependency`

Thanks @ryley-o

PR #1212, opened 2023-11-23
Update GenArt721CoreV3.sol

0xScratch:
## Description of the change

Same as #1193 

@ryley-o


packages/contracts/contracts/GenArt721CoreV3.sol
```diff
@@ -1176,6 +1176,10 @@ contract GenArt721CoreV3 is
                 }
             }
             revert("Improperly formatted aspect ratio");
+
+            unchecked {
```
ryley-o, line 1180: This introduces a bug because it is unreachable code - we use continue statements above.
I think sacrificing some gas efficiency for iterator incrementing readability is probably ideal here, so prefer the original pattern of incrementing the iterator in line 1161.

Aryan9592: hmm, let me check it out

ryley-o: fixed in [2a0930d](https://github.com/ArtBlocks/artblocks-contracts/pull/1212/commits/2a0930d3a8fe460a81b39e74681d7623b2ac0498)

ryley-o:
closing this due to many refactors, including retiring the GenArt721CoreV3.sol contract (replaced by Engine contracts)

PR #1215, opened 2023-11-28
Dependency registry naming updates

yoshiwarab:
## Description of the change

This PR updates the naming within DepencyRegistryV0 to be consistent with the new minter contract (i.e. no underscores for function/event param names). Additionally updates some of the names based on suggestions by Sterling.


packages/contracts/contracts/interfaces/v0.8.x/IDependencyRegistryV0.sol
```diff
@@ -3,96 +3,96 @@
...
-        uint256 _index
+        bytes32 dependencyNameAndVersion,
+        uint256 index
     ) external view returns (address);
 
     /**
-     * @notice Returns script for dependency type `_dependencyType` at script index `_index`.
-     * @param _dependencyType Dependency type to be queried.
-     * @param _index Index of script to be queried.
+     * @notice Returns script for dependency type `_dependencyNameAndVersion` at script index `_index`.
```
ryley-o, line 90:
```suggestion
     * @notice Returns script for dependency type `dependencyNameAndVersion` at script index `_index`.
```

packages/contracts/contracts/interfaces/v0.8.x/IDependencyRegistryV0.sol
```diff
@@ -3,96 +3,96 @@
...
      */
     function getDependencyScriptCount(
-        bytes32 _dependencyType
+        bytes32 dependencyNameAndVersion
     ) external view returns (uint256);
 
     /**
      * @notice Returns address with bytecode containing script for
-     * dependency type `_dependencyTypes` at script index `_index`.
+     * dependency type `_dependencyNameAndVersions` at script index `_index`.
```
ryley-o, line 82:
```suggestion
     * dependency type `dependencyNameAndVersions` at script index `_index`.
```

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -53,617 +53,662 @@ contract DependencyRegistryV0 is
...
     }
 
-    // dependency types, i.e. "type@version"
-    EnumerableSet.Bytes32Set private _dependencyTypes;
-    // mapping from dependencyTypes to Dependency, which stores the properties of each dependency
-    mapping(bytes32 => Dependency) dependencyDetails;
+    // dependency ID's are bytes32 of the format "name@version"
+    EnumerableSet.Bytes32Set private _dependencyNameVersionIds;
+    // mapping from dependencyNameAndVersion to Dependency, which stores the properties of each dependency
+    mapping(bytes32 => Dependency) dependencyRecords;
```
ryley-o, line 67: can used named descriptions now in mappings
```suggestion
    mapping(bytes32 dependencyNameAndVersion => Dependency) dependencyRecords;
```

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -53,617 +53,662 @@ contract DependencyRegistryV0 is
...
     EnumerableSet.AddressSet private _supportedCoreContracts;
-    mapping(address => mapping(uint256 => bytes32)) projectDependencyTypeOverrides;
 
-    function _onlyNonZeroAddress(address _address) internal pure {
-        require(_address != address(0), "Must input non-zero address");
+    // Mapping that allows for the overriding of project dependencies.
+    // The first key is the address of the core contract, the second key is the project ID,
+    // and the value is the bytes32 representation of the dependency name and version (i.e. name@version).
+    // This allows for specific projects to use different versions of dependencies than what's stored on the core contract.
+    mapping(address => mapping(uint256 => bytes32)) projectDependencyOverrides;
```
ryley-o, line 79: suggest named mappings
```suggestion
    mapping(address coreContract => mapping(uint256 projectId => bytes32 dependencyNameAndVersion)) projectDependencyOverrides;
```

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -53,617 +53,662 @@ contract DependencyRegistryV0 is
...
-        require(bytes(_string).length != 0, "Must input non-empty string");
+    function _onlyNonEmptyString(string memory string_) internal pure {
+        require(bytes(string_).length != 0, "Must input non-empty string");
     }
 
-    function _onlyAdminACL(bytes4 _selector) internal {
+    function _onlyAdminACL(bytes4 selector) internal {
         require(
-            adminACLAllowed(msg.sender, address(this), _selector),
+            adminACLAllowed(msg.sender, address(this), selector),
```
ryley-o, line 91: suggest using named args here, especially since there are two adjacent addresses

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -53,617 +53,662 @@ contract DependencyRegistryV0 is
...
             dependency.additionalCDNCount == 0 &&
                 dependency.additionalRepositoryCount == 0 &&
                 dependency.scriptCount == 0,
             "Cannot remove dependency with additional CDNs, repositories, or scripts"
         );
 
-        _dependencyTypes.remove(_dependencyType);
-        delete dependencyDetails[_dependencyType];
+        _dependencyNameVersionIds.remove(dependencyNameAndVersion);
+        delete dependencyRecords[dependencyNameAndVersion];
```
ryley-o, line 214: Want to double-click that this won't delete the mapping items in a `Dependency` struct. I think this is fine, since they are limited to `additionalCDNs`, `additionalRepositories`, and `scriptBytecodeAddresses `.
I definitely worth a dev comment in the function and in-line here though!
Alternatively, we could loop through these items and delete them, as this is a managed contract, and they should never get near a block gas limit to perform the looping operation.

ryley-o (approved):
LGTM - left a few additional styling suggestions

packages/contracts/contracts/DependencyRegistryV0.sol
yoshiwarab: @ryley-o we have the require above to check that those arrays are empty before this function can be called so I think it's okay as is. Let me know if you think I should remove the require and delete them in this function instead.

ryley-o: ohhhh nice! maybe an in-line dev comment right above this would be helpful to remind anyone reading!

yoshiwarab:
@ryley-o, I ended up implementing the storage abstraction pattern you recommended a while ago (https://art-blocks.slack.com/archives/C03R58ZGFBN/p1679953858606619) just want to make sure you take a look at that bit before I merge!

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -42,628 +44,643 @@ contract DependencyRegistryV0 is
 
     uint8 constant AT_CHARACTER_CODE = uint8(bytes1("@")); // 0x40
```
ryley-o, line 45: @yoshiwarab since this is a constant, I'm not sure if this ends up being an issue without using gap storage of diamond storage patterns in `DependencyRegistryV0Storage`.
That being said, I think the storage abstraction is a good thing overall here, but I think we would benefit from using a diamond storage pattern in a follow-on PR. So overall, this is okay to merge as-is as long as we do that follow-up work.

ryley-o (approved):

yoshiwarab:
@ryley-o, sorry one more check, I didn't realize Sterling had an open PR with the license text changes so I just pulled them over manually in [this commit](https://github.com/ArtBlocks/artblocks-contracts/pull/1215/commits/adcf72abf5c393ebc76fd2d5ccce066fc98cb096). Let me know if that looks okay!

ryley-o (approved):

packages/contracts/contracts/DependencyRegistryV0.sol
ryley-o: resolving due to #1224 follow-on 🙌

PR #1224, opened 2023-11-29
Update dependency registry to use the diamond storage pattern

yoshiwarab:


packages/contracts/contracts/DependencyRegistryStorageLib.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT
```
ryley-o, line 1:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-only
```

packages/contracts/contracts/DependencyRegistryStorageLib.sol
```diff
@@ -0,0 +1,59 @@
...
+pragma solidity ^0.8.19;
+
+import "./interfaces/v0.8.x/IAdminACLV0.sol";
+import "@openzeppelin-4.7/contracts/utils/structs/EnumerableSet.sol";
+
+library DependencyRegistryStorageLib {
```
ryley-o, line 7: nit: add natspec comments to library, structs, and functions, including internal functions in libraries. See `SetPriceLib.sol` for example natspec

packages/contracts/contracts/DependencyRegistryStorageLib.sol
```diff
@@ -0,0 +1,59 @@
...
+        // Mapping that allows for the overriding of project dependencies.
+        // The first key is the address of the core contract, the second key is the project ID,
+        // and the value is the bytes32 representation of the dependency name and version (i.e. name@version).
+        // This allows for specific projects to use different versions of dependencies than what's stored on the core contract.
+        mapping(address => mapping(uint256 => bytes32)) projectDependencyOverrides;
+    }
+
+    function s() internal pure returns (Storage storage storageStruct) {
+        bytes32 position = DIAMOND_STORAGE_POSITION;
+        assembly {
```
ryley-o, line 55:
```suggestion
        assembly ("memory-safe") {
```

packages/contracts/contracts/DependencyRegistryStorageLib.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.19;
+
```
ryley-o, line 3: Recommend putting this library file in `libs/v0.8.x/`

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -66,17 +65,23 @@ contract DependencyRegistryV0 is
...
     function _onlyExistingDependency(
         bytes32 dependencyNameAndVersion
     ) internal view {
+        DependencyRegistryStorageLib.Storage
+            storage ds = DependencyRegistryStorageLib.s();
+
         require(
-            _dependencyNameVersionIds.contains(dependencyNameAndVersion),
+            ds.dependencyNameVersionIds.contains(dependencyNameAndVersion),
             "Dependency does not exist"
```
ryley-o, line 85: In cases where you are only grabbing a single variable from the Storage struct, you could do the following. However, I think how you have done it probably results in a bit more readable code, and is good as-is. Curious what your opinion is.
```
// can accomplish the check without defining an intermediate storage pointer `ds`
DependencyRegistryStorageLib.s().supportedCoreContracts.contains(
                coreContractAddress
            )
```

packages/contracts/contracts/DependencyRegistryStorageLib.sol
```diff
@@ -0,0 +1,59 @@
...
+import "./interfaces/v0.8.x/IAdminACLV0.sol";
+import "@openzeppelin-4.7/contracts/utils/structs/EnumerableSet.sol";
+
+library DependencyRegistryStorageLib {
+    bytes32 constant DIAMOND_STORAGE_POSITION =
+        keccak256("dependencyregistrystoragelib.storage");
+
+    struct License {
+        // mapping from license index to address storing script in bytecode
+        mapping(uint256 => address) licenseBytecodeAddresses;
```
ryley-o, line 13: recommend using names on all mappings, such as:
```suggestion
        mapping(uint256 licenseIndex => address bytecodeAddress) licenseBytecodeAddresses;
```

ryley-o (approved):
This is great! While it does add more lines of code, I think use of diamond storage in the upgradeable registry contract cleans up a lot of implicit assumptions and removes a lot of potential footguns!

packages/contracts/contracts/DependencyRegistryV0.sol
yoshiwarab: Yeah, I don't think the intermediate storage pointer makes things meaningfully more readable, updated!

ryley-o (approved):

PR #1250, opened 2023-12-08
Deploy V2 cores for BMG on Sepolia

ryley-o:
## Description of the change

Deploy V2 cores for BMG on Sepolia.

Note: I messed up randomizer - both contracts need to be updated to use the randomizer at `Sepolia:0xb4CBEE71aA18f28fdD49837e0a1038935F54931a` 

Note: manually ran a sync-from for these contracts, since V2 deployments are not automatically picked up via a registry event.


ryley-o:
@yoshiwarab I force-merged this to enable another V2 deployment that I have to do today 🙏

PR #1267, opened 2023-12-18
Update MinterERC20SetPrice with Price and Currency

lindsgil:
## Description of the change

Update MinterSetPriceERC20V0, MinterSetPriceERC20V1, and MinterSetPriceERC20V4 purchase and purchaseTo functions with price and currency address

### DIFF
--- a/packages/contracts/contracts/minter-suite/Minters/MinterSetPriceERC20V4.sol
+++ b/packages/contracts/contracts/minter-suite/Minters/MinterSetPriceERC20V4.sol
@@ -328,7 +328,7 @@ contract MinterSetPriceERC20V4 is
         uint256 _projectId
     ) external payable returns (uint256 tokenId) {
         // pass max price as msg.value, currency address as ETH
-        tokenId = purchaseTo_do6({
+        tokenId = purchaseTo({
             _to: msg.sender,
             _projectId: _projectId,
             _maxPricePerToken: msg.value,
@@ -349,7 +349,7 @@ contract MinterSetPriceERC20V4 is
         uint256 _maxPricePerToken,
         address _currencyAddress
     ) external payable returns (uint256 tokenId) {
-        tokenId = purchaseTo_do6({
+        tokenId = purchaseTo({
             _to: msg.sender,
             _projectId: _projectId,
             _maxPricePerToken: _maxPricePerToken,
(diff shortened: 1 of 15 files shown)


packages/contracts/contracts/minter-suite/Minters/MinterSetPriceERC20V4.sol
```diff
@@ -337,12 +364,35 @@ contract MinterSetPriceERC20V4 is
...
+            _currencyAddress: address(0)
+        });
         return tokenId;
     }
 
     /**
-     * @notice Purchases a token from project `_projectId` and sets
+     * @notice gas-optimized version of purchase(uint256,uint256,address).
+     */
+    function purchase_H4M(
```
ryley-o, line 380: nit: this would be `purchase_a1A`, but I would be fine just removing the gas-optimized version in this update.
ref: https://emn178.github.io/solidity-optimize-name/

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceERC20V4.sol
```diff
@@ -361,6 +443,25 @@ contract MinterSetPriceERC20V4 is
     function purchaseTo_do6(
         address _to,
         uint256 _projectId
```
ryley-o, line 445: would be fine removing this gas-optimized version in this update

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceERC20V4.sol
```diff
@@ -361,6 +443,25 @@ contract MinterSetPriceERC20V4 is
...
+                _projectId: _projectId,
+                _maxPricePerToken: msg.value,
+                _currencyAddress: address(0)
+            });
+    }
+
+    /**
+     * @notice gas-optimized version of purchaseTo(address, uint256, uint256, address).
+     */
+    function purchaseTo_do6(
```
ryley-o, line 460: I think can just make this `purchaseTo_bjU` to update the gas-optimized name for the new function signature.
Could also just remove all the gas-optimized function names if you want.
reference: https://emn178.github.io/solidity-optimize-name/

ryley-o (approved):
LGTM!

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceERC20V4.sol
lindsgil: Ah, thanks for sharing! Removed this 🙏

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceERC20V4.sol
lindsgil: Updated!

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceERC20V4.sol
lindsgil: Updated! 🙏

PR #1268, opened 2023-12-18
Add SplitAtomicV0 and EIP1167 clone factory

ryley-o:
## Description of the change

Add a new SplitAtomicV0 contract and EIP-1167 clone factory to enable proxy-pattern split contracts.

The SplitAtomicV0 contract behaves in the following way:
- Splits native token (ETH) at the time that it is received
- Has `drainETH` function available to drain ETH received via `SELFDESTRUCT` (or `SENDALL`) opcodes
- Has `drainERC20` function available to drain ERC20 tokens received (since we don't want to rely on a safe-transfer pattern calling any hooks).

## Factory Design
The factory is designed to support one immutable required split (address and BPS). This helps with security around eventual implementations for certain use-cases. Since the factory contract does maintain state, it is made "abandonable" if it becomes no longer an appropriate atomic split creation mechanism.

## Proxy Design

An EIP-1167 is a reasonable choice for this contract, because many splitter contracts are expected to be created.  Using a minimal proxy pattern greatly reduces the gas costs associated with defining a new splitter contract, while only adding a minimal amount of gas (due to additional DELEGATECALL) during split operations.

Note that EIP-1167 proxy patterns are non-upgradeable, which is appropriate for this use case, since we desire an immutable splitter contract.

Other proxy/non-proxy designs were studied in other PRs to determine the deployment cost vs. split gas cost trade-space. Split gas costs can be reduced by ~10k gas by storing the splits as immutable variables because it removes many SLOADs during splitting, but deployment costs of new split contracts get very expensive for artists. Overall, the EIP-1167 solution is recommended, and the alternate designs that were studied are described in the table below:

| Design Type | Artist gas used for new split creation | Gas used for 3-way split execution | Comments | PR |
| --- | --- | --- | --- | -- |
| EIP-1167 Minimal Proxy | 187321 | 43572 | Minimizes artist deployment cost, reasonable split execution cost | (this PR) |
| No Proxy, Immutable Vars | 699899 (could be $100's) | 31551 | ~maximizes deployment cost, minimizes split execution cost | #1290 |
| DELEGATECALL to helper functions | 580532 | 55533 | Reduces deployment cost, poor overall solution | #1291 |

## Splitter Design

This implementation atomically splits the funds upon receipt of ETH. This adds gas costs during payment operations, however, keeps bookkeeping consistent with current behaviors in our system. Additionally, for auction mechanics such as DA w/Settlement, this doesn't add more gas costs to purchases made during the auction, since the minter holds funds in those cases.

Future iterations of this could be gas-optimized to potentially remove the atomic-split functionality, and always require calls to the drain functions (becoming more similar to a 0xsplits contract). This would add draining overhead and add other process updates.

The splitter factory was made "abandonable" such that the deployer can stop all new splitter contract creations. This is useful in situations where old AB payment addresses are desired to be retired.

## Tests

- [x] Tests are added for full coverage of new contracts
  - exceptions are redundant (fault-tolerant) reentrancy checks, and reentrancy checks on ERC20 tokens (due to heavy test lift on an already fault-tolerant design)

## Other
- bumped solidity compiler to latest supported version by Hardhat, which enables use of OpenZeppelin 5.0 contracts

### References
- good EIP-1167 reference: https://medium.com/taipei-ethereum-meetup/reason-why-you-should-use-eip1167-proxy-contract-with-tutorial-cbb776d98e53

---
  - https://app.asana.com/0/0/1206266912756248


packages/contracts/contracts/split/SplitAtomicV0.sol
```diff
@@ -0,0 +1,254 @@
...
+
+    /**
+     * @notice Drains the contract's balance to the configured `splits`.
+     * Reverts if not initialized.
+     * @dev This function is useful for draining the contract's balance to the
+     * configured `splits` in the event that the contract receives funds via
+     * a force-send (e.g. `SELFDESTRUCT` or `SENDALL`) operation.
+     * @dev This function relies on being non-reentrant for security.
+     */
+    function drainETH() external nonReentrant {
```
lindsgil, line 114: Should we emit an event here and `drainERC20`?

ryley-o: was considering that, but I can't see how it would affect our indexing, so was thinking less gas is better here? idk, what do you think?

lindsgil: that makes sense 🙏 was thinking it might be helpful if we want to track/monitor those calls but that might just be noise?

yoshiwarab (approved):

lindsgil (approved):

lyaunzbe (approved):

packages/contracts/contracts/interfaces/v0.8.x/ISplitAtomicFactoryV0.sol
```diff
@@ -0,0 +1,44 @@
...
+     * @notice Initializes the contract with the provided `splits`.
+     * Only callable once.
+     * @param splits Splits to configure the contract with. Must add up to
+     * 10_000 BPS.
+     * @return splitAtomic The address of the newly created split atomic
+     * contract
+     */
+    function createSplit(
+        Split[] calldata splits
+    ) external returns (address splitAtomic);
```
ryley-o, line 31: TODO - add a 22.22% constraint to AB Admin here

ryley-o, line 31:
```suggestion
    ) external returns (address splitAtomic, uint256 intendedProjectIdMetadata);
    // emit event connecting the new split and intended project ID
```

ryley-o: @Jbern16 @lindsgil I think we ended up handling this with frontend solution (localstorage), which I like much more since this is a permissionless call and subject to spamming. going to go ahead and close this one.

ryley-o: added in [a2c9c2e](https://github.com/ArtBlocks/artblocks-contracts/pull/1268/commits/a2c9c2e556520326d4f329784f6e2845ad7d8951)

packages/contracts/contracts/split/SplitAtomicV0.sol
ryley-o: added here: [1049411](https://github.com/ArtBlocks/artblocks-contracts/pull/1268/commits/10494119e20b21b9d6b7fe689d1f45f762c417f6)

packages/contracts/test/util/fixtures.ts
```diff
@@ -49,3 +55,33 @@ export async function setupConfigWitMinterFilterV2Suite() {
...
+    { recipient: config.accounts.deployer.address, basisPoints: 2222 },
+    { recipient: config.accounts.artist.address, basisPoints: 2778 },
+    { recipient: config.accounts.additional.address, basisPoints: 5000 },
+  ]);
+  const receipt = await ethers.provider.getTransactionReceipt(tx.hash);
+  // get splitter address from logs
+  const splitterCreationLog = receipt.logs[receipt.logs.length - 1];
+  const splitterAddress = ethers.utils.getAddress(
+    "0x" + splitterCreationLog.topics[1].slice(-40)
+  );
```
ryley-o, line 84: @Jbern16 here is an example of how to create a new splitter contract, as well as get the newly created address from the emitted event in the transaction

ryley-o:
merging after discussion yesterday, and ~24 hours of additional review time.

PR #1320, opened 2024-01-12
Create GenArt721GeneratorV0

yoshiwarab:
## Description of the change

This PR adds the GenArt721GeneratorV0 contract which utilizes scripty.sol to generate the full HTML for a given token by stitching together data from a dependency registry and core contract. 

Also created a little demo repo showing the full html pulled from the contract to display an AB token [here](https://github.com/yoshiwarab/ab-gen-on-chain).

https://sepolia.etherscan.io/address/0xdC862938cA0a2D8dcabe5733C23e54ac7aAFFF27#code


packages/contracts/contracts/generator/scripty-support/ETHFSFileStorage.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: MIT
```
yoshiwarab, line 1: @ryley-o There's a new version of ethFS (currently the only version on sepolia) which uses a different File.sol/IFile.sol so we needed to deploy this ourselves. That's also why I've added the ethfs directory here. For prod we shouldn't need to deploy anything ourselves for scripty/ethfs so we may want to remove this.

yoshiwarab:
@ryley-o I didn't make any of the address updatable on the contract because I figured it's light enough to just redeploy but let me know what you think!

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+        uint256 offset;
+        if (storageVersion == BytecodeStorageReader.V0_VERSION_STRING) {
+            offset = 104;
+        } else if (storageVersion == BytecodeStorageReader.V1_VERSION_STRING) {
+            offset = 65;
+        } else {
+            revert("Unsupported storage version");
+        }
+
+        return AddressChunks.mergeChunks(scriptBytecodeAddresses, offset);
```
ryley-o, line 105: I think the reason we get the data this way instead of using `BytecodeStorageReader.readFromBytecode` and copying strings around in memory is for gas efficiency, right? I think probably worth a small dev comment about why this implementation was selected.

packages/contracts/contracts/interfaces/v0.8.x/IGenArt721CoreProjectScriptV1.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.0;
+
+/**
+ * @dev This interface is implemented by GenArt721CoreV3 and above
```
ryley-o, line 5: WDYT about adding this to the V3 core contracts to enforce this with compiler?

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+     * supported by the DependencyRegistry or if the contract at the address
+     * does not implement the IGenArt721CoreProjectScriptV0 or IGenArt721CoreProjectScriptV1.
+     */
+    function getProjectScript(
+        address coreContractAddress,
+        uint256 projectId
+    ) external view returns (bytes memory) {
+        _onlySupportedCoreContract(coreContractAddress);
+
+        try
```
ryley-o, line 125: nit: maybe add dev comment indicating something like "try V3 Core interface first"?
could also add another dev comment before the subsequent try block as well.

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+            if (scriptCount == 0) {
+                return "";
+            }
+
+            string memory script;
+            for (uint256 i = 0; i < scriptCount; i++) {
+                string memory scriptChunk = IGenArt721CoreProjectScriptV0(
+                    coreContractAddress
+                ).projectScriptByIndex(projectId, i);
+                script = string.concat(script, scriptChunk);
```
ryley-o, line 175: I could see this potentially using a lot of gas? which is overall fine for reads of course, but I'm guessing we could improve this with some assembly code in future follow-on activity 👍

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+            }
+
+            return abi.encodePacked(script);
```
ryley-o, line 178: can you remind me why we use abi.encodePacked here?

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+     * @param tokenId The ID of the token to retrieve the HTMLRequest for.
+     * @return The HTMLRequest for the token.
+     */
+    function getTokenHtmlRequest(
+        address coreContractAddress,
+        uint256 tokenId
+    ) internal view returns (HTMLRequest memory) {
+        _onlySupportedCoreContract(coreContractAddress);
+
+        uint256 projectId = tokenId / ONE_MILLION;
```
ryley-o, line 196: nit: can use ABHelpers library for this

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+        // This will revert for older contracts that do not have an override set.
+        bytes32 dependencyNameAndVersion = dependencyRegistry
+            .getDependencyNameAndVersionForProject(
+                coreContractAddress,
+                projectId
+            )
+            .stringToBytes32();
+
+        bytes32 tokenHash;
+        if (tokenHash == bytes32(0)) {
```
ryley-o, line 206: nit: don't need if statement in initial block (always true)

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+            try
```
ryley-o, line 207: nit: suggest adding some dev comments about which core versions are being tried along the way here

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+            } catch {
+                revert("Unable to retrieve token hash.");
+            }
+        }
+
+        HTMLTag[] memory headTags = new HTMLTag[](2);
+        headTags[0].tagOpen = "<style>";
+        headTags[0]
+            .tagContent = "html{height:100%}body{min-height:100%;margin:0;padding:0}canvas{padding:0;margin:auto;display:block;position:absolute;top:0;bottom:0;left:0;right:0}";
+        headTags[0].tagClose = "</style>";
```
ryley-o, line 232: nit: headTag 0 could come from a pure helper function if you want

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+        if (scriptCount == 0) {
+            bodyTags[0].tagOpen = abi.encodePacked(
+                '<script type="text/javascript" src="',
+                preferredCDN,
+                '">'
+            );
+            bodyTags[0].tagContent = "// Noop"; // ScriptyBuilder requires scriptContent for this to work
+            bodyTags[0].tagClose = "</script>";
+        } else {
+            bytes memory dependencyScript = this.getDependencyScript(
```
ryley-o, line 268: any reason to not make `getDependencyScript` public and remove `this.` here? should be cheaper to make call internal.

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+                dependencyNameAndVersion
+            );
+            bodyTags[0].tagContent = dependencyScript;
+            bodyTags[0].tagType = HTMLTagType.scriptGZIPBase64DataURI; // <script type="text/javascript+gzip" src="data:text/javascript;base64,[script]"></script>
+        }
+
+        bodyTags[1].name = "gunzipScripts-0.0.1.js";
```
ryley-o, line 275: wdyt about a dev comment explaining why gzip scripts are included?

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,331 @@
...
+        bodyTags[1].tagType = HTMLTagType.scriptBase64DataURI; // <script src="data:text/javascript;base64,[script]"></script>
+        bodyTags[1].contractAddress = ethFS;
+
+        bytes memory projectScript = this.getProjectScript(
```
ryley-o, line 279: same comment about external vs. public to remove `this.`

packages/contracts/contracts/generator/scripty-support/ETHFSFileStorage.sol
ryley-o: SGTM - Is this a forked version of scripty? Wasn't quite sure why this file has to exist?
A little readme in here and ethfs directories that points to the source repos to give credit probably makes sense. IIUC, we cannot just import ethfs, because there are no official release of it?
Also, IIUC, scripty is relying on ethfs's FileStore interface, which is owned by an EOA on mainnet. I think it may be worth noting that dependency/risk in the natspec of GenArt721GeneratorV0.sol, yeah? (I think it is an acceptable risk for now, fwiw, just want to document it).

packages/contracts/contracts/libs/v0.8.x/Bytes32Strings.sol
```diff
@@ -63,4 +63,23 @@ library Bytes32Strings {
...
+     */
+    function stringToBytes32(
+        string memory source
+    ) internal pure returns (bytes32 result) {
+        bytes memory tempEmptyStringTest = bytes(source);
+        if (tempEmptyStringTest.length == 0) {
+            return 0x0;
+        }
+
+        assembly {
```
ryley-o, line 81:
```suggestion
        assembly ("memory-safe") {
```

packages/contracts/contracts/libs/v0.8.x/Bytes32Strings.sol
```diff
@@ -63,4 +63,23 @@ library Bytes32Strings {
         }
         return _occurrences == targetQty;
     }
+
+    /**
+     * @notice Converts string to bytes32.
+     * https://ethereum.stackexchange.com/a/9152
+     * @param source string to be converted.
+     * @return result bytes32 representation of string.
```
ryley-o, line 71: I think worth adding a note that strings whose representations are >32 bytes will be truncated. Additionally, I'm not entirely sure what would happen with a string with a multi-byte character that spans across the end of the 32'nd byte, might throw in that case?

packages/contracts/contracts/mock/MockFileStore.sol
```diff
@@ -0,0 +1,22 @@
...
+
+import {IContractScript} from "scripty.sol/contracts/scripty/interfaces/IContractScript.sol";
+
+contract MockFileStore is IContractScript {
+    string constant GUNZIP_SCRIPT_NAME = "gunzipScripts-0.0.1.js";
+
+    function getScript(
+        string calldata name,
+        bytes memory
+    ) external view returns (bytes memory script) {
```
ryley-o, line 12:
```suggestion
    ) external pure returns (bytes memory script) {
```

packages/contracts/contracts/generator/scripty-support/ETHFSFileStorage.sol
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * @notice Get the full script from ethfs's FileStore contract
+     * @param name - Name given to the script. Eg: threejs.min.js_r148
+     * @param data - Arbitrary data. Not used by this contract.
+     * @return script - Full script from merged chunks
+     */
+    function getScript(
+        string calldata name,
+        bytes memory data
```
ryley-o, line 35: unused function parameter - throws compiler warning, could comment out the name like this + update natspec
```suggestion
        bytes memory /*data*/
```

ryley-o (approved):
LGTM! Mostly nits, as well as I do wonder if it would just be worth not all of the complexity to integrate with ethfs and just upload the gzip script via bytecode storage ourselves?

ryley-o:
>@ryley-o I didn't make any of the address updatable on the contract because I figured it's light enough to just redeploy but let me know what you think!

Hmm, yes, I think this is a design decision we can make. In general, I think if we don't mind our on-chain generator address changing over time, the current implementation is best. For example, seems more likely that we would update this for a new BytecodeStorage version before having to change the addresses.

I do think we should consider making this contract an upgradeable proxy so we have a static on-chain generator address (like we do with dependency registry). @yoshiwarab what are your thoughts on that?

packages/contracts/contracts/generator/scripty-support/ETHFSFileStorage.sol
yoshiwarab: This is pretty much just for the gunzip script. Scripty allows you to pass in a contract address and name for script tags like [this](https://int-art.gitbook.io/scripty.sol-v2/building-html/getting-html-on-chain#generating-script-tag-with-its-content-fetched-from-ethfs) but the address needs to implement [IContractScript](https://github.com/intartnft/scripty.sol/blob/main/contracts/scripty/interfaces/IContractScript.sol). They have an IContractScript [wrapper](https://github.com/intartnft/scripty.sol/blob/main/contracts/scripty/externalStorage/ethfs/ETHFSFileStorage.sol) around the ethfs FileStore contract so that it's possible to use in this way but the wrapper they have uses an older version of ethfs' File and IFileStore which is not compatible with the new version of ethfs (the only one deployed to sepolia). This file is the same as the scripty ETHFSFileStorage but imports the updated ethfs files for compatibility with the new version.

yoshiwarab: Also, ethfs isn't available as an npm package so we need the supporting files manually added to our repo.

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
yoshiwarab: I don't think we need to, just need to cast to bytes, will update!

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
yoshiwarab: Yeah, will optimize in a follow on!

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
yoshiwarab: Mainly I thought it might be nice to be able to get the full dependency script on its own easily but I guess we're returning bytes so it wouldn't be immediately human readable anyway. Will go ahead and make internal and we can always add another public view function in the future if we want.

yoshiwarab: I'm actually gonna rename these to getDependencyScriptBytes and getProjectScriptBytes and then add external views to easily grab the full scripts without concatenating on-chain.

packages/contracts/contracts/libs/v0.8.x/Bytes32Strings.sol
yoshiwarab: I don't think it would throw, I'd imagine it would still just return truncated bytes that would be parsed to a garbled string

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
ryley-o: cool - yeah and just to double check, was suggesting to change from `external` to `public`, so we have both external function + ability to call as internal as well.

packages/contracts/contracts/libs/v0.8.x/Bytes32Strings.sol
ryley-o: might be good to test this out and determine if solidity throws an "invalid continuation byte" or something. just want to make sure we document any revert behavior to prevent any denial of service bugs if this were to be used on any future contracts.

ryley-o: I do think most likely would be a garbled string though 🤣

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
yoshiwarab: Ah that makes sense! I was sort of conflating external and public so I thought you were suggesting to make it internal. I think I may keep the update anyway so we have a function where input and output are both human readable.

yoshiwarab:
@ryley-o Ripped out the ETHFs stuff in favor of a self deployed storage contract containing the gunzip script. Also made the contract upgradable and added write functions for the relevant state. Please give it another look over when you get the chance!

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,436 @@
...
+                .getDependencyScriptBytecodeAddress(
+                    dependencyNameAndVersion,
+                    i
+                );
+        }
+
+        bytes32 storageVersion = BytecodeStorageReader
+            .getLibraryVersionForBytecode(scriptBytecodeAddresses[0]);
+
+        uint256 offset;
```
ryley-o, line 147: nit: might be worth a dev comment about why offset is needed here

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,436 @@
...
+     * @param coreContractAddress The address of the core contract to retrieve
+     * the project script from.
+     * @param projectId The ID of the project to retrieve the script for.
+     * @return The project script stiched together from all of the script chunks.
+     * @dev This function will revert if the core contract address is not
+     * supported by the DependencyRegistry or if the contract at the address
+     * does not implement the IGenArt721CoreProjectScriptV0 or IGenArt721CoreProjectScriptV1.
+     */
+    function getProjectScriptBytes(
+        address coreContractAddress,
```
ryley-o, line 192: super minor nit: could name this `coreContract`, as address is implied by the type (applies to the whole contract). This could also be done in a future upgrade to expedite shipping this (or ignored enirely!)

ryley-o (approved):
LGTM - really like this setup, feels cleaner + more chain-agnostic than pulling in ethfs

0xdude:
@yoshiwarab this looks really great and clean!

Just heads up. I released a small update for scripty.sol. Contracts are already deployed but changes are not merged to main repo yet. You can find the updated branch here:
https://github.com/intartnft/scripty.sol/tree/scripty-v3

Changes affecting `GenArt721GeneratorV0` is minimal:
- scripty. sol now supports EthFS V2. I know you already ripped out EthFS but wanted to share. Y
- The small issue in `L386` in `GenArt721GeneratorV0` now should be fixed. Scripty fetches content from other contracts only if the `contractAddress` is set. Otherwise, it uses `tagOpen`, `tagClose` and `tagContent`.
- Renamed `IContractScript.sol` to `IScriptyContractStorage`. Method signature is updated from `getScript` to `getContent`.

Also, I'm not sure if you test it on other chains but I deployed scripty.sol to multiple chains. Please let me know

yoshiwarab:
@ryley-o contract + test diffs since your last review:

contract: https://www.diffchecker.com/eVywtMrl/
test: https://www.diffchecker.com/J7XmVGUf/

packages/contracts/scripts/one-off/deploy-dev-generator.ts
```diff
@@ -0,0 +1,128 @@
...
...MxfHxyWzFdIT0xMzl8fHJbMl0hPTgpJiZ4KDYsImludmFsaWQgZ3ppcCBkYXRhIik7dmFyIG49clszXSx0PTEwO24mNCYmKHQrPXJbMTBdfChyWzExXTw8OCkrMik7Zm9yKHZhciBlPShuPj4zJjEpKyhuPj40JjEpO2U+MDtlLT0hclt0KytdKTtyZXR1cm4gdCsobiYyKX0sTXI9ZnVuY3Rpb24ocil7dmFyIG49ci5sZW5ndGg7cmV0dXJuKHJbbi00XXxyW24tM108PDh8cltuLTJdPDwxNnxyW24tMV08PDI0KT4+PjB9O2Z1bmN0aW9uIEgocixuKXtyZXR1cm4genIoci5zdWJhcnJheShTcihyKSwtOCksbnx8bmV3IHYoTXIocikpKX12YXIgVHI9dHlwZW9mIFRleHREZWNvZGVyPCJ1IiYmbmV3IFRleHREZWNvZGVyLENyPTA7dHJ5e1RyLmRlY29kZShBcix7c3RyZWFtOiEwfSksQ3I9MX1jYXRjaHt9dmFyIGZyPSgpPT57dmFyIG47bGV0IHI9ZG9jdW1lbnQucXVlcnlTZWxlY3RvckFsbCgnc2NyaXB0W3R5cGU9InRleHQvamF2YXNjcmlwdCtnemlwIl1bc3JjXScpO2ZvcihsZXQgdCBvZiByKXRyeXtsZXQgZT10LnNyYy5tYXRjaCgvXmRhdGE6KC4qPykoPzo7KGJhc2U2NCkpPywoLiopJC8pO2lmKCFlKWNvbnRpbnVlO2xldFthLHUsZyxzXT1lLGk9VWludDhBcnJheS5mcm9tKGc/YXRvYihzKTpkZWNvZGVVUklDb21wb25lbnQocyksdz0+dy5jaGFyQ29kZUF0KDApKSxoPW5ldyBUZXh0RGVjb2RlcigpLmRlY29kZShIKGkpKSxsPWRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoInNjcmlwdCIpO2wudGV4dENvbnRlbnQ9aCwobj10LnBhcmVudE5vZGUpPT1udWxsfHxuLnJlcGxhY2VDaGlsZChsLHQpfWNhdGNoKGUpe2NvbnNvbGUuZXJyb3IoIkNvdWxkIG5vdCBndW56aXAgc2NyaXB0Iix0LGUpfX07ZnIoKTt3aW5kb3cuZ3VuemlwU3luYz1IO3dpbmRvdy5ndW56aXBTY3JpcHRzPWZyO30pKCk7";
```
ryley-o, line 17: we have this a few places in ts files - wdyt about moving to util/common.ts?

ryley-o (approved):
LGTM - I'll follow up re: script compression comment in AddressChunks.sol 😃

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-only
-pragma solidity 0.8.19;
+pragma solidity 0.8.22;
```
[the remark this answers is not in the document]
yoshiwarab: Ah this was just so that I could import it in GenArt721Generator. Do you think I should break out an interface instead?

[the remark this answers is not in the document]
ryley-o: oh yeah, I would probably just import IDependencyRegistryV0 in the generator contract, since the interface shouldn't be version locked. any functions that are needed can just be added to the interface.

packages/contracts/test/generator/GenArt721GeneratorV0.test.ts
```diff
@@ -0,0 +1,976 @@
...
+      // Project script
+      expect(tokenHtml).to.include(getScriptTag(projectScript));
+
+      // Base64 encoded data uri
+      expect(encodedTokenHtml).to.equal(
+        `data:text/html;base64,${Buffer.from(tokenHtml).toString("base64")}`
+      );
+    });
+
+    it("gets html for a given V3 core contract token with dependency on chain", async function () {
```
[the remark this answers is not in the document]
ryley-o: fixed in [5f93c89](https://github.com/ArtBlocks/artblocks-contracts/pull/1320/commits/5f93c8989b20962dc188116c21cdcf566082be45)

packages/contracts/contracts/generator/AddressChunks.sol
```diff
@@ -0,0 +1,47 @@
+// Forked from https://github.com/intartnft/scripty.sol/blob/28cc612d7dc3a709f35c534c981bfc6bbfce4209/contracts/scripty/utils/AddressChunks.sol
+// with adjustment to data offset for compatibility with Art Blocks BytecodeStorage.sol
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
+library AddressChunks {
+    function mergeChunks(
```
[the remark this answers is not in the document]
ryley-o: resolved in [5f93c89](https://github.com/ArtBlocks/artblocks-contracts/pull/1320/commits/5f93c8989b20962dc188116c21cdcf566082be45)

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,561 @@
...
+                storageVersion == BytecodeStorageReader.V1_VERSION_STRING
+            ) {
+                offset = 65;
+            } else {
+                revert("Unsupported storage version");
+            }
+
+            // @dev We concatenate the script chunks together with assembly
+            // in AddressChunks for gas efficiency.
+            return AddressChunks.mergeChunks(scriptBytecodeAddresses, offset);
```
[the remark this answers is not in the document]
ryley-o: resolved in [5f93c89](https://github.com/ArtBlocks/artblocks-contracts/pull/1320/commits/5f93c8989b20962dc188116c21cdcf566082be45)

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,561 @@
...
+        // to know the offset of the script content in the bytecode. This is
+        // different for each storage version.
+        uint256 offset;
+        if (storageVersion == BytecodeStorageReader.V0_VERSION_STRING) {
+            offset = 104;
+        } else if (storageVersion == BytecodeStorageReader.V1_VERSION_STRING) {
+            offset = 65;
+        } else {
+            revert("Unsupported storage version");
+        }
```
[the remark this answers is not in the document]
ryley-o: fixed in [5f93c89](https://github.com/ArtBlocks/artblocks-contracts/pull/1320/commits/5f93c8989b20962dc188116c21cdcf566082be45)

packages/contracts/contracts/DependencyRegistryV0.sol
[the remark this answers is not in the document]
ryley-o: resolved in [2254aea](https://github.com/ArtBlocks/artblocks-contracts/pull/1320/commits/2254aea5bc33ed01e3cd604c076356eec8d92ff0)

packages/contracts/contracts/generator/GenArt721GeneratorV0.sol
```diff
@@ -0,0 +1,561 @@
...
+    // storage when upgrading.
+    DependencyRegistryV0 public dependencyRegistry;
+    IScriptyBuilderV2 public scriptyBuilder;
+    address public gunzipScriptBytecodeAddress;
+
+    event DependencyRegistryUpdated(address indexed _dependencyRegistry);
+    event ScriptyBuilderUpdated(address indexed _scriptyBuilder);
+    event GunzipScriptBytecodeAddressUpdated(
+        address indexed _gunzipScriptBytecodeAddress
+    );
```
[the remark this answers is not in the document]
ryley-o: resolved in [2254aea](https://github.com/ArtBlocks/artblocks-contracts/pull/1320/commits/2254aea5bc33ed01e3cd604c076356eec8d92ff0)

PR #1324, opened 2024-01-12
study min heap RAM minter

ryley-o:
## Description of the change

Study min heap implementation of RAM minter vs. bucket sort.

This will likely be closed without merging (as both implementations are study-quality, not production quality), but the branch may be used as a reference for study performed on the minter.


ryley-o:
This was reviewed with the team today and it was decided to move forward with the bucket-sort minter, primarily due to the much better gas performance. The study is currently documented internally, and may be documented in a more externally-facing manner in the future.

PR #1343, opened 2024-01-23
Add prototype min heap lib to repo

ryley-o:
## Description of the change

Add a prototype min heap library to this repo.

The library was developed while studying different RAM minter implementations, but figured it made sense to capture the library as an untested public good, as well as something we could potentially pick up and use in the future 🙏


ryley-o:
since this is untested, probably shouldn't commit to our repo. will close this PR, but keep for historical documentation purposes.

PR #1347, opened 2024-01-25
Prototype new many-superAdmin AdminACLV2

ryley-o:
## Description of the change

Prototype new many-superAdmin AdminACLV2

Note that indexing additional superAdmins would require subgraph updates.


ryley-o:
Cc @lindsgil @yoshiwarab @lyaunzbe as something we could start from if we ever implement this 🙏

PR #1413, opened 2024-02-15
Add standalone SEA testnet minter info

ABDruid:
## Description of the change

Included a link to the SEA minter deployed for TDG only, before available in the shared minter suite.


ryley-o (commented):
@ABDruid While they are using the minter early and approving for their contract, the referenced SEA minter is already recorded as deployed here: https://github.com/ArtBlocks/artblocks-contracts/blob/main/packages/contracts/deployments/minters/sepolia-staging/DEPLOYMENTS.md#shared-minter-deployment-11

I don't think anything additional is needed in this file, yeah?

ABDruid:
Oh sorry, I didn't know that file existed and thought this minter was deployed for their contract only. Please disregard and thanks for sharing

PR #1449, opened 2024-03-26
Add BytecodeStorageV2 w/optional on-chain compression

ryley-o:
## Description of the change

Add BytecodeStorageV2 with optional on-chain compression/decompression.

This updates our BytecodeStorage library to optionally accept compressed strings as input to write operations. A new byte is added to the "meta bytes" section of the library to indicate if the stored data are compressed or not.

A utility view function is added to the BytecodeStorageReader library, `getCompressed`, which provides an easy way to obtain the compressed representation of a string via smart-contract static-call. Of course for gas savings, this would typically be performed off-chain prior to executing a write transaction.

## Not included

ArtBlocks core contracts are not updated to use this library at this time. The update may be implemented some time in the future.

## Performance

*40%-60% cheaper writes, 6x more expensive reads*

Initial explorations indicate ~40%-60% gas cost reductions during write operations for uploading strings such as the Chromie Squiggle script and larger scripts. This is in addition to the existing gas cost efficiency of using contract bytecode to store data.

Reads of compressed data now use a bit more gas, because inflation occurs on-chain. A test is added for to show that when reading a 23.95 KB script, 6x more gas is used during reads (1.5M gas vs. ~0.25M gas). While this is typically performed off-chain, it's worthwhile to note. Additionally, while outside of any spec, some RPC providers impose limits on read operations. Fortunately, 1.5M gas is still relatively minor for a previously maximim-sized bytecode storage contract, and the overall impact of read operations is not expected to be a major hurdle when data are stored in a compressed format.

## TODO

- [x] After initial review, tests may be added for coverage (similar to V1 library)


packages/contracts/contracts/libs/v0.8.x/BytecodeStorageV2.sol
```diff
@@ -0,0 +1,581 @@
...
+                writerAddress,
+                0x20 // return size is entire slot, as it is zero-padded
+            )
+        }
+    }
+
+    /**
+     * @notice Get version for given contract bytecode
+     * @param _address address of deployed contract with bytecode stored in the V0 or V1 format
+     * @return version version read from contract bytecode
```
lindsgil, line 234: version version (sp)

lindsgil:
🔥 🔥 🔥

lindsgil (approved):

lyaunzbe (approved):

ryley-o:
only thing preventing merge is coverage running out of memory - trying skip on some of the memory-heavy new tests during coverage

ryley-o:
Cc @aaronpenne on this one as an interesting PR

aaronpenne:
The practical result is it cost only 0.00538832519430326 ETH to upload my 15kb artwork script! Amazing!
https://etherscan.io/tx/0xc96e24dc89cd36c39d6ca22e503e969eb5305628834a501c27d358033293dcb1

PR #1452, opened 2024-03-29
Offload Flex logic to external library V3 flex bytecode size by offloading logic into external library

ryley-o:
## Description of the change

Reduce V3 Flex bytecode size by offloading logic into a new external library.

## Design

Because the V3 Flex contract is by far the closest to the 24-kb contract size limit, Flex-specific logic and data storage are offloaded into an external library. A similar pattern could be used for all project metadata, but this may be sufficient to accomplish our goals while minimizing impact across the V3 core variants.

## Performance

The following changes are good:

* ~1-mb reduced contract size for Flex V3 core
* Clearer distinction between logic associated with Flex vs. other core contract variants (e.g. data storage is separate)
* Less lines in V3 Flex contract itself

The following changes are drawbacks:
* Slightly higher costs associated with uploading on-chain assets and updating flex-related data
  * note: ~1% script upload cost increase, but this may enable on-chain compression, which reduces that by up to ~50%
 
## Alternatives

* Adding/Updating asset functions may be made internal in the library to eliminate all gas increases associated with uploading assets, but this erodes the 1-mb bytecode reduction to ~0.2-mb bytecode reduction (almost all of the benefit)
* offloading non-flex project metadata may be more lucrative at reducing bytecode size, but would have impacts in performance and complexity across more core contract variants

## Recommendation

Keep this in a draft state until we determine if it is needed and sufficiently impactful to enable other V3 core updates.

### DIFF
--- a/packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine_Flex.sol
+++ b/packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine_Flex.sol
@@ -13,6 +13,7 @@ import "../../interfaces/v0.8.x/IManifold.sol";
 import "@openzeppelin-4.7/contracts/access/Ownable.sol";
 import "../../libs/v0.8.x/ERC721_PackedHashSeed.sol";
 import "../../libs/v0.8.x/BytecodeStorageV1.sol";
+import {V3FlexLib} from "../../libs/v0.8.x/V3FlexLib.sol";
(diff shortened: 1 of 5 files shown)


lindsgil (approved):

lyaunzbe (commented):
reviewing post merge, this looks great to me 🔥 🔥 🚀

PR #1455, opened 2024-04-01
Support ERC-2981 RoyaltyInfo

ryley-o:
## Description of the change

Support the ERC-2981 Royalty standard.

This updates all V3.2 and-on core contracts to support ERC-2981, maximizing our compatibility with marketplaces to maximize support of creator royalties. Note that the Royalty Registry, a less-specific spec than ERC-2981, is automatically compatible with ERC-2981, preventing any regression.

## Design

* Core contracts handle deployment and management of all splitter contracts to minimize changes to project configuration process and to prevent potential for misallocation of funds
* A `SplitProvider` abstraction layer is implemented to enable future updates to be made such that any splitter architecture could be used with the core contracts, as desired
* Initial implementation of `SplitProviderV0` integrates with 0xSplits v2
* The native contract address abstraction is used to deterministically define unique split parameters per deployed splitter address
  * this prevents misallocation of funds, because if a splitter's split parameters are modified while funds are still in the splitter, they will be split incorrectly. Funds can never be assuredly "distributed" from splitter, due to arbitrary ERC-20 tokens potentially being used as royalty payment
* Splitter contracts are namespaced by-core-contract by `SplitProviderV0`, ensuring that a splitter will not be used by more than one core contract
  * intentionally, the same splitter contract will be used if a project has identical split parameters on the core contract (gas optimization, mental model simplicity for artist-owned studio contracts)
* Provider royalty payment details (addresses and BPS) are now considered "default values" at the contract level, and are copied to individual projects. This is required to properly capture changes and ensure state is consistent between all projects and their respective active splitters (looping over all projects is not a safe operation due to block gas limits).
  * admin-only function is provided for admin to sync new defaults to any project, enabling providers to update their payment details within safe block gas limits.
  * Public variable names are updated to reflect the new model, and interfaces are updated (pre-V3.2 Engine interface is added to support legacy subgraph queries in previous model)

```mermaid
---
title: Proposed ERC-2981 Royalty Splitter Architecture
---
erDiagram
    CoreV3_ProjectFinance }|--|| CoreSplitProvider : calls-on-project-updates
    CoreV3_ProjectFinance {
        addressx2 provider_royalty_addrs
        uint16x2 provider_royalty_bps
        address artist
        uint8 artist_total_royalty_pcnt
        address artistAdditional
        uint8 additional_secondary_split_pcnt
    }
    CoreSplitProvider ||--|| v2_0xSplitFactory : createSplit
    CoreSplitProvider {
        function getOrCreateSplitter(splitInputs)returns(address)
    }
    v2_0xSplitFactory {
        function createSplitDeterministic(splitStruct)returns(address)
        function isDeployed(inputs)returns(bool)
    }
    FUTURE_CoreSplitProvider ||--|| FUTURE_DifferentSplitterFactoryService : createSplit
    FUTURE_CoreSplitProvider {
        function getOrCreateSplitter(splitInputs)returns(address)
    }

    FUTURE_DifferentSplitterFactoryService {
        function anyCreateSplitFunction(splitStruct)returns(address)
        function anySplitExistsFunction(splitHash)returns(bool)
    }
    CoreV3_ProjectFinance }|..|| FUTURE_CoreSplitProvider : admin-only-update-split-provider
```

## TODO

- [x] Ensure distribution incentive of 0 is desired for this V0 implementation
- [x] Update existing tests for new changes
- [x] Add new tests for full coverage of new functionality

## Tests

Tests are updated to achieve full coverage of the new functionality, and attempt to check that new splitters are created in all appropriate scenarios to prevent misallocation of funds.

## bytecode size

This adds roughly 0.462 kb of bytecode to the V3 Engine Flex contract implementation. Design choices were made with a heavy-awareness of impact on contract bytecode size due to low margin to 24 kb limit.

## Follow-on

Recommend indexing new `interface IGenArt721CoreContractV3_RoyaltySplitters` event emitted when a project's royalty splitter address is payment. This should be indexed and displayed on frontend to provide proper introspection to person configuring project details.


packages/contracts/contracts/GenArt721CoreV3.sol
```diff
@@ -1268,73 +1367,6 @@ contract GenArt721CoreV3 is
...
     }
 
-    /**
-     * @notice View function returning Artist's secondary market royalty
-     * percentage for project `_projectId`.
-     * This does not include Art Blocks portion of secondary market royalties.
-     * @param _projectId Project ID to be queried.
-     * @return uint256 Artist's secondary market royalty percentage.
-     */
-    function projectIdToSecondaryMarketRoyaltyPercentage(
```
ryley-o, line 1278: removing is to reduce contract bytecode for duplicate accessor - requires subgraph follow-on in `mapping-v3-core.ts:handleProjectSecondaryMarketRoyaltyPercentageUpdated` function to add another try in the case where trying to call the removed function fails.

ryley-o:
This still has failing tests due to interface and constructor changes, but it is ready for review. Tests will be updated and added for full coverage after review, to prevent rework.

packages/contracts/contracts/split/split-provider/SplitProviderV0.sol
```diff
@@ -0,0 +1,190 @@
...
+     */
+    function type_() external pure returns (bytes32) {
+        return TYPE;
+    }
+
+    function _getSplitParams(
+        SplitInputs calldata splitInputs
+    ) private pure returns (ISplitFactoryV2.Split memory splitParams) {
+        // define allocations in BPS to avoid any rounding errors
+        uint256 ArtistAndAdditionalRoyaltyPercentage = splitInputs
```
lindsgil, line 118: any reason why these vars are capitalized 👀 ?

packages/contracts/contracts/split/split-provider/SplitProviderV0.sol
```diff
@@ -0,0 +1,190 @@
...
+            ArtistAndAdditionalRoyaltyBPS +
+            renderProviderRoyaltyBPS +
+            platformProviderRoyaltyBPS;
+
+        // populate each party's split
+        uint256 currentIndex = 0;
+        if (artistNonZero) {
+            splitParams.recipients[currentIndex] = splitInputs.artist;
+            // @dev percent * 100 converts to BPS
+            splitParams.allocations[currentIndex++] =
```
lindsgil, line 164: would the allocation be:

= (ArtistAndAdditionalRoyaltyPercentage - splitInputs.additionalPayeePercentage) * 100?

lindsgil:
Looking good! 🔥 Added a question around how we're calculating allocations

packages/contracts/contracts/split/split-provider/SplitProviderV0.sol
ryley-o: in this case, since "allocation" is an arbitrary value that only defines how royalties are split, we are setting allocations equal to our finest-grained measure of royalty allocation, Basis Points, to prevent any rounding errors.


Basis points are equal to `percent * 100`, so in this case, we can say:
```
totalArtistAndAdditionalAllocation = artistAndAdditionalRoyaltyPercentage * 100;
```
However, we need to further break down artist and additional's split between that total:
```
artistAllocation = totalArtistAndAdditionalAllocation * ((100-additionalPayeePercentage) / 100);
additionalPayeeAllocation = totalArtistAndAdditionalAllocation * (additionalPayeePercentage / 100);
```
>note: the above have major int division rounding errors, do not use directly!

combining those two equations and cancelling out (100/100) gives:
```
artistAllocation = artistAndAdditionalRoyaltyPercentage * (100 - additionalPayeePercentage);
additionalPayeeAllocation = artistAndAdditionalRoyaltyPercentage * additionalPayeePercentage;
```

...which also happen to be free from any rounding errors, and very gas efficient 😄 

I think this is correct, but definitely let me know if it still looks wrong. I'm bad at percentages 💀

packages/contracts/contracts/split/split-provider/SplitProviderV0.sol
ryley-o: good call - fixed in [8557573](https://github.com/ArtBlocks/artblocks-contracts/pull/1455/commits/8557573a08596612c2e7277fcbc173747019b34d)

packages/contracts/contracts/split/split-provider/SplitProviderV0.sol
lindsgil: Nice, that makes sense to me! Thanks for the explanation 🙏

ryley-o:
@lindsgil heads-up I re-added `projectIdToSecondaryMarketRoyaltyPercentage` to greatly reduce downstream complexity for subgraph and potentially other integrations, at the cost of 0.05 kb. much worth it imo to protect some nasty interface complexity in the subgraph. [a21eacf](https://github.com/ArtBlocks/artblocks-contracts/pull/1455/commits/a21eacfdc201f6c5899190c5b77c35834445d543)

packages/contracts/contracts/GenArt721CoreV3.sol
ryley-o: I ended up re-adding this because we now have sufficient bytecode margin, and re-adding removes some undesirable interface complexity in subgraph

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -224,7 +223,7 @@ contract GenArt721CoreV3_Engine is
     bool public immutable autoApproveArtistSplitProposals;
 
     /// version & type of this core contract
-    bytes32 constant CORE_VERSION = "v3.1.4";
+    bytes32 constant CORE_VERSION = "v3.2.4";
```
lindsgil, line 226: nit: should we make this 3.2.0?

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -975,17 +1016,40 @@ contract GenArt721CoreV3_Engine is
...
+            projectId
+        ];
+        projectFinance.artistAddress = _artistAddress;
         projects[projectId].name = _projectName;
         projects[projectId].paused = true;
         projects[projectId].maxInvocations = ONE_MILLION_UINT24;
         projects[projectId].projectBaseURI = defaultBaseURI;
+        // assign default artist royalty of 5% to artist
+        projectFinance
+            .secondaryMarketRoyaltyPercentage = _DEFAULT_ARTIST_SECONDARY_ROYALTY_PERCENTAGE;
```
lindsgil, line 1029: Should we accept `artistSecondaryRoyaltyPercentage` as a param?

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -1073,8 +1140,13 @@ contract GenArt721CoreV3_Engine is
         if (_secondMarketRoyalty > ARTIST_MAX_SECONDARY_ROYALTY_PERCENTAGE) {
```
yoshiwarab, line 1140: Super minor nit/not a new change but do we want to use `_secondaryMarketRoyalty` instead of `_secondMarketRoyalty` for consistency?

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -188,18 +183,22 @@ contract GenArt721CoreV3_Engine is
...
-    /// Basis Points of secondary sales royalties allocated to the
-    /// render provider
-    uint256 public renderProviderSecondarySalesBPS = 250;
-    /// The platform provider payment address for all secondary sales royalty
-    /// revenues
-    address payable public platformProviderSecondarySalesAddress;
-    /// Basis Points of secondary sales royalties allocated to the
-    /// platform provider
-    uint256 public platformProviderSecondarySalesBPS = 250;
+    /// The default render provider payment address for all secondary sales royalty
```
yoshiwarab, line 186: Just to clarify, the only way these could be different on a per project basis is if we updated these and then didn't call `syncProviderSecondaryForProjectToDefaults` for the project right?

packages/contracts/test/core/V3/GenArt721CoreV3_ProjectSplitters.test.ts
```diff
@@ -0,0 +1,616 @@
...
+  "GenArt721CoreV3_Engine", // V3 core Engine contract
+  "GenArt721CoreV3_Engine_Flex", // V3 core Engine Flex contract
+];
+
+/**
+ * Tests for V3 core dealing with configuring splitters on projects.
+ */
+for (const coreContractName of coreContractsToTest) {
+  describe(`${coreContractName} Project Configure`, async function () {
+    async function _beforeEach() {
```
yoshiwarab, line 27: Def not a blocker, but do we want to type the config a little more tightly here?

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine_Flex.sol
```diff
@@ -1085,17 +1126,40 @@ contract GenArt721CoreV3_Engine_Flex is
...
+        projectFinance.artistAddress = _artistAddress;
         projects[projectId].name = _projectName;
         projects[projectId].paused = true;
         projects[projectId].maxInvocations = ONE_MILLION_UINT24;
         projects[projectId].projectBaseURI = defaultBaseURI;
+        // assign default artist royalty of 5% to artist
+        projectFinance
+            .secondaryMarketRoyaltyPercentage = _DEFAULT_ARTIST_SECONDARY_ROYALTY_PERCENTAGE;
+        // copy default platform and render provider royalties to ProjectFinance
+        projectFinance
```
lindsgil, line 1141: Nit: should we update secondaryMarketRoyaltyPercentage to units of BPS so that artist royalty and platform/render provider royalty have the same units

yoshiwarab (approved):

lindsgil (approved):

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
ryley-o: We could, but I wanted to minimize changes to creator dashboard. @Jbern16 - any thoughts on using default vs. adding param to the addProject function?

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine_Flex.sol
ryley-o: I don't want to at this time, mainly to maintain consistency with existing V3 interfaces. Definitely everything should be in BPS when we build a V4 though!

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
ryley-o: exactly 👍

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
ryley-o: absolutely! updated in  [aaf33ff](https://github.com/ArtBlocks/artblocks-contracts/pull/1455/commits/aaf33ff2b1542cbfd5c7f7a1b789b4c7932c07ad)

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
ryley-o: good call - also updated Flex to be v3.2.1! updated in [aaf33ff](https://github.com/ArtBlocks/artblocks-contracts/pull/1455/commits/aaf33ff2b1542cbfd5c7f7a1b789b4c7932c07ad)

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
yoshiwarab: cool cool, do you think it makes sense to add a clarifying dev comment somewhere that calls this out? I think I might assume these were more readily updatable per project just based on the usual context of the `default` prefix.

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
Jbern16: We can add a param if you think it makes sense, changes would be minimal to support it. Could it fallback to default if param isn't provided? I also do wonder if changes like that belong in a major version update though, but I don't have a strong opinion either way.

ryley-o: >Could it fallback to default if param isn't provided?
Only if we added a new external function so one with, one w/o the additional arg, and I don't think we have enough bytecode size margin to justify that

I think in the interest of keeping changes minimal (consider Engine partners with their own UIs as well as CD), I would rather keep this interface as unchanged. @lindsgil do feel more strongly? this could be something we configure in the initializer as well, so it could be defined per-contract.

packages/contracts/test/core/V3/GenArt721CoreV3_ProjectSplitters.test.ts
ryley-o: good call - no more typing issues in my vscode for this file after the changes in [0d1c455](https://github.com/ArtBlocks/artblocks-contracts/pull/1455/commits/0d1c4556570bea61b6d8bdf6f30d138ae3931e1d) 🥳

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
ryley-o: good call - Hopefully the comments in the update functions help a bit as well, and I've added some clarifying comments surrounding these state variables in [38a57bc](https://github.com/ArtBlocks/artblocks-contracts/pull/1455/commits/38a57bc073edd092bff0013eb6fa2af402a2845b)

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
lindsgil: That sounds good to me! Good call out on configuring in the initializer 🙏

ryley-o: cool, let's do that after PR #1514 to prevent the turbulence in current chain of PRs, and we can do some follow-on indexing work to index the default configured value 🙏

PR #1480, opened 2024-04-10
Reference to next contract

ryley-o:
## Description of the change

Add admin-configurable metadata field to core contracts to reference next contract in a series of core contracts.

This provides a forward-looking indicator for the next canonical core contract, if new core contract versions are used for future projects in a series of projects.

## Tests

Tests are added for coverage of the new functionality.


packages/contracts/contracts/GenArt721CoreV3.sol
```diff
@@ -160,6 +161,8 @@ contract GenArt721CoreV3 is
     /// Art Blocks Project ID range: [3-373]
     address public constant ART_BLOCKS_ERC721TOKEN_ADDRESS_V1 =
         0xa7d8d9ef8D8Ce8992Df33D8b8CF4Aebabd5bD270;
+    /// pointer to next contract associated with this contract
```
lindsgil, line 164: Do you think it's worth re-iterating that this is the next `core` contract? 

```
address public nextCoreContract;
```

lindsgil (approved):

packages/contracts/contracts/GenArt721CoreV3.sol
ryley-o: good call! updated in [011ea8c](https://github.com/ArtBlocks/artblocks-contracts/pull/1480/commits/011ea8c84ea909a40648f80a933ac726d8ef8216)

PR #1504, opened 2024-04-18
V3 core - use enums in event fields

ryley-o:
## Description of the change

Use enums in event fields for Project and Platform updated events to reduce contract bytecode size.

This update introduces minimal contract changes to reduce V3 core contract bytecode sizes by ~0.75kb, for which more margin is required to add all v3.2 updates.

This update requires subgraph changes downstream, but changes will be minimal and are straightforward.


yoshiwarab (approved):
LGTM! Does this mean we need add a corresponding map in the subgraph code to decode the platform/project updated events?

ryley-o:
>Does this mean we need add a corresponding map in the subgraph code to decode the platform/project updated events?

Yes definitely 😃

PR #1508, opened 2024-04-18
Custom error codes

ryley-o:
## Description of the change

Revert using custom error codes to reduce core contract bytecode size.

## Overview

This reduces core contract bytecode size by ~1.3 KB, nearly all of which come from removing the revert strings from the contract bytecode (~38 reverts * 32 bytes = ~1.2 KB).

Other changes include fully removing the waffle package, and switching from openzeppelin test helpers to hardhat's chai matchers due to the lack of custom error support in openzeppelin test helpers (`expectRevert` only supports revert strings at this time).

In general, tests were updated and some test bugs were fixed (mostly missing await statements). The use of enums as error codes cleans up our test codebase a bit. Some minter event tests needed to be updated to use a contract instance instead of a library's ABI, specifically in cases where the minter doesn't have the library's event in its ABI.

## Change Driver

These further reductions are required to implement our desired functionality for v3.2 core contracts.

## Tests

Tests are updated to test for appropriate error codes. No new tests are required because there was a 1:1 mapping between revert strings and error codes (no new granularity)

## Follow-on

Parsing of errors on the frontend will be required follow-on work to prevent regression in UX (see https://rc.viem.sh/docs/contract/decodeErrorResult.html)


lindsgil:
LGTM! 🔥

lindsgil (approved):

PR #1514, opened 2024-04-23
Engine Factory Clone Army

lindsgil:
## Description of the change

Engine clone factory for deploying V3.2 Engine and Engine Flex contracts (and beyond). Engine and Engine Flex contracts have been updated to be compatible with ERC-1167, and ERC721_PackedHashSeed was forked to ERC721_PackedHashSeedV1 and updated to be compatible with ERC-1167.

Still outstanding:
- Transfer ownership of the Engine Factory contract to a multisig post-deployment
- Deployment script for Engine/Engine Flex contracts
- Tests


packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,215 @@
...
+                bytecode: type(AdminACLV0).creationCode
+            });
+        }
+
+        address implementation = engineCoreType ==
+            IEngineFactoryV0.EngineCoreType.Engine
+            ? engineImplementation
+            : engineFlexImplementation;
+
+        engineContract = Clones.clone({implementation: implementation});
```
ryley-o, line 118: nit: using cloneDeterministic with a prng salt will essentially prevent the same core contract addresses across multiple chains. alternatively, we could have an input salt to this function which would enable vanity core contract addresses, which might be pretty cool 😃

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,215 @@
...
+     * contract. The address is also emitted in both the `EngineCreated` and
+     * `EngineFlexCreated` events.
+     */
+    function createEngineContract(
+        IEngineFactoryV0.EngineCoreType engineCoreType,
+        EngineConfiguration calldata engineConfiguration,
+        address adminACLContract
+    ) external onlyOwner returns (address engineContract) {
+        require(!isAbandoned, "factory is abandoned");
+        // validate engine contract configuration
```
ryley-o, line 101: would it make sense to validate that adminACLContract and newSuperAdminAddress are either both zero or both non-zero?

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,215 @@
...
+        _onlyNonZeroAddress(engineConfiguration.renderProviderAddress);
+        _onlyNonZeroAddress(engineConfiguration.randomizerContract);
+        if (adminACLContract == address(0)) {
+            // deploy new AdminACLV0 contract
+            adminACLContract = Create2.deploy({
+                amount: 0,
+                salt: keccak256(abi.encodePacked(msg.sender, block.timestamp)),
+                bytecode: type(AdminACLV0).creationCode
+            });
```
ryley-o, line 110: I think you can update superAdmin on the adminACL here instead of later, and just pass an empty array into the core contracts to be updated (since that won't be indexed until later)

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,215 @@
...
+                engineConfiguration.newSuperAdminAddress,
+                genArt721CoreAddressesToUpdate
+            );
+        }
+
+        string memory coreType = IGenArt721CoreContractV3_Engine(engineContract)
+            .coreType();
+        string memory coreVersion = IGenArt721CoreContractV3_Engine(
+            engineContract
+        ).coreVersion();
```
ryley-o, line 140: nit: these can be queried and stored in local state variables in this contract's constructor, since they will be the same across all deployments (one pair for engine, one pair for engine flex)

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,215 @@
...
+        uint256 balance = token.balanceOf(address(this));
+        if (balance > 0) {
+            SafeERC20.safeTransfer({
+                token: token,
+                to: recipient,
+                value: balance
+            });
+        }
+    }
+
```
ryley-o, line 198: I think it makes sense to implement something that allows us to arbitrarily execute calls from this contract, in case third parties make weird assumptions about contract "deployers" being eligible for airdrops, it would prevent loss of funds. Should be doable following something like this: https://github.com/0xSplits/splits-contracts-monorepo/blob/dfaf6f5cfa02ca96989b00ed183f053e44964ae8/packages/splits-v2/src/utils/Wallet.sol#L54
(note: don't recommend using that example directly, as we don't want to broadcast that this contract is _intended_ to actually hold NFTs)

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -208,7 +211,8 @@ contract GenArt721CoreV3_Engine is
     address public minterContract;
 
     /// starting (initial) project ID on this contract
-    uint256 public immutable startingProjectId;
+    /// set on initialization
```
ryley-o, line 214: nit: in general, I think it would be nice to comment on all variables like this that are immutable in practice, but aren't immutable in the compiler due to use of ERC-1167 minimal proxy pattern.

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -346,63 +350,53 @@ contract GenArt721CoreV3_Engine is
...
-        bool _autoApproveArtistSplitProposals,
-        address _splitProviderAddress,
-        bool _nullPlatformProvider,
-        bool _allowArtistProjectActivation
-    ) ERC721_PackedHashSeed(_tokenName, _tokenSymbol) {
-        _onlyNonZeroAddress(_renderProviderAddress);
-        // @dev checks on platform provider addresses performed in _updateProviderSalesAddresses
-        if (_nullPlatformProvider) {
+     */
+    function initialize(
```
ryley-o, line 360: hmm, just realized for state variables that we assign in the contract, I don't see how that would be allowed in the clone pattern. I think these should probably always be set to these values in the initialize function, and not assigned in the contract state variable section (note we do set the platform provider ones to zero sometimes already below). Happy to chat more on this if you have more thoughts!
```
uint8 private _renderProviderPrimarySalesPercentage = 10;
uint8 private _platformProviderPrimarySalesPercentage = 10;
uint256 public defaultRenderProviderSecondarySalesBPS = 250;
uint256 public defaultPlatformProviderSecondarySalesBPS = 250;
```

packages/contracts/contracts/engine/V3/forks/GenArt721CoreV3_Engine_Flex_PROOF.sol
```diff
@@ -99,7 +99,7 @@ import "../../../libs/v0.8.x/Bytes32Strings.sol";
  * Additional admin and artist privileged roles may be described on minters,
  * registries, and other contracts that may interact with this core contract.
  */
-contract GenArt721CoreV3_Engine_Flex_PROOF is
+abstract contract GenArt721CoreV3_Engine_Flex_PROOF is
```
ryley-o, line 102: I think I made these forks a different interface, so they don't need to be abstract anymore?

packages/contracts/hardhat.solidity-config.ts
```diff
@@ -6,7 +6,7 @@ export const solidityConfig = {
       settings: {
         optimizer: {
           enabled: true,
-          runs: 25,
+          runs: 10,
```
ryley-o, line 9: would be great to capture any affect this has on mint gas performance by checking with/without the change using `test/core/gas-tests/GenArt721CoreV3_GasTests_Mint.test.ts`, and noting in PR description.

ryley-o (commented):
The overall architecture here looks great to me! Just minor things on a first pass 👍

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -257,6 +261,8 @@ contract GenArt721CoreV3_Engine is
     // royalty split provider
     ISplitProviderV0 public splitProvider;
 
+    constructor() Ownable(msg.sender) {}
```
ryley-o, line 264: nit - I think continuing to use the non-upgradeable ownable is the right move, and we just transfer to AdminACL in initializer like you are doing. Because this constructor is only ever ran on the implementation contract, I would recommend heavily commenting on why this is only required to satisfy the constructor requirements of the Ownable abstract contract, it not actually used. I think setting owner as `0xdead` or something other than `msg.sender` would help make that clear too?

ryley-o (commented):
Ah, just remembered, let's add convenience functions to add/remove arrays of contracts to core registry, since this will be the only contract that will now be able to do that 😄

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
lindsgil: Talking through this one out loud, LMK if this is correct 🤔 : 

I think one of two things can happen, either:

1. There is no existing AdminACL Contract so we deploy a new one and transfer ownership to the newSuperAdminAddress.. In this case the passed in `adminACLContract` would be the null address (indicating we need to deploy a new one) and the passed in newSuperAdminAddress would be "0x..."

2. There is an existing AdminACL Contract, passed in as "0x...", which we use to initialize and transfer ownership of the new Engine contract (no need to deploy a new one but in this case whether the newSuperAdmin address was null or not, we can't call transferOwnership of the AdminACL contract bc it would need to be called by the super admin)

So I think the only check we would need to make is: 
`if (adminACLContract == address(0)) -> newSuperAdminAddress != address(0)`

ryley-o: ah, my bad, yes you are correct! do you think it also makes sense to also validate that they don't pass in an existing AdminACL contract with a non-null newSuperAdmin, because that wouldn't be possible to assign in practice?

lindsgil: that's a good call! is there any other validation that should be done there? validating that the newSuperAdmin address passed in is the owner of the AdminACL contract passed in?

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,215 @@
...
+    constructor(
+        address engineImplementation_,
+        address engineFlexImplementation_,
+        address coreRegistry_
+    ) {
+        coreRegistry = coreRegistry_;
+        engineFlexImplementation = engineFlexImplementation_;
+        engineImplementation = engineImplementation_;
+        // set owner of core registry contract
+        Ownable(coreRegistry_).transferOwnership(address(this));
```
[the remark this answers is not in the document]
lindsgil: 🙏 Good catch, removed this!

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,215 @@
...
+     * implementation contract
+     * @param engineFlexImplementation_ address of the Engine Flex
+     * implementation contract
+     * @param coreRegistry_ address of the core registry contract
+     */
+    constructor(
+        address engineImplementation_,
+        address engineFlexImplementation_,
+        address coreRegistry_
+    ) {
```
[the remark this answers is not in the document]
lindsgil: Good call! I added owner here, checked that it is nonZero and then made a call to _transferOwnership 🙏

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
lindsgil: 🔥 🔥  vanity core contract addresses, so cool! Updated this one to `cloneDeterministic` with salt param 🙏

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
lindsgil: Updated! Does that get indexed when we call `_transferOwnership(_adminACLContract);`?

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
lindsgil: Nice! I ported the function over but removed `value` from the struct as I don't think claims typically need any value sent in the tx 🤔  Do you think it's worth any clarifying comments that this is intended to prevent loss of funds?

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
lindsgil: 🙏 Added additional clarification here and to the other affected variables

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
lindsgil: Good call! I moved the assignment of these into the initialize function 🙏

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
ryley-o: I believe we only ping the admin ACL for superAdmin in the subgraph when a contract is assigned to it - so as long as we update superAdmin before the contract starts using it, it should index appropriately 😃

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
lindsgil: Thanks for catching this!! Added a comment here and in flex + updated to 0xdead 🙏 :rip:

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
ryley-o: lol sorry I think not helping 🤣 wdyt about the following logic?
```
if (adminACLContract == addr(0)) {
    // require non-zero newSuperAdminAddress
    // deploy new splitter, assign to new super admin
} else {
    // require zero newSuperAdminAddress
    // use existing splitter, a non-zero new super admin is nonsense here
}
```

ryley-o: or could totally do what you describe and just check the superAdmin on an existing superAdmin 👍 I think either way would be good!

packages/contracts/contracts/libs/v0.8.x/ERC721_PackedHashSeedV1.sol
```diff
@@ -0,0 +1,592 @@
...
+/**
+ * @dev Forked version of the OpenZeppelin v5.0.0 ERC721 contract. Updated
+ * with an initialize function to ensure EIP 1167 compatibility. Utilizes
+ * a struct to pack owner and hash seed into a single storage slot.
+ * ---------------------
+ * @dev Implementation of https://eips.ethereum.org/EIPS/eip-721[ERC721] Non-Fungible Token Standard, including
+ * the Metadata extension, but not including the Enumerable extension, which is available separately as
+ * {ERC721Enumerable}.
+ */
+contract ERC721_PackedHashSeedV1 is
```
ryley-o, line 24: can this be marked as `abstract contract` to stay consistent w/OpenZeppelin?

packages/contracts/contracts/libs/v0.8.x/ERC721_PackedHashSeedV1.sol
```diff
@@ -0,0 +1,592 @@
...
+pragma solidity ^0.8.20;
+
+import "@openzeppelin-5.0/contracts/token/ERC721/IERC721.sol";
+import "@openzeppelin-5.0/contracts/token/ERC721/IERC721Receiver.sol";
+import "@openzeppelin-5.0/contracts/token/ERC721/extensions/IERC721Metadata.sol";
+import "@openzeppelin-5.0/contracts/utils/Address.sol";
+import "@openzeppelin-5.0/contracts/utils/Context.sol";
+import "@openzeppelin-5.0/contracts/utils/Strings.sol";
+import "@openzeppelin-5.0/contracts/utils/introspection/ERC165.sol";
+import "@openzeppelin-5.0/contracts/interfaces/draft-IERC6093.sol";
```
ryley-o, line 13: can you update this to use named imports, same as OpenZeppelin's implementation?

packages/contracts/contracts/libs/v0.8.x/ERC721_PackedHashSeedV1.sol
```diff
@@ -0,0 +1,592 @@
...
+ * {ERC721Enumerable}.
+ */
+contract ERC721_PackedHashSeedV1 is
+    Context,
+    ERC165,
+    IERC721,
+    IERC721Metadata,
+    IERC721Errors
+{
+    using Address for address;
```
ryley-o, line 31: is this addition needed?

packages/contracts/contracts/libs/v0.8.x/ERC721_PackedHashSeedV1.sol
```diff
@@ -0,0 +1,592 @@
...
+    mapping(uint256 => OwnerAndHashSeed) internal _ownersAndHashSeeds;
+
+    // Mapping owner address to token count
+    mapping(address => uint256) private _balances;
+
+    // Mapping from token ID to approved address
+    mapping(uint256 => address) private _tokenApprovals;
+
+    // Mapping from owner to operator approvals
+    mapping(address => mapping(address => bool)) private _operatorApprovals;
```
ryley-o, line 62: nit: can you used named variables in the mappings to stay consistent with OpenZeppelin implementation?

ryley-o (approved):
Gave this a full review and lgtm 🚀 ! Only nits, and will mark as do not merge until you are able to fully test 🥇

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,300 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+// @dev fixed to specific solidity version for clarity and for more clear
+// source code verification purposes.
+pragma solidity 0.8.19;
```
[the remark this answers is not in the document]
lindsgil: I think AdminACLV0 is preventing us from bumping, I get the following error during compile 

```
Error HH606: The project cannot be compiled, see reasons below.
These files import other files that use a different and incompatible version of Solidity:
  * contracts/engine/V3/EngineFactoryV0.sol (0.8.22) imports contracts/AdminACLV0.sol (0.8.19)
  ```
  
  OpenZeppelin V5 uses ^0.8.20 in Create2 and Clones 🤔

File: packages/contracts/contracts/engine/V3/EngineFactoryV0.sol (hunk 5 above)
Comment (lindsgil) [reply packages/contracts/contracts/engine/V3/EngineFactoryV0.sol:251]: Awesome 🙏 Updated the comment here

File: packages/contracts/contracts/engine/V3/EngineFactoryV0.sol [hunk 18]
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        address implementation = engineCoreType ==
+            IEngineFactoryV0.EngineCoreType.Engine
+            ? engineImplementation
+            : engineFlexImplementation;
+
+        engineContract = Clones.cloneDeterministic({
+            implementation: implementation,
+            salt: salt
```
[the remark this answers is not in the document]
lindsgil: 🙏  Added this!

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,300 @@
...
+     * @notice Execute a batch of calls.
+     * @dev The calls are executed in order, reverting if any of them fails. Can
+     * only be called by the owner.
+     * @param _calls The calls to execute
+     */
+    function execCalls(
+        Call[] calldata _calls
+    )
+        external
+        onlyOwner
```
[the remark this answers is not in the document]
lindsgil: 🙏 Updated!

packages/contracts/contracts/mock/MockAdminACLV0_PROHIBITION.sol
```diff
@@ -63,6 +63,27 @@ contract MockAdminACLV0Events_PROHIBITION is IAdminACLV0_PROHIBITION, ERC165 {
             contractArtistApprovals[hashArtistApprovalKey(_contract, _caller)];
     }
 
+    /**
```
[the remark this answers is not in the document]
lindsgil: So I added `changeSuperAdmin` to the interface of AdminACLV0 (called within createEngineContract of the EngineFactory) and IAdminACLV0_PROHIBITION inherits from IAdminACLV0, getting the `Contract "MockAdminACLV0Events_PROHIBITION" should be marked as abstract` compile error without it

packages/contracts/contracts/libs/v0.8.x/ERC721_PackedHashSeedV1.sol
lindsgil: Good call, updated!

packages/contracts/contracts/libs/v0.8.x/ERC721_PackedHashSeedV1.sol
lindsgil: Ah, good catch - removed this!

packages/contracts/contracts/libs/v0.8.x/ERC721_PackedHashSeedV1.sol
lindsgil: Good call! Updated 🙏

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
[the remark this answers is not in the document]
ryley-o: Can AdminACLV0 also be bumped up to v0.8.22 to resolve the compile error?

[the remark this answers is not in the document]
lindsgil: Great call, bumped and updated to OpenZeppelin to v5 saving .444 !! 🚀 🙏

ryley-o:
@yoshiwarab would be great to get your eyes on this one, especially the re-forking of ERC721_PackedHashSeedV1.sol from OpenZeppelin V5. That one goes deep into our core, but the diff between OpenZeppelin's implementation and our own is minimal 👍 🙏

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,349 @@
...
+        // emit event
+        emit Abandoned();
+    }
+
+    /**
+     * @dev This contract is not intended to hold funds. This function,
+     * `drainETH`, and `drainERC20` are implemented to prevent the loss
+     * of funds that might be sent to this contract inadvertently.
+     */
+    receive() external payable {}
```
lindsgil, line 250: @ryley-o Added with a clarifying comment

ryley-o: nice - was this function required?

packages/contracts/contracts/interfaces/v0.8.x/IAdminACLV0.sol
```diff
@@ -23,6 +23,18 @@ interface IAdminACLV0 {
...
+     * @param _newSuperAdmin The new superAdmin address.
+     * @param _genArt721CoreAddressesToUpdate Array of genArt721Core
+     * addresses to update to the new superAdmin, for indexing purposes only.
+     * @dev this function is gated to only superAdmin address.
+     */
+    function changeSuperAdmin(
+        address _newSuperAdmin,
+        address[] calldata _genArt721CoreAddressesToUpdate
+    ) external;
+
```
ryley-o, line 37: @lindsgil I think this needs to be broken out into a new interface, because I believe it will prevent _current_ AdminACLV0's from transferring ownership to the new AdminACLV0's, because when switching there is a forward-looking check for conformance on the new AdminACL for `type(IAdminACLV0).interfaceId`, which changes if we change this interface file.

ryley-o: let's also add a test that hard-codes that this file's interfaceId never changes 😅 because how would anyone ever know this 😅 😅

packages/contracts/hardhat.solidity-config.ts
lindsgil: Great call! 🙏 Updated the description 🙏

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -353,63 +383,61 @@ contract GenArt721CoreV3_Engine is
...
-            // set platform to zero revenue splits
-            _platformProviderPrimarySalesPercentage = 0;
-            defaultPlatformProviderSecondarySalesBPS = 0;
+     */
+    function initialize(
+        EngineConfiguration calldata engineConfiguration,
+        address _adminACLContract
+    ) external {
+        // can only be initialized once
+        if (initialized) {
```
ryley-o, line 398: nit: would be good to test if-branch here to ensure test failure if the check were removed

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -0,0 +1,349 @@
...
+     * Flex contracts.
+     */
+    address public immutable engineFlexImplementation;
+
+    // Address of core registry contract.
+    address public immutable coreRegistry;
+
+    /// version and type of Engine implementation contract
+    string public coreType;
+    string public coreVersion;
```
ryley-o, line 52: considering there are also `flexCoreType`/version, wdyt about naming these something like `engineCoreType`/version? idk if that is better or not, but just wanted to double click

packages/contracts/contracts/interfaces/v0.8.x/IEngineFactoryV0.sol
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * @notice Drains the contract's balance of an input ERC20 token to
+     * the `recipient`.
+     * @param ERC20TokenAddress The address of the ERC20 token to withdraw.
+     * @param recipient The address to send ERC20 tokens to.
+     * Only callable by the owner.
+     */
+    function drainERC20(address ERC20TokenAddress, address recipient) external;
+
+    /**
```
ryley-o, line 72: realized it might be helpful to have a function for `transferCoreRegistryOwnership` 🤷 We could use the execCalls function to do this, but would be nice to have a function that helps us cleanly perform this operation, since it is expected behavior once we upgrade to new core contract versions.

ryley-o (approved):
Took a final look here - a couple very minor nits/comments, appreciate all the useful details in this PR! 🙏 ❤️

yoshiwarab (approved):

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine_Flex.sol
```diff
@@ -268,9 +287,27 @@ contract GenArt721CoreV3_Engine_Flex is
     bytes4 private constant _INTERFACE_ID_ERC2981 = 0x2a55205a;
     uint8 private constant _DEFAULT_ARTIST_SECONDARY_ROYALTY_PERCENTAGE = 5;
 
+    /// strings library from OpenZeppelin, modified for no constants
+    /// set upon initialization to conform to the ERC-1167 minimal
+    /// minimal proxy pattern, which necessitates setting these
+    /// values post-deployment.
+    bytes16 private _HEX_SYMBOLS;
```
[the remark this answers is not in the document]
ryley-o: good call - also perhaps we could just use the strings library like Engine does. we did this for bytecode reasons a while back, but I think a little more margin at the moment?

packages/contracts/contracts/interfaces/v0.8.x/IAdminACLV0.sol
lindsgil: Thanks for catching this!! 😅 😅 I added tests to check that AdminACLV0 supports hardcoded IAdminACLV0 and IAdminACLV0_Engine (new interface) interface IDs.. LMK what you think of that approach or if there's a more robust way!

```
    it("supports IAdminACLV0 and IAdminACLV0_Engine Interface IDs", async function () {
      const config = await loadFixture(_beforeEach);
      // @dev these interface IDs are hardcoded here to ensure
      // any future changes do not break deployed contracts
      // IAdminACLV0 interface ID
      const iadminACLV0InterfaceId = 0xc00707bc;
      // IAdminACLV0_Engine interface ID
      const iadminACLV0_EngineInterfaceId = 0x377a7d1a;

      expect(
        await config.adminACL_InterfaceBroadcast.supportsInterface(
          iadminACLV0InterfaceId
        )
      ).to.be.true;
      expect(
        await config.adminACL_InterfaceBroadcast.supportsInterface(
          iadminACLV0_EngineInterfaceId
        )
      ).to.be.true;
    });
    ```

File: packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol (hunk 23 above)
Comment (lindsgil) [reply packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol:398]: good call 🙏 Added tests to `GenArt721CoreV3_ContractConfigure.test` to make sure re-initialization is not possible 🙏

File: packages/contracts/contracts/engine/V3/EngineFactoryV0.sol (hunk 24 above)
Comment (lindsgil) [reply packages/contracts/contracts/engine/V3/EngineFactoryV0.sol]: good call! updated this to reflect `engineCoreType/version`

File: packages/contracts/contracts/interfaces/v0.8.x/IEngineFactoryV0.sol (hunk 25 above)
Comment (lindsgil) [reply packages/contracts/contracts/interfaces/v0.8.x/IEngineFactoryV0.sol:72]: Good call! I added this function + tests in `EngineFactoryV0/integration.test.ts` .. I gave it the onlyOwner modifier but is there anything else we should check for in that function?

File: packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine_Flex.sol (hunk 26 above)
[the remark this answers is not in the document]
Comment (lindsgil) [reply packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine_Flex.sol]: Good call! was able to remove these and associated helpers in favor of the strings library.. Also brought down the bytecode size a tiny bit (23.274 vs. 23.338) 🙏

File: packages/contracts/contracts/engine/V3/EngineFactoryV0.sol (hunk 21 above)
Comment (ryley-o) [reply packages/contracts/contracts/engine/V3/EngineFactoryV0.sol:260]: cool, I forgot that is required - from the docs
```
When Ether is sent directly to a contract (without a function call, i.e. sender uses send or transfer) but the receiving contract does not define a receive Ether function or a payable fallback function, an exception will be thrown, sending back the Ether (this was different before Solidity v0.4.0).
```

packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine_Flex.sol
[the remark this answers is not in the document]
ryley-o: haha funny because we originally did this because it reduced bytecode size 😆 good job OZ v5

packages/contracts/contracts/interfaces/v0.8.x/IEngineFactoryV0.sol
ryley-o: lgtm! I think no additional checks needed, especially since the engine factory is really just forwarding owner calls to the coreRegistry 👍

packages/contracts/contracts/interfaces/v0.8.x/IAdminACLV0.sol
ryley-o: lgtm as discussed and updated 👍

PR #1518, opened 2024-04-26
Provision future Flex 💪 

ryley-o:
## Description of the change

Provision the Flex contract with a few key items to unlock at a future date.

linear: https://linear.app/art-blocks/issue/PRO-194/provision-v32-core-wfuture-flex-capabilities

Provisions include: 

* Add ability to define on-chain from existing BytecodeStorage address
* Add new type that points to dependency registry type and version
* Add ability to perform on-chain compressed flex uploads

Additionally, for a cleaner UX, add guards to only allow removal of last Flex asset to prevent accidental script breaking (already constraining on CD)

Note: this also has a small bug-fix dealing with stale bytecode addresses not being properly cleared, which has existed on previous V3 Flex contracts, but doesn't meaningfully impact the function of the contract, in [520b0d4](https://github.com/ArtBlocks/artblocks-contracts/pull/1518/commits/520b0d41e600df5560236c1875bcc00193d7729b)

## Design

A few decisions to note:
* opted to NOT perform on-chain validation of input dependency name and version against what is populated on the contract's referenced on-chain dependency registry.
  * partially because L1 reads are not available from L2 contracts, partially because crafting a "dependency exists" function is a bit hacky at the moment.
* the existing cidOrData field was used to store a string representation of the dependency registry's `bytes32 dependencyNameAndVersion`
* new external functions were added for uploading existing on-chain assets and uploading compressed assets, instead of adding new optional args to existing asset upload functions
  * maintains backwards-compatible interface conformance
  * adds bytecode for the new external functions
* library functions were kept external, even when dealing with large quantities of data transfer, due to bytecode size limits

## Bytecode changes

These changes add roughly 0.53 kb of bytecode to the V3 Engine Flex contract, largely due to the new external functions.

## Tests

- [ ] Add tests for full coverage of new functionality
- [x] Confirm minor stale field bug fix behavior is resolved 

## Immediate follow-on

The following updates are required to avoid potential subgraph and general indexing failures when indexing these contracts:

- New enum type added to subgraph for FLEX_CONTRACT_EXTERNAL_ASSET_DEP_TYPES
- New enum type added to database

Other follow-on updates required include updating CD to provide interfaces to these new functions.

### DIFF
--- a/packages/contracts/contracts/libs/v0.8.x/V3FlexLib.sol
+++ b/packages/contracts/contracts/libs/v0.8.x/V3FlexLib.sol
@@ -4,7 +4,7 @@
 pragma solidity ^0.8.0;
 
 import {IGenArt721CoreContractV3_Engine_Flex} from "../../interfaces/v0.8.x/IGenArt721CoreContractV3_Engine_Flex.sol";
-import {BytecodeStorageWriter, BytecodeStorageReader} from "./BytecodeStorageV1.sol";
+import {BytecodeStorageWriter, BytecodeStorageReader} from "./BytecodeStorageV2.sol";
 
 /**
  * @title Art Blocks V3 Engine Flex - External Helper Library
@@ -16,6 +16,7 @@ import {BytecodeStorageWriter, BytecodeStorageReader} from "./BytecodeStorageV1.
 
 library V3FlexLib {
     using BytecodeStorageWriter for string;
+    using BytecodeStorageWriter for bytes;
     // For the purposes of this implementation, due to the limited scope and
     // existing legacy infrastructure, the library emits the events
     // defined in IGenArt721CoreContractV3_Engine_Flex.sol. The events are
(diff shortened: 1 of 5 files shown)


packages/contracts/contracts/libs/v0.8.x/V3FlexLib.sol
```diff
@@ -177,7 +186,13 @@ library V3FlexLib {
             // we don't want to emit data, so we emit the cid as an empty string
             _cidOrData = "";
         } else {
+            // incoming dependency type is not ONCHAIN, so we set the cid directly with either
+            // the incoming cid or string representation of the dependencyNameAndVersion
             flexProjectData.externalAssetDependencies[_index].cid = _cidOrData;
+            // clear any previously populated bytecode address
+            flexProjectData
+                .externalAssetDependencies[_index]
+                .bytecodeAddress = address(0);
```
ryley-o, line 195: this is a small bug fix that clears stale data. exists in all previously deployed V3 Flex contracts, but inconsequential.

yoshiwarab (approved):

lindsgil (approved):

packages/contracts/contracts/libs/v0.8.x/V3FlexLib.sol
ryley-o: confirmed that this does fix a bug via testing 👍 [4de1cc2](https://github.com/ArtBlocks/artblocks-contracts/pull/1518/commits/4de1cc2aa2208c9d5ef0336b8b7897c3fa45ebfb) vs. [2bba50e](https://github.com/ArtBlocks/artblocks-contracts/pull/1518/commits/2bba50e72baaa45063b88297f7e51274420d369c)

PR #1531, opened 2024-05-07
Update minter contract during initialization

lindsgil:
## Description of the change

Update minter contract on Engine and Engine Flex contracts during initialization to the passed in value if it is not the zero address.


packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -421,7 +421,10 @@ contract GenArt721CoreV3_Engine is
             engineConfiguration.tokenName,
             engineConfiguration.tokenSymbol
         );
-
+        // update minter if populated
+        if (engineConfiguration.minterFilterAddress != address(0)) {
+            minterContract = engineConfiguration.minterFilterAddress;
```
lindsgil, line 426: @ryley-o Note - I think we need to emit the `MinterUpdated` event here but we ran out of space 💀 C

lindsgil: Resolved offline 🙏

ryley-o (approved):
🙏 thank you for finding this!

PR #1532, opened 2024-05-08
Integrate XState project sale machines into sdk

yoshiwarab:
## Description of the change
This PR updates the `@artblocks/sdk` package to include and export a set of XState state machines that model the logic of a project sale. Additionally it updates the organization of the ArtBlocksClient (the main export of the package) to accept both WalletClient and PublicClient objects in its constructor. All other functions called on the client should use the classes wallet/public clients for web3 interactions.


packages/sdk/abis/iSharedMinterDAExpSettlementV0Abi.ts
```diff
@@ -0,0 +1,123 @@
+export const iSharedMinterDAExpSettlementV0Abi = [
```
ryley-o, line 1: any reason why the abis aren't imported from the @artblocks/contracts package?

ryley-o (approved):
lgtm! this is a big one!

packages/sdk/abis/iSharedMinterDAExpSettlementV0Abi.ts
yoshiwarab: So we get type safety with viem we need them exported from a ts file `as const`. Definitely would like to add a step that generates ts files for the contracts package as well

grantoesterling (approved):

PR #1540, opened 2024-05-13
Initial studio deploys

ryley-o:
## Description of the change

Initial studio deployments + staging deployments.

Includes the following deployments:

- [x] mainnet studio 0, 1
- [x] artist-staging, engine + flex contracts to begin using for new shells

note: will follow-on with arbitrum deployment PR! (need address before deploying)

## Process
All production deployments are deployed via multisig.

Ideally,
1. the PR creator proposes the transactions to the Safe
2. PR reviewer executes the transactions as part of approval, executes post-deployment script, merges PR


packages/contracts/deployments/engine/V3/studio/mainnet/_ENGINE_efficient_addresses_mainnet-delete-when-used.txt
```diff
@@ -0,0 +1,336 @@
+0x0000000000000000000000000000000000000000da5f4baa4a7e013604000020 => 0x0000002Fd3370D5cF4b8f546AaC06c574F3d0244 => 1
```
lindsgil, line 1: Good call adding these! :) 🔥 🙏

packages/contracts/scripts/engine/V3/post-batch-create-engine-contracts.ts
```diff
@@ -183,12 +183,16 @@ async function main() {
 
       // create image bucket
       let imageBucketCreated = false;
+      // For studio contracts use token ticker as bucket name, otherwise use token
```
lindsgil, line 186: Note - Updated the bucket naming logic slightly so that studio contracts use the token ticker instead of the token name, alternatively we could remove the pipe symbol in the logic, I'm open to either! 🙏

yoshiwarab (approved):

PR #1549, opened 2024-05-22
improve multichain support via configurable defaultBaseURIHost

ryley-o:
## Description of the change

Add Default Base URI Prefix as EngineFactory setting for improved multi-chain support.

This enables Engine factories to automatically populate the correct token API prefix (e.g. `https://token.arbitrum.artblocks.io/`) without additional input parameters to the factory per-deployment.

Note that `EngineConfiguration` is unchanged in this implementation, enabling `EngineConfiguration` to remain as the input struct for new deployment calls to the EngineFactoryV0.

Makes constructor args (an therefore deployed address when using create2) necessarily unique per chain, whereas prior to this change they just-so-happened to be unique per chain (due to how coreRegistry was previously deployed). This is acceptable, and helps avoid address collisions across chains, a characteristic we deemed desirable.

Engine and Engine Flex patch versions incremented.

Adds 0.063 kb to bytecode.

## TODO

- [x] Update code
- [x] Review
- [x] Update tests

## Follow-on

* new implementation and engine factory deployments should be made on arbitrum mainnet.  Likely will deploy across all networks to keep common addresses the same for simplified docs.


packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
```diff
@@ -33,6 +33,9 @@ contract EngineFactoryV0 is Ownable, IEngineFactoryV0 {
     // public type
     bytes32 public constant type_ = "EngineFactoryV0";
 
+    // base URI prefix for all Engine contracts on this network/chainID
+    string public defaultBaseURIPrefix;
```
lindsgil, line 37: nit: should we call this `defaultBaseURIHost`?

lindsgil (approved):
🔥 🔥 🔥

packages/contracts/contracts/engine/V3/EngineFactoryV0.sol
ryley-o: YES! updated in [3040145](https://github.com/ArtBlocks/artblocks-contracts/pull/1549/commits/3040145ef1841d6e514e9657c195d62234715905)

ryley-o:
note that no new tests are required, as this functionality is already covered in test labeled `returns default base URI if projectURI is not populated`

PR #1587, opened 2024-06-11
v3.2.4, .5 future proof bytecode storage reader

ryley-o:
## Description of the change

Update Engine and Flex contracts to use universal bytecode storage reader instead of being tied to a specific version.

This adds the ability for a core contract to read on-chain assets deployed with future versions of our BytecodeStorage library. While non-impacting for the current Engine contract design, this does provide improved functionality for the current Engine Flex contract, where shared on-chain assets may be used.

TODO

- [x] Update contracts
- [x] Reviews
- [x] Update existing tests to pass
- [x] Add new tests for coverage


packages/contracts/contracts/engine/V3/GenArt721CoreV3_Engine.sol
```diff
@@ -2390,6 +2376,87 @@ contract GenArt721CoreV3_Engine is
...
+     * @notice Initializes the contract with the provided `engineConfiguration`.
+     * This function should be called atomically, immediately after deployment.
+     * Only callable once. Validation on `engineConfiguration` is performed by caller.
+     * @param engineConfiguration EngineConfiguration to configure the contract with.
+     * @param adminACLContract_ Address of admin access control contract, to be
+     * set as contract owner.
+     * @param defaultBaseURIHost Base URI prefix to initialize default base URI with.
+     * @param bytecodeStorageReaderContract_ Address of bytecode storage reader contract.
+     */
+    function _initialize(
```
ryley-o, line 2389: moved initialize to internal function - something we want to implement before #1576 for better inheritance properties

yoshiwarab (approved):

lindsgil (approved):

PR #1615, opened 2024-06-24
Beny/plt 519 support for settlement claim flow on old contracts

yoshiwarab:
## Description of the change

Updates excess settlement funds claim machine to support older non-shared DAExpSettlement minters

Fixes: PLT-519

### DIFF
--- a/packages/sdk/src/machines/utils.ts
+++ b/packages/sdk/src/machines/utils.ts
@@ -20,6 +20,13 @@ export const SUPPORTED_MINTER_TYPES = [
   Minter_Type_Names_Enum.MinterDaExpSettlementV3,
 ];
 
+export const SUPPORTED_SETTLEMENT_CLAIM_MINTER_TYPES = [
+  Minter_Type_Names_Enum.MinterDaExpSettlementV0,
+  Minter_Type_Names_Enum.MinterDaExpSettlementV1,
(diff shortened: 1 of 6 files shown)


mattetre (approved):
🙌 💰 

Nice work! Thanks for getting this in so quick 🙏 So for clients (platform) to take advantage of this all that is needed is to bump the SDK version right?

packages/sdk/src/machines/utils.ts
```diff
@@ -20,6 +20,13 @@ export const SUPPORTED_MINTER_TYPES = [
   Minter_Type_Names_Enum.MinterDaExpSettlementV3,
 ];
 
+export const SUPPORTED_SETTLEMENT_CLAIM_MINTER_TYPES = [
+  Minter_Type_Names_Enum.MinterDaExpSettlementV0,
```
ryley-o, line 24: I'm pretty sure we never deployed a V0 minter, but fine to leave here if you want

ryley-o (approved):
LGTM!

packages/sdk/src/machines/utils.ts
yoshiwarab: Yeah, I just figured I'd include all of them because they have the same ABI for the relevant functions

PR #1631, opened 2024-07-29
hide polyptych minters from minter selection form

ryley-o:
## Description of the change

Hide polyptych minters from minter selection form.

This is because polyptych minters are generally unsupported in the SDK at this time.


yoshiwarab:
@ryley-o, added a comment in linear, but I think we probably don't want to put this in the SDK logic, and instead just provide messaging when a specific minter isn't supported on marketplace. I don't think we want to prevent users from using the CD to select minters that require minting flow outside of marketplace.

PR #1641, opened 2024-08-12
Support ram bid UI updates

yoshiwarab:
## Description of the change

Minor changes to RAM bid machine to support UI updates. Also updates husky hooks to work with corepack.

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@artblocks/sdk",
-  "version": "0.1.28-alpha",
+  "version": "0.1.29",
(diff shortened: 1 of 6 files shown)


ryley-o (approved):

packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@artblocks/sdk",
-  "version": "0.1.29",
```
lindsgil, line 3: do we want the extra `-0`?

yoshiwarab: Yeah, that's how yarn4 version command does prerelease versioning and didn't want to do a full release for now.

see https://yarnpkg.com/cli/version

PR #1649, opened 2024-08-27
Minter MinPrice

ryley-o:
## Description of the change

Add new min-price minter.

Design Doc: https://docs.google.com/document/d/1oPHNkciNWQuZOQ5rSZQQdEeG04kBpx_-5sG-n11GnZg/edit#heading=h.4h02abif4pct 

closes https://linear.app/art-blocks/issue/PLT-660/develop-minterminprice-smart-contracts

diffs (as of [9302903](https://github.com/ArtBlocks/artblocks-contracts/pull/1649/commits/9302903b3d6dcceddfa9ae5af06edbec583b1ee3)):
* MinterSetPriceV5<->MinterMinPriceV0: https://www.diffchecker.com/YNJK2ih1/
* MinterSetPriceMerkleV5<->MinterMinPriceMerkleV0: https://www.diffchecker.com/xKOcLBY8/

## Design

* The new minters use a new `MinPriceLib` to store and emit events for the admin-controlled minimum fee at the contract level. A new handler for the lib's single event will be added to the subgraph.
* Same as existing fixed price minters, the minters require the artist to call `updatePricePerTokenInWei(projectId, coreContract, pricePerTokenInWei)` with a price greater than or equal to the current minimum mint fee for their project. The project price is then set to the input price, and is defined at the project level like our other set price minters.
* Funds are sent to the render provider only instead of being split by percentages, since this is a minimum fee minter.

## TODO

- [ ] review of smart contracts
- [x] add tests for 100% coverage


packages/contracts/contracts/interfaces/v0.8.x/ISharedMinterMinPriceV0.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+// Created By: Art Blocks Inc.
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title This interface adds support for serial English auction minting.
```
lindsgil, line 7: SEA -> Min price

packages/contracts/contracts/libs/v0.8.x/minter-libs/SplitFundsLib.sol
```diff
@@ -96,6 +96,44 @@ library SplitFundsLib {
...
+     * called directly from the contract that is receiving the payment.
+     * @dev possible DoS during splits is acknowledged, and mitigated by
+     * business practices, including end-to-end testing on mainnet, and
+     * admin-accepted artist payment addresses.
+     * @param projectId Project ID for which funds shall be split.
+     * @param pricePerTokenInWei Current price of token, in Wei.
+     * @param coreContract Address of the GenArt721CoreContract associated
+     * with the project.
+     */
+    function sendAllToRenderProviderETHRefundSender(
```
lindsgil, line 116: when would this function get called?

lindsgil:
Looking great! 🔥 🔥

packages/contracts/contracts/interfaces/v0.8.x/ISharedMinterMinPriceV0.sol
ryley-o: fixed in [93ed329](https://github.com/ArtBlocks/artblocks-contracts/pull/1649/commits/93ed329f51b2b5ae6e8253c778267e22bf40ea2d)

packages/contracts/contracts/libs/v0.8.x/minter-libs/SplitFundsLib.sol
ryley-o: we use it in the new minters because we allow the purchaser to send >price (consistent with existing fixed price minters). We could definitely update the logic in these minters to force `msg.value == price`, and then not run through the refund logic.
what do you think?

lindsgil: makes sense, I think fine to leave as is! 🙏

lindsgil (approved):
🚀 🚀 🚀

PR #1673, opened 2024-09-19
Sort minter selection list alphabetically

ryley-o:
## Description of the change

Sort minter selection list alphabetically for improved UX.

Result looks like:

Screenshot 2024-09-19 at 11 38 45 AM


lindsgil (approved):
🔡

PR #1686, opened 2024-10-02
Handle post ram auction state with a different machine/project sale manager state

yoshiwarab:
## Description of the change

This PR updates the ramMachine and introduces a new postRamMachine. Specifically instead of including a postSale state in the ram machine we now have a postRamMachine that is spawned when a project has a RAM minter assigned and the minter is in a post auction state. This reduces edge-cases where the machine might end up in an invalid state if the auction ends while the user is in the middle of placing a bid.


lindsgil (approved):

packages/sdk/src/machines/post-ram-machine/index.ts
```diff
@@ -0,0 +1,126 @@
...
+};
+
+export type PostRAMMachineInput = Pick<
+  PostRAMMachineContext,
+  "artblocksClient" | "project"
+>;
+
+export const postRAMMachine = setup({
+  types: {
+    input: {} as Pick<PostRAMMachineContext, "project" | "artblocksClient">,
```
ryley-o, line 32: nit: can this be `PostRAMMachineInput`?

packages/sdk/src/machines/post-ram-machine/index.ts
```diff
@@ -0,0 +1,126 @@
...
+    context: {} as PostRAMMachineContext,
+    events: {} as PostRAMMachineEvents,
+  },
+  actors: {
+    fetchUserBids: fromPromise(
+      async ({
+        input,
+      }: {
+        input: Pick<PostRAMMachineContext, "artblocksClient" | "project">;
```
ryley-o, line 41: same `PostRAMMachineInput`?

ryley-o (approved):

packages/sdk/src/machines/post-ram-machine/index.ts
yoshiwarab: This one I think  I'd actually prefer to leave as is because PostRAMMachineInput is specifically for the input required to spawn the machine and this is input to a specific actor.

PR #1687, opened 2024-10-02
Filter RAM from CD minter options if not curated

ryley-o:
## Description of the change

Filter RAM from CD minter options if not curated.

This logic should be removed once RAM is supported on platform frontend.

sdk prerelease version is bumped.


ryley-o:
closing in favor of frontend solution

PR #1696, opened 2024-10-18
Resimulate transaction on revert during waiting for confirmation to get error message

yoshiwarab:
https://linear.app/art-blocks/issue/CUR-1837/change-error-when-topupbid-is-reverted

### DIFF
--- a/packages/sdk/src/machines/ram-machine/index.ts
+++ b/packages/sdk/src/machines/ram-machine/index.ts
@@ -1,5 +1,6 @@
 import {
   Chain,
+  ContractFunctionRevertedError,
   Hex,
   JsonRpcAccount,
   SimulateContractReturnType,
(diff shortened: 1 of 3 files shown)


Asupkay (approved):

yoshiwarab:
@ryley-o I think this is the only way we can get the revert message but let me know if you know of something more direct!

asaltste (approved):

ryley-o (approved):
there might be something more direct with `getTransactionError`, and re-simulating could result in a different error here, but I think this is good in the short term

packages/sdk/src/machines/ram-machine/index.ts
```diff
@@ -303,17 +305,38 @@ export const ramMachine = setup({
...
+          // been updated between the receipt being returned and the trace
+          // being fetched it is possible that the revert reason will not be
+          // accurate in this case.
+          try {
+            const traceResult = await traceTransaction(publicClient, txHash);
+
+            throw new ContractFunctionRevertedError({
+              abi: minterRAMV0Abi,
+              functionName: txRequest.functionName,
+              data: traceResult?.[0]?.result?.output,
```
ryley-o, line 323: Just to check, does this get properly decoded on the frontend? I think the revert reason is in args[0], but maybe we are accessing that when decoding later on?

ryley-o (approved):
LGTM - have you validated this is wai?

PR #1708, opened 2024-10-31
Use universal reader in on-chain generator

ryley-o:
## Description of the change

Update on-chain generator to use universal reader to expand compatibility and simplify code.

This is intended to fold into #1320 

Updating the on-chain generator to use the universal reader improves maintainability of the on-chain generator by delegating compatibility responsibilities to the universal reader contract.

Some gas-efficiency is traded for broader support + maintainability in the PR. There are now more data being passed through contract calls. However, the concatenation of all script chunks is still performed in assembly, preventing any gas costs that scale faster than `n`, where `n` is number of script chunks/dependency chunks.

A test is added to validate support of BytecodeStorageV2 on-chain compression.

fixes https://linear.app/art-blocks/issue/PLT-829/add-on-chain-compression-capability-to-gen

### DIFF
--- a/packages/contracts/contracts/generator/AddressChunks.sol
+++ b/packages/contracts/contracts/generator/AddressChunks.sol
@@ -4,15 +4,45 @@
 pragma solidity ^0.8.17;
 
 library AddressChunks {
+    /**
+     * @notice Merge multiple chunks of contract code into a single bytes array.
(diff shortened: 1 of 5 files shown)


lindsgil (approved):
🔥 🔥 🔥

yoshiwarab (approved):
🚀

PR #1726, opened 2024-11-22
Update Dependency Registry to use core registry

lindsgil:
## Description of the change

Update Dependency Registry to use core registry for determining if supported instead of it's hardcoded list

### DIFF
--- a/packages/contracts/contracts/DependencyRegistryV0.sol
+++ b/packages/contracts/contracts/DependencyRegistryV0.sol
@@ -43,8 +43,6 @@ contract DependencyRegistryV0 is
     using EnumerableSet for EnumerableSet.AddressSet;
     using SafeCast for uint24;
 
-    address public coreRegistryAddress;
-
     uint8 constant AT_CHARACTER_CODE = uint8(bytes1("@")); // 0x40
 
     function _onlyNonZeroAddress(address address_) internal pure {
@@ -70,9 +68,9 @@ contract DependencyRegistryV0 is
         address coreContractAddress
     ) internal view {
         require(
-            ICoreRegistryV1(coreRegistryAddress).isRegisteredContract(
-                coreContractAddress
-            ),
+            ICoreRegistryV1(
+                DependencyRegistryStorageLib.s().coreRegistryContract
+            ).isRegisteredContract(coreContractAddress),
             "Core contract not supported"
         );
     }
@@ -107,18 +105,12 @@ contract DependencyRegistryV0 is
      * @notice Initializes contract.
      * @param adminACLContract_ Address of admin access control contract, to be
      * set as contract owner.
-     * @param coreRegistryAddress_ Address of core registry contract.
      */
-    function initialize(
-        address adminACLContract_,
-        address coreRegistryAddress_
-    ) public initializer {
+    function initialize(address adminACLContract_) public initializer {
         __Ownable_init();
 
         // set AdminACL management contract as owner
         _transferOwnership(adminACLContract_);
-        // set core registry address
-        coreRegistryAddress = coreRegistryAddress_;
     }
 
     /**
@@ -128,7 +120,9 @@ contract DependencyRegistryV0 is
     function updateCoreRegistryAddress(address _coreRegistryAddress) external {
         _onlyAdminACL(this.updateCoreRegistryAddress.selector);
         _onlyNonZeroAddress(_coreRegistryAddress);
-        coreRegistryAddress = _coreRegistryAddress;
+        DependencyRegistryStorageLib.s().coreRegistryContract = ICoreRegistryV1(
+            _coreRegistryAddress
+        );
     }
 
     /**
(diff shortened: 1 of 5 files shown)


packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -104,12 +107,28 @@ contract DependencyRegistryV0 is
      * @notice Initializes contract.
      * @param adminACLContract_ Address of admin access control contract, to be
      * set as contract owner.
+     * @param coreRegistryAddress_ Address of core registry contract.
      */
-    function initialize(address adminACLContract_) public initializer {
+    function initialize(
+        address adminACLContract_,
+        address coreRegistryAddress_
+    ) public initializer {
```
ryley-o, line 115: the initialize function cannot be called after an upgrade.
we have two options:
* define a new function with modifier `reinitializer(2)` and make it public, and call it atomically when upgrading via multisig call to `ProxyAdmin.upgradeAndCall` (we would need to build our bytes manually)
* skip reinitializer abstraction, and rely on the `updateCoreRegistryAddress`, while accepting some minor downtime on the dep registry until we update core registry address via `function updateCoreRegistryAddress`

I vote for the latter, which means initialize can remain unchanged in this PR

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -616,40 +635,6 @@ contract DependencyRegistryV0 is
...
-        require(
-            // @dev the remove function returns false if set does not contain value
-            DependencyRegistryStorageLib.s().supportedCoreContracts.remove(
-                contractAddress
-            ),
-            "Core contract already removed or not in set"
-        );
-        emit SupportedCoreContractRemoved(contractAddress);
-    }
-
```
ryley-o, line 652: would be acceptable to leave this as commented code with a note that it was removed in an upgrade. up to you!

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -1023,16 +1001,13 @@ contract DependencyRegistryV0 is
...
-        );
+        uint256 count = ICoreRegistryV1(coreRegistryAddress)
+            .getNumRegisteredContracts();
+        address[] memory supportedCoreContracts = new address[](count);
 
-        for (uint256 i; i < supportedCoreContractCount; ) {
-            supportedCoreContracts[i] = ds.supportedCoreContracts.at(i);
+        for (uint256 i; i < count; ) {
+            supportedCoreContracts[i] = ICoreRegistryV1(coreRegistryAddress)
+                .getRegisteredContractAt(i);
```
ryley-o, line 1010: can use the core registry function `getAllRegisteredContracts`

packages/contracts/contracts/libs/v0.8.x/DependencyRegistryStorageLib.sol
```diff
@@ -66,9 +66,6 @@ library DependencyRegistryStorageLib {
         mapping(bytes32 dependencyNameAndVersion => Dependency dependency) dependencyRecords;
         // mapping from licenseTypes to License, which stores the properties of each license
         mapping(bytes32 licenseType => License license) allLicenses;
-        // Set of addresses for the core contracts that are supported by the DependencyRegistry.
-        // Each address represents a unique core contract in the Art Blocks ecosystem.
-        EnumerableSet.AddressSet supportedCoreContracts;
```
ryley-o, line 71: cannot change existing storage layout in upgradeable contracts, ref https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#modifying-your-contracts

ryley-o (commented):
a few notes on upgrades - interfaces lgtm

packages/contracts/contracts/DependencyRegistryV0.sol
```diff
@@ -42,6 +43,8 @@ contract DependencyRegistryV0 is
     using EnumerableSet for EnumerableSet.AddressSet;
     using SafeCast for uint24;
 
+    address public coreRegistryAddress;
+
```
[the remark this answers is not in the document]
lindsgil: updated!

packages/contracts/contracts/DependencyRegistryV0.sol
lindsgil: thanks for flagging 🙏  went with the latter and reverted the changes to the initialize function.. is there any documentation i should update so we don't forget to call `updateCoreRegistryAddress`?

packages/contracts/contracts/DependencyRegistryV0.sol
lindsgil: added this back in with a code comment!

packages/contracts/contracts/DependencyRegistryV0.sol
lindsgil: ah perfect, i updated this (and added getAllRegisteredContracts to the core registry interface)

packages/contracts/contracts/libs/v0.8.x/DependencyRegistryStorageLib.sol
lindsgil: 🙏 🙏  thanks for catching - I added this back with a dev comment

packages/contracts/contracts/DependencyRegistryV0.sol
ryley-o: I wouldn't worry about it too much, since we only have to do it once on the upcoming upgrade 👍

lindsgil: sounds great :D

packages/contracts/test/dependency-registry/DependencyRegistryV0.test.ts
```diff
@@ -1973,155 +1973,13 @@ describe(`DependencyRegistryV0`, async function () {
...
           dependencyWebsite
         );
+      config.coreRegistry = await deployAndGet(config, "CoreRegistryV1", []);
+      await config.dependencyRegistry
+        .connect(config.accounts.deployer)
+        .updateCoreRegistryAddress(config.coreRegistry.address);
       // pass config to tests in this describe block
       this.config = config;
     });
-    describe("addSupportedCoreContract", function () {
```
ryley-o, line 1979: Love all this removed logic!

ryley-o (approved):
LGTM!

PR #1728, opened 2024-11-25
Audit engine registry to ensure all contracts are registered

lindsgil:
## Description of the change

Add one-off script to register any unregistered contracts on the core registry 

TODO:
- [x] PR Review
- [x] Queue transactions
- [x] Execute tx: https://etherscan.io/tx/0x373e9dcab825fc5ec253e6bc95bac125ce71fa84b526381244a956118c0683b4
- [x] Merge

### DIFF
--- a/packages/contracts/scripts/one-off/add-unregistered-contracts.ts
+++ b/packages/contracts/scripts/one-off/add-unregistered-contracts.ts
@@ -1,0 +1,206 @@
+import hre, { ethers } from "hardhat";
+import { GetUnregisteredContractsDocument } from "../../generated/graphql";
+import { getClient } from "../util/graphql-client-utils";
+import { getNetworkName } from "../util/utils";
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
+import { getDeployerWallet } from "../util/get-deployer-wallet";
+import { Wallet } from "ethers";
+import Safe, { EthersAdapter } from "@safe-global/protocol-kit";
+import SafeApiKit from "@safe-global/api-kit";
+import { MetaTransactionData } from "@gnosis.pm/safe-core-sdk-types";
+import { CoreRegistryV1__factory } from "../contracts";
+
+//////////////////////////////////////////////////////////////////////////////
+// CONFIG BEGINS HERE
+//////////////////////////////////////////////////////////////////////////////
+type BaseConfig = {
+  network: string;
+  coreRegistryAddress: string;
+  useLedgerSigner: boolean;
+};
+
+type GnosisSafeConfig = BaseConfig & {
+  useGnosisSafe: true;
+  safeAddress: string;
+  transactionServiceUrl: string;
+};
+
+type NoGnosisSafeConfig = BaseConfig & {
+  useGnosisSafe: false;
+  safeAddress?: never;
+  transactionServiceUrl?: never;
+};
+
+type Config = GnosisSafeConfig | NoGnosisSafeConfig;
+
+const config: Config = {
+  network: "mainnet",
+  coreRegistryAddress: "0x652490c8bb6e7ec3fd798537d2f348d7904bbbc2",
+  useLedgerSigner: false,
+  useGnosisSafe: false,
+};
+
+//////////////////////////////////////////////////////////////////////////////
+// CONFIG ENDS HERE
+//////////////////////////////////////////////////////////////////////////////
+
+async function main() {
+  const networkName = await getNetworkName();
+
+  if (networkName !== config.network) {
+    throw new Error(
+      `network name ${networkName} does not match expected network name ${config.network}`
+    );
+  }
+
+  if (
+    config.useGnosisSafe &&
+    (!config.safeAddress || !config.transactionServiceUrl)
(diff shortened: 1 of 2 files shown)


packages/contracts/scripts/one-off/add-unregistered-contracts.ts
```diff
@@ -0,0 +1,206 @@
...
+  const coreRegistry = CoreRegistryV1__factory.connect(
+    config.coreRegistryAddress,
+    signer
+  );
+
+  const txData: MetaTransactionData[] = [];
+  for (const contract of res.data.contracts_metadata) {
+    const contractCoreVersion = contract.core_version
+      ? ethers.utils.formatBytes32String(contract.core_version)
+      : ethers.constants.HashZero; // "0x0000000000000000000000000000000000000000000000000000000000000000"
```
ryley-o, line 141: thould we throw here (and on contractType) instead of falling back to hashZero?

packages/contracts/scripts/one-off/add-unregistered-contracts.ts
```diff
@@ -0,0 +1,206 @@
...
+          contractType
+        );
+        await tx.wait();
+        console.log(`Successfully registered contract: ${contract.address}`);
+      } catch (error) {
+        console.error(
+          `Failed to register contract ${contract.address}:`,
+          error
+        );
+      }
```
ryley-o, line 171: see note about how this needs to go through the EngineFactory
ref https://github.com/ArtBlocks/artblocks-contracts/blob/main/packages/contracts/INFRASTRUCTURE.md#mainnet

additionally, recommend using registerMultipleContracts to stay gas efficient, as well as avoiding nonce errors in gnosis safe (the `gnosisSetup.protocolKit.getNonce()` function returns next nonce to execute, not next nonce to queue)

packages/contracts/scripts/one-off/add-unregistered-contracts.ts
lindsgil: Skimming through the result of the query it looks like there are quite a few `GenArt721CoreV2_PBAB` contracts that have a null core_version field, I think throwing on a null contractType makes sense but we'd still need to handle the NULL core_version contracts 🤔 

Example -

```
        "contract_type": "GenArt721CoreV2_PBAB",
        "core_version": null,
        "registered_on": null,
        "address": "0xff124d975c7792e706552b18ec9da24781751cab"
```

yoshiwarab: Every v0-v2 contract will have a null version. Should we just hard code versions for those contract types?

packages/contracts/scripts/one-off/add-unregistered-contracts.ts
```diff
@@ -0,0 +1,206 @@
...
+  useGnosisSafe: false;
+  safeAddress?: never;
+  transactionServiceUrl?: never;
+};
+
+type Config = GnosisSafeConfig | NoGnosisSafeConfig;
+
+const config: Config = {
+  network: "mainnet",
+  coreRegistryAddress: "0x652490c8bb6e7ec3fd798537d2f348d7904bbbc2",
```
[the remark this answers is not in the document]
lindsgil: updated this to the engine factory address per below!

packages/contracts/scripts/one-off/add-unregistered-contracts.ts
```diff
@@ -0,0 +1,206 @@
...
+  useLedgerSigner: false,
+  useGnosisSafe: false,
```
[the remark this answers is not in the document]
lindsgil: ahh thanks 🙏  updated!

packages/contracts/scripts/one-off/add-unregistered-contracts.ts
lindsgil: Good call - updated this to set the version depending on the contract_type

packages/contracts/scripts/one-off/add-unregistered-contracts.ts
lindsgil: thanks! refactored this to call registerMultipleContracts on the engine factory 🙏

packages/contracts/scripts/one-off/add-unregistered-contracts.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  //////////////////////////////////////////////////////////////////////////////
+  // ACTION BEGINS HERE
+  //////////////////////////////////////////////////////////////////////////////
+
+  // Query for unregistered contracts
+  const registryAddresses = [
+    "0x2ee7b9bb2e038be7323a119701a191c030a61ec6",
+    "0x652490c8bb6e7ec3fd798537d2f348d7904bbbc2",
```
ryley-o, line 119: I think we ONLY want `0x2ee7b9bb2e038be7323a119701a191c030a61ec6` - `0x652490c8bb6e7ec3fd798537d2f348d7904bbbc2` is legacy, and won't be checked by the dependency registry

ryley-o (approved):

PR #1730, opened 2024-11-26
Update dependency registry with supported contract overrides

ryley-o:
## Description of the change

Update dependency registry with supported contract overrides.

fixes https://linear.app/art-blocks/issue/PLT-901/update-dependency-registry-with-supported-contract-overrides

This allows the dependency registry to add supported contracts in addition to what is defined on the core registry.

This supports the use case where a core contract is banned from minting on the shared minter suite, but has projects should be supported by the dependency registry (e.g. safe fallback if compromised partner admin keys).

Note: an event is also added when core registry is set, for future indexing capabilities.

## TODO

- [x] add tests for coverage of new logic
- [x] merge #1727 before merging this

### DIFF
--- a/packages/contracts/contracts/DependencyRegistryV0.sol
+++ b/packages/contracts/contracts/DependencyRegistryV0.sol
@@ -67,10 +67,9 @@ contract DependencyRegistryV0 is
     function _onlySupportedCoreContract(
         address coreContractAddress
     ) internal view {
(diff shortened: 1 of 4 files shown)


lindsgil (approved):
In the future do you think it makes sense to explore role based access control for contract addresses? 🤔

yoshiwarab (approved):

PR #1766, opened 2025-02-21
Update project sale machines to fetch user eligibility data while paused

yoshiwarab:
## Description of the change

This change is necessary for the implementation of the purchase flow in the new unified site designs

### DIFF
--- a/packages/sdk/src/machines/purchase-initiation-machine/index.ts
+++ b/packages/sdk/src/machines/purchase-initiation-machine/index.ts
@@ -108,9 +108,12 @@ export type UserPurchaseContext =
  * necessary data, such as the user's token balance and allowance.
  *
  * If the user is eligible for the purchase, the machine transitions to the
- * 'readyForPurchase' state, indicating that it is prepared to initiate the purchase
- * process. If the user is ineligible, the machine moves to the
- * 'userIneligibleForPurchase' state.
+ * 'waitingForStart' state, where it polls until the sale has officially started.
+ * Once the sale begins, it automatically transitions to 'readyForPurchase'. If the
+ * user is ineligible, the machine moves to the 'userIneligibleForPurchase' state.
+ * This approach prepares and makes available all necessary purchase context before
+ * the sale actually begins, allowing users to visit a sale page and see if they're
+ * eligible (for example, if they're on an allowlist) even before the sale starts.
  *
  * When the 'INITIATE_PURCHASE' event is received in the 'readyForPurchase' state, the
  * machine transitions to the 'initiatingPurchase' state. In this state, it invokes the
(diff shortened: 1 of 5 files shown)


packages/sdk/src/machines/purchase-initiation-machine/index.ts
```diff
@@ -389,6 +395,35 @@ export const purchaseInitiationMachine = setup({
     ) => {
       return !isERC20AllowanceSufficient(context.project, allowance);
     },
+    isSaleStarted: ({ context }) => {
+      const project = context.project;
+      const liveSaleData: LiveSaleData =
+        context.projectSaleManagerMachine.getSnapshot().context.liveSaleData;
+      const walletClient = context.artblocksClient.getWalletClient();
+
+      if (!liveSaleData || !walletClient?.account) {
```
mattetre, line 404: Not 100% sure how `isSaleStarted` is used (but I am reading it as "has the sale for the project started already") in which case couldn't the sale be started even if the users wallet isn't connected via `walletClient?.account`? (but maybe that isn't the intention of this and its more of a "has the sale started and can I buy it"?)

packages/sdk/src/machines/purchase-initiation-machine/index.ts
```diff
@@ -389,6 +395,35 @@ export const purchaseInitiationMachine = setup({
...
+        return false;
+      }
+
+      console.log(liveSaleData.paused);
+      console.log(project.artist_address);
+      console.log(walletClient.account.address.toLowerCase());
```
mattetre, line 410: nit: logs

mattetre (approved):
🥇

packages/sdk/src/machines/purchase-initiation-machine/index.ts
yoshiwarab: Ah yeah it's more meant to be "is the project purchasable by the user" will update the name to make more clear!

packages/sdk/src/machines/purchase-initiation-machine/index.ts
```diff
@@ -451,6 +489,18 @@ export const purchaseInitiationMachine = setup({
         },
       },
     },
+    waitingForStart: {
+      always: {
+        target: "readyForPurchase",
+        guard: "isSaleStarted",
+      },
+      after: {
+        1000: {
```
ryley-o, line 498: Just to double, this shouldn't trigger any api or web3 calls, right? that's what it looks like to me, but checking since this is 1sec

ryley-o (approved):

packages/sdk/src/machines/purchase-initiation-machine/index.ts
yoshiwarab: yup! it just pulls the live sale data from the parent project sale machine which is updated with new data every 10 seconds.

yoshiwarab: and just web3 calls

PR #1775, opened 2025-03-11
On-Chain Generator support of Web3Call

ryley-o:
## Description of the change

Update the On-Chain generator to support Web3Call capabilities.

This updates the on-chain generator to inject web3call data if detected via the special string `#web3call_contract#` returned from the universal reader/bytecode storage reader contracts.

## Design

All key/value pairs are base64 encoded to support special characters, and all decoding is done automatically via updated reviver function.

## Tests

Tests are added to ensure backwards compatibility + support of new web3call cases.

### DIFF
--- a/packages/contracts/test/generator/GenArt721GeneratorV0.test.ts
+++ b/packages/contracts/test/generator/GenArt721GeneratorV0.test.ts
@@ -79,6 +79,10 @@ interface GenArt721GeneratorV0TestConfig extends T_Config {
   universalReader: UniversalBytecodeStorageReader;
 }
 
+// the json parser's reviver script used when parsing the token data
+const REVIVER_SCRIPT =
+  '(key, value) => key === "data" && value !== null ? value.startsWith("#web3call#") ? Object.entries(JSON.parse(atob(value.slice(10)))).reduce((acc, [k, v]) => ((acc[atob(k)] = atob(v)), acc), {}) : atob(value) : value)';
+
 describe(`GenArt721GeneratorV0`, async function () {
   const p5NameAndVersion = "p5js@1.0.0";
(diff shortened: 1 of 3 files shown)


lindsgil (approved):
🙌

packages/contracts/test/generator/GenArt721GeneratorV0.test.ts
```diff
@@ -290,7 +299,7 @@ describe(`GenArt721GeneratorV0`, async function () {
       const hash = hashes[0];
       expect(tokenHtml).to.include(
         getScriptTag(
-          `let tokenData = JSON.parse(\`{"tokenId":"0","hashes":["${hash}"]}\`, (key, value) => key === "data" && value !== null ? atob(value) : value);`
+          `let tokenData = JSON.parse(\`{"tokenId":"0","hashes":["${hash}"]}\`, (key, value) => key === "data" && value !== null ? value.startsWith("#web3call#") ? Object.entries(JSON.parse(atob(value.slice(10)))).reduce((acc, [k, v]) => ((acc[atob(k)] = atob(v)), acc), {}) : atob(value) : value);`
```
yoshiwarab, line 302: Nit: might make sense to break out a function to generate this string so we can just update in one place if this changes again

yoshiwarab (approved):

packages/contracts/test/generator/GenArt721GeneratorV0.test.ts
ryley-o: good call - I broke out the reviver function as a constant since it is used across several of the tests. The other variation is mostly due to certain contract-version-specific formats, so figured okay to let those stay as is. If we update reviver in future, will only be one place now 👍

PR #1776, opened 2025-03-11
PMP Gas Optimizations

ryley-o:
## Description of the change

Improve PMP gas costs by optimizing storage patterns.

## Summary

(typical) configureProject Gas Reduction
* Before: 572,274 gas
* After: 403,079 gas
* Change: 🔽 169,195 less gas (~29.6% reduction)

(typical) configureTokenParams Gas Reduction
* Before: 154,304 gas
* After: 152,324 gas
* Change: 🔽 1,980 less gas (~1.3% reduction)

## Changes

The primary way gas costs are reduced are by using the new Immutable Strings library to store select options.

Additionally, because only the array length of select options is used when configuring a token, the array length is (redundantly) stored packed as a uint8 in an already-hot slot to remove a costly SLOAD on a cold slot.


lindsgil (approved):
🔥 ⛽

PR #1786, opened 2025-03-27
on-chain generator gas measurements

ryley-o:
## Description of the change

Measure some read operations for on-chain generator


lindsgil (approved):
⛽

ryley-o:
looks like this is too much memory for our CI - going to just close, but we will have a record of it anyway in case we ever want to re-look 🙏

PR #1805, opened 2025-04-11
simple open select example

ryley-o:
## Description of the change

\<add description here\>


ryley-o:
closing since this was just a temporary simple example until we ship frontend updates

PR #1814, opened 2025-04-24
Update dependency registry with three new script fields

ryley-o:
## Description of the change

Update dependency registry with three new script fields to fully prescribe generator behavior.

The three new fields enable the dependency registry to sufficiently prescribe how each project should be rendered in the generator. This will lead to the removal of any dependency-name-based logic from the on-chain generator in a subsequent PR.


yoshiwarab (approved):

ryley-o:
closing in favor of #2005 2005

PR #1815, opened 2025-04-26
event data for PMP "changed by"

ryley-o:
## Description of the change

Add data to the event when a PMP is configured to indicate which authenticated address was used to perform the change.

This is important to provide provenance information about the change history of a token.

The changes here respect delegation - if a delegate executes a change on behalf of the token owner, the token owner's address will be indicated as the address that made the change, not the delegate.

## Gas Impacts

Gas tests indicate +1.8% gas to the 2x typical PMP gas test.

This is expected due to (first order) more event data being emitted, and (second order) a slightly less optimal order of typical auth checks.

## Test

Tests are updated to check for the new event data, including the case of a delegate making a change on behalf of a token owner.

### DIFF
--- a/packages/contracts/contracts/interfaces/v0.8.x/IPMPV0.sol
+++ b/packages/contracts/contracts/interfaces/v0.8.x/IPMPV0.sol
@@ -52,11 +52,13 @@ interface IPMPV0 is IWeb3Call {
      * @param coreContract The address of the core contract.
      * @param tokenId The token ID for which parameters were configured.
      * @param pmpInputs Array of parameter inputs that were configured.
+     * @param authAddresses Array of addresses that authenticated the parameters. Aligned by index with pmpInputs.
      */
     event TokenParamsConfigured(
         address coreContract,
         uint256 tokenId,
-        PMPInput[] pmpInputs
+        PMPInput[] pmpInputs,
+        address[] authAddresses
     );
 
     /**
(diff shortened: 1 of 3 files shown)


packages/contracts/contracts/interfaces/v0.8.x/IPMPV0.sol
```diff
@@ -52,11 +52,13 @@ interface IPMPV0 is IWeb3Call {
      * @param coreContract The address of the core contract.
      * @param tokenId The token ID for which parameters were configured.
      * @param pmpInputs Array of parameter inputs that were configured.
+     * @param authAddresses Array of addresses that authenticated the parameters. Aligned by index with pmpInputs.
      */
     event TokenParamsConfigured(
         address coreContract,
         uint256 tokenId,
-        PMPInput[] pmpInputs
+        PMPInput[] pmpInputs,
```
lindsgil, line 60: What do you think about adding a new field on the PMPInput struct, `configuringAddress`, instead of isolating authAddresses as a new parameter?

ryley-o: I chose this format because the `pmpInputs` is the input struct to the configure token params function. Totally agree that it would be ideal to add the new field to the PMPInput struct from the event's perspective, but it's also used as function input so we would need a new struct for the event. I didn't want to spend the gas during execution to copy all the calldata a different struct in memory, so figured best to have a slightly awkward event param for efficiency. Happy to chat more about it though!

lindsgil: ah makes sense! 🙏  appreciate the details

lindsgil (approved):
🔥 🔥

PR #1827, opened 2025-05-13
Example hook for block height + artist pmp overrides

ryley-o:
## Description of the change

Add example hook for injecting block height and artist project overrides.

The overrides:
* overwrite any existing pmp values
* get written as new pmp values if their key was not included in the hook input

TODO

- [x] tests


lindsgil (approved):

ryley-o:
re-requested review because took this in a slightly different, more generic direction 🙏

lindsgil (approved):
🔥 🔥 🔥

PR #1852, opened 2025-07-24
Add LIFT PostParam hooks

ryley-o:
## Description of the change

Add LIFT PostParam hooks.

Adds ability to configure any owned squiggle as a PostParam, and injects the configured squiggle id and hash.

Considers delegate.xyz V1 or V2. Supports V2 subdelegation "rights" consistent with PMPV0. Also allows configuring squig #9999 if token owner is inscribed on relic contract. 

Allows resetting token back to default squiggle #1981 at any point, which view augmentation will strip squiggle details back to default.

Supports arbitrary additional PostParams (e.g. PFP_Mode boolean), merging results as necessary.

## Tests

Tests are added for core functionality. Mainnet fork testing was used due to complex integration patterns across various legacy contracts, including squiggle core contract, PMPV0, delegate v1, delegate v2, and relic.

## Checks

The hook assumes:
* PostParam name is `"Featured_Squiggle"`
  * type: Uint256Range
  * Range: 0-9999


lindsgil (approved):
💪 💪 💪

PR #1987, opened 2026-05-20
Add MinterSetPriceTieredOnChainAllowV0

lindsgil:
## Description of the change

Add MinterSetPriceTieredOnChainAllowV0, fork of MinterSetPriceOnChainAllowV0 that adds separate allowlist and public pricing, as well as the ability to batch mint

- [x] Tests
- [x] PR review
- [x] Deploy to dev


packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
```diff
@@ -0,0 +1,844 @@
...
+            projectId: projectId,
+            coreContract: coreContract,
+            sender: msg.sender
+        });
+        SetPriceLib.updatePricePerToken({
+            projectId: projectId,
+            coreContract: coreContract,
+            pricePerToken: publicPricePerTokenInWei
+        });
+        _allowlistPriceStorage().prices[coreContract][
```
yoshiwarab, line 165: Do we want to emit an event for allowlist price indexing?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
```diff
@@ -0,0 +1,844 @@
...
+
+    /**
+     * @notice Gets the allowlist price for project `projectId`.
+     * @param projectId Project ID to get allowlist price information for
+     * @param coreContract Contract address of the core contract
+     * @return allowlistPricePerTokenInWei current allowlist price per token
+     * on this minter, in Wei. Only valid if prices have been configured
+     * (check via `getPriceInfo`).
+     */
+    function getAllowlistPriceInfo(
```
yoshiwarab, line 527: Do you think it would be useful to add a getPriceInfoForAddress function (honestly not sure but the thought came to mind!)?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
lindsgil: Nice to have symmetry with SetPriceLib's event - added this `AllowlistPricePerTokenUpdated` event and updated the tests!

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
lindsgil: seems like it would be convenient for the front-end if we ever add addresses to the allowlist other than the executor wallet! added for convenience + tests

lindsgil:
It looks like yarn 4.15.0 was released a few days ago causing the ci pipelines to fail, will update this in a separate PR https://github.com/yarnpkg/berry/releases

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
```diff
@@ -0,0 +1,894 @@
...
+        uint256 projectId,
+        address coreContract
+    ) external view returns (uint256 allowlistPricePerTokenInWei) {
```
ryley-o, line 530: minor nit: might be also nice to mirror the return type of getPriceInfo here, including isConfigured and currency symbol, address - just thinking about how to reduce multiple calls later on.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
```diff
@@ -0,0 +1,894 @@
...
+     * @return isConfigured true only if prices have been configured for the
+     * project
+     * @return tokenPriceInWei effective price per token for the given wallet,
+     * in Wei
+     */
+    function getPriceInfoForAddress(
+        uint256 projectId,
+        address coreContract,
+        address wallet
+    ) external view returns (bool isConfigured, uint256 tokenPriceInWei) {
```
ryley-o, line 552: same minor nit about probably worth including currency symbol and address here to mirror getPriceInfo return types

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
```diff
@@ -0,0 +1,894 @@
...
+        for (uint256 i; i < quantity; ) {
+            SplitFundsLib.splitRevenuesETHNoRefund({
+                projectId: projectId,
+                valueInWei: pricePerTokenInWei,
+                coreContract: coreContract
+            });
+            unchecked {
+                ++i;
+            }
+        }
```
ryley-o, line 867: gas efficiency nit - can we just do this in one single split for gas efficiency? might even be cleanest to use the battle tested splitFundsETHRefundSender() function with a price per token set to pricePerTokenInWei * quantity or something. then we could remove the one-off refund logic in this minter.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
```diff
@@ -0,0 +1,894 @@
...
+    // to reduce contract bytecode size
+    // @dev contract uses AuthLib for some modifier-like functions
+
+    // Diamond storage for allowlist price configuration (minter-specific)
+    bytes32 constant ALLOWLIST_PRICE_STORAGE_POSITION =
+        keccak256("mintersetpricetieredonchainallowv0.allowlistprice.storage");
+
+    struct AllowlistPriceStorage {
+        mapping(address coreContract => mapping(uint256 projectId => uint256 allowlistPricePerToken)) prices;
+    }
```
ryley-o, line 89: Using addressed storage is fine here, but since the definition is in the concrete minter implementation instead of a library (where , I don't think there is a benefit vs. just using a typical contract mapping definition.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
```diff
@@ -0,0 +1,894 @@
...
+
+    event AllowlistPricePerTokenUpdated(
+        uint256 indexed projectId,
+        address indexed coreContract,
+        uint256 indexed allowlistPricePerToken
```
ryley-o, line 94: nit: typically the value here wouldn't be an indexed parameter

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
```diff
@@ -0,0 +1,894 @@
...
+    ISharedMinterV0
+{
+    /// Minter filter address this minter interacts with
+    address public immutable minterFilterAddress;
+
+    /// Minter filter this minter may interact with.
+    IMinterFilterV1 private immutable _minterFilter;
+
+    /// minterType for this minter
+    string public constant minterType = "MinterSetPriceTieredOnChainAllowV0";
```
ryley-o, line 73: I don't know if we ever cast these as bytes32 items, but we have kept these to <31 characters just in case in the past. I think the minterfilter might do something with that, so we might want to pick something shorter if we make this a globally approved minter...

ryley-o (approved):
everything I see is pretty minor nits - nothing critical, nice work here!
would like to talk through why we decided allow purchase multiple in this one. I think it might also might be worth calling out a bit more specifically in the contract natspec about how this allowlist is likely used with more of a privileged relay-type system to integrate off-chain payment options, and it is not fully featured to support many allowlist scenarios due to lack of per-allowlisted-wallet invocation limits.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
lindsgil: updated in https://github.com/ArtBlocks/artblocks-contracts/pull/1998

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
lindsgil: updated in https://github.com/ArtBlocks/artblocks-contracts/pull/1998

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
lindsgil: updated in https://github.com/ArtBlocks/artblocks-contracts/pull/1998

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
lindsgil: updated in https://github.com/ArtBlocks/artblocks-contracts/pull/1998

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredOnChainAllowV0.sol
lindsgil: updated in https://github.com/ArtBlocks/artblocks-contracts/pull/1998

PR #1998, opened 2026-07-14
Add MinterSetPriceTieredAllowV1 and tests

lindsgil:
## Description of the change

Adds Tiered Price Allowlist USDC minter + Tests


packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
```diff
@@ -0,0 +1,816 @@
...
+            address currencyAddress
+        )
+    {
+        SetPriceLib.SetPriceProjectConfig
+            storage setPriceProjectConfig_ = SetPriceLib
+                .getSetPriceProjectConfig({
+                    projectId: projectId,
+                    coreContract: coreContract
+                });
+        isConfigured = setPriceProjectConfig_.priceIsConfigured;
```
ryley-o, line 562: I think worth a dev comment explaining why && currencyAddress != address(0) check is not needed on this minter

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
```diff
@@ -0,0 +1,816 @@
...
+                projectId: projectId,
+                coreContract: coreContract
+            });
+    }
+
+    /**
+     * @notice Gets your balance of this minter's fixed USDC token.
+     * @return balance Balance of USDC
+     */
+    function getYourBalanceOfProjectERC20()
```
ryley-o: might be worth including input params of projectId, coreConttract to this function and in `checkYourAllowanceOfProjectERC20` for consistency with `MinterSetPriceERC20V5` interface. Consistent feels slightly better than "correct" here, for downstream system simplification... wdyt?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
```diff
@@ -0,0 +1,816 @@
...
+     * @param projectId Project ID to mint a token on.
+     * @param coreContract Core contract address for the given project.
+     * @param maxPricePerToken Maximum price of token being allowed by the
+     * purchaser, in USDC base units.
+     * @param currencyAddress Currency address of token. Must equal this
+     * minter's fixed USDC address.
+     * @param hashSeed Hash seed to assign to the token. Must be non-zero.
+     * @return tokenId Token ID of minted token
+     */
+    function purchaseToWithHashSeed(
```
ryley-o, line 722: something we haven't thought about too much, but I think this would be the first potentially open system where we could allow a "mint with hash seed" function to be called.
just a callout - there is no inherent hash seed uniqueness in this minter's flow. polyptych and server-only minting both can manage unique hash seed enforcement, but in this pattern, twin tokens (or any number of tokens) could end up with the same hash seeds. that's not necessarily a bug, but it might be an unintended result that isn't obvious.
I suggest maybe adding a project-scoped mapping of "used hash seeds" on the minter itself. That helps mitigate a lot of the unintended cases.
open to pushback or alternate solutions here though!

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
```diff
@@ -0,0 +1,816 @@
...
+
+    /// @notice Mapping of core contract => projectId => allowlist price per token in USDC base units
+    mapping(address coreContract => mapping(uint256 projectId => uint256 allowlistPricePerToken))
+        private _allowlistPricePerToken;
+
+    event AllowlistPricePerTokenUpdated(
+        uint256 indexed projectId,
+        address indexed coreContract,
+        uint256 allowlistPricePerToken
+    );
```
ryley-o: could this use the generic `GenericMinterEventsLib.ConfigValueSet`, for automatic indexing into extra minter details?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
```diff
@@ -0,0 +1,816 @@
...
+            });
+    }
+
+    /**
+     * @notice Returns whether `wallet` is the privileged allowlist address
+     * for this minter.
+     * @param wallet The address to check.
+     * @return bool True if the address is the configured allowlist address.
+     */
+    function isAllowlisted(address wallet) external view returns (bool) {
```
ryley-o: wdyt about updating this to be more in line with the allowlist inputs of `isAllowlisted(uint256,address,address)`, for consistency and compatibility with standard tooling?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
```diff
@@ -0,0 +1,816 @@
...
+     * @return isConfigured true only if prices have been configured on this
+     * minter
+     * @return tokenPriceInWei current allowlist price of token on this minter,
+     * in USDC base units
+     * @return currencySymbol currency symbol for purchases of project on this
+     * minter.
+     * @return currencyAddress currency address for purchases of project on
+     * this minter.
+     */
+    function getAllowlistPriceInfo(
```
ryley-o, line 583: wdyt about adding this function + `getPriceInfoForAddress` to a new minter family interface like `ISharedMinterTieredAllowV0`? I think those are uniquely new things and feels ready to canonically define those interfaces since this will be globally approved on minter filter, yeah?

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
```diff
@@ -0,0 +1,816 @@
...
+    address public allowlistAddress;
+
+    // MODIFIERS
+    // @dev contract uses modifier-like internal functions instead of modifiers
+    // to reduce contract bytecode size
+    // @dev contract uses AuthLib for some modifier-like functions
+
+    /// @notice Mapping of core contract => projectId => allowlist price per token in USDC base units
+    mapping(address coreContract => mapping(uint256 projectId => uint256 allowlistPricePerToken))
+        private _allowlistPricePerToken;
```
ryley-o, line 107: I think these could in theory be moved to more of a namespaced/diamond storage pattern and even library, but also understand that they are totally fine being defined inline here, since we aren't really using a library on this minter at this time. if we do choose to use a minter, we should namespace the storage as well, allowing the library functions to deterministically and easily access these storage parameters.

ryley-o (approved):
generally lgtm!
feedback is all minor, and touches mostly on consistency opportunities and potential for a new library to standardize some of this minter's interface going forward.

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
lindsgil: 🙏  added in https://github.com/ArtBlocks/artblocks-contracts/pull/1998/changes/c1b6a60a522a0eebeb4b53bc678bdebebd796f0b

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
lindsgil: good call! updated in 067162bb0943951975d60e24f7918c7789043caf

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
lindsgil: good call out 🙏  added that mapping along with a `projectHashSeedIsUsed` view function to return whether or not a hashSeed has already been used for purchase on a project. it feels weird to enforce uniqueness so i didn't add that but added some dev comments calling that out! added in 709fd0678974511f3639f9a66bd2db577af8f571

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
lindsgil: nice, added in 368ddfacdc61df38ab4b327d7e2389b446cbd08e

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
lindsgil: nice, added in 878079ba4ed0f967c47fccdac8fe5998b49aa033

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
lindsgil: good call, added in 02b7a38dca6ca3ef67ae2f1fc62fe1d81e8b817f

packages/contracts/contracts/minter-suite/Minters/MinterSetPriceTieredAllowV1.sol
lindsgil: nice! will leave it like this for now, but seems like it'd be worth extracting into a lib if we introduce a second TieredAllow variant 🙏

PR #2005, opened 2026-08-13
dependency registry and on-chain generator updates

ryley-o:
## Description of the change

Update dependency registry with new metadata fields that inform the on-chain generator how to build html template without special-case logic per dependency name.

# Registry-driven rendering directives (fixes ES module projects)

## Description of the change

Projects declaring `three@0.167.0` — including Jack Butcher's **Gas Wars** — do not render in the
on-chain generator. Rather than adding a fourth hardcoded dependency branch, this PR moves all
rendering directives into the dependency registry and **deletes every dependency-name-based branch
from the generator**. Supporting a new dependency type becomes an admin transaction instead of a
contract upgrade.

This completes the intent of #1814, which added registry fields but explicitly deferred the
generator cleanup to a follow-up. That PR is superseded by this one.

> [!WARNING]
> **The order of deployment is important.** The new field defaults are behavior-*changing*. Upgrading
> the generator before backfilling the registry would drop the canvas from every
> `js`/`babylon`/`tone`/`zdog` project, stop emitting `application/processing`, and re-wrap
> `custom@na` documents in a script tag. See [Rollout order](#rollout-order).

## TODO

- [x] review proposed contract updates
- [x] update tests
- [x] deploy per deployment plan
- [x] verify testnet is wai
- [x] upgrade mainnet


packages/contracts/contracts/interfaces/v0.8.x/IDependencyRegistryV0.sol
```diff
@@ -87,6 +87,89 @@ interface IDependencyRegistryV0 {
...
+        string dependencyWebsite;
+        // whether the dependency is available on chain
+        bool availableOnChain;
+        // count of on-chain scripts for dependency
+        uint24 scriptCount;
+        // whether the dependency's own script is an ES module, and therefore must be exposed to
+        // project scripts via an import map rather than loaded with a plain `<script src>` tag
+        bool loadAsModule;
+        // canvas tag requirement for the dependency
+        CanvasTag canvasTag;
```
yoshiwarab: Tiny nit, wonder if this shuld be CanvasTagType and canvasTagType to match project script. Also feels a little clearer, that it's not the literal tag bytes.

yoshiwarab (approved):

packages/contracts/contracts/interfaces/v0.8.x/IDependencyRegistryV0.sol
```diff
@@ -87,6 +87,89 @@ interface IDependencyRegistryV0 {
...
+    );
+
+    event DependencyProjectScriptTagTypeUpdated(
+        bytes32 indexed dependencyNameAndVersion,
+        ProjectScriptTagType projectScriptTagType,
+        string projectScriptSpecialType
+    );
+
+    /**
+     * @notice Enum representing the canvas tag requirements for a dependency.
```
lindsgil, line 107: does the dependency drive this or is it unique to artist script canvas handling? wondering if we can derive it by dependency

lindsgil (approved):

packages/contracts/contracts/interfaces/v0.8.x/IDependencyRegistryV0.sol
ryley-o: agree - updated in [c2df940](https://github.com/ArtBlocks/artblocks-contracts/pull/2005/commits/c2df940e0af4364fb65ed0ff524973f85a465800)

packages/contracts/contracts/interfaces/v0.8.x/IDependencyRegistryV0.sol
ryley-o: yup yup per-dependency.
now we store that in the dependency registry mappings instead of hard-coded logic per dependency in the on-chain generator 🫡 
Will be part of the backfill process

PR #2007, opened 2026-08-25
add transfer hooks to v3 cores

ryley-o:
## Description of the change

add transfer hooks to v3 cores

Per-project transfer hooks are in v3.3.0 Engine and Engine Flex (updated in place).

## Permissions

- **Artist or Admin ACL** can `configureProjectTransferHook` at any time (not tied to the 4-week metadata lock).
- **Artist only** can `lockProjectTransferHook` — one-way, any time, separate from the 4-week lock.
- **project auto-lock at `address(0)`** forever forbids assigning a hook, so that project stays on today's transfer security profile.
- **project auto-Lock with a hook set** does not freeze the hook - defers to `lockProjectTransferHook` state instead (auto-lock should not have potential to brick transfers).

## Dispatch

- Transfers: after the ERC-721 ownership write.
- Mints: after `assignTokenHash`, so the hook can read `tokenIdToHash`.
- A reverting hook aborts the whole mint/transfer (no latent failure).

## Security

- Unhooked transfers only pay extra SLOADs (no `DELEGATECALL`).
- A global `executing` flag blocks the hook from reentering any transfer/mint on the core (including other projects), or `configure`/`lock`.
- Non-zero hooks must ERC-165-advertise `ITransferHook`.
- A mutable/proxy hook can still change behavior after it is set — locking (at zero, or at a fixed implementation) is the intended mitigation.

**Deploy note:** new Engine/Flex implementations must link `V3TransferHookLib`. Engine Flex is **23.995 KiB** (just under the 24KB cap).

## IMPORTANT

24kb contract size limit required some logic offload into new external libs. In this instance, opted to keep Engine and Engine Flex consistently using a new V3 Engine Lib, even though non-flex did not completely require offloading. Value consistency for security and auditing purposes.


lindsgil (approved):
🔥 🔥

yoshiwarab (approved):

